PR #5, opened 2019-07-25
pages/getting-started: modify example unit file to start after network-online target.

EnigmaCurry:
Adds `After=network-online.target` to the example systemd unit file so that it only starts the container after networking has started. Otherwise the container would not start on boot.

### DIFF
--- a/modules/ROOT/pages/getting-started.adoc
+++ b/modules/ROOT/pages/getting-started.adoc
@@ -84,6 +84,8 @@ systemd:
       contents: |
         [Unit]
         Description=MyApp
+        After=network-online.target
+        Wants=network-online.target
 
         [Service]
         TimeoutStartSec=0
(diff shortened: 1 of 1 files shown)


EnigmaCurry:
I've double checked, `After=network.target` does not work, the image pull fails.

lucab:
https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/ has a good explanation.

In short, it would be better for the unit to support retries on network errors and just use `network.target`. If that is not possible for other reasons, an After+Wants on `network-online.target` is a good workaround with a few drawbacks.

modules/ROOT/pages/getting-started.adoc
```diff
@@ -114,6 +114,7 @@ systemd:
       contents: |
         [Unit]
         Description=Run single node etcd
+        After=network-online.target
```
lucab: This needs to be coupled with a `Wants=network-online.target`, as this target is not activated by default and thus needs to be pulled-in where needed.

EnigmaCurry: Ok, I tested that works.

EnigmaCurry:
I added the `Wants` as well as adding the config to both example on the page.

PR #7, opened 2019-08-21
pages/faq: clarify stance on upgrading directly from Atomic Host 

dustymabe:
Fixes https://github.com/coreos/fedora-coreos-tracker/issues/257

### DIFF
--- a/modules/ROOT/pages/faq.adoc
+++ b/modules/ROOT/pages/faq.adoc
@@ -130,9 +130,11 @@ https://discussion.fedoraproject.org/t/launch-faq-how-do-i-migrate-from-containe
 As with Container Linux, the best practice will be re-provisioning, due
 to the cloud-init/Ignition transition at least. Since Fedora CoreOS will
 be using rpm-ostree technology, it may be possible to rebase from Fedora
-Atomic Host to Fedora CoreOS. This will be part of a "migrating from
-Fedora Atomic Host" guide which will be published as soon as an actual
-version of Fedora CoreOS is available.
+Atomic Host to Fedora CoreOS, but it will not be recommended. It will
+be preferable to gain experience deploying systems using Ignition so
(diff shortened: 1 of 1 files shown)


dustymabe:
cc @keithy

modules/ROOT/pages/faq.adoc
```diff
@@ -130,9 +130,11 @@ https://discussion.fedoraproject.org/t/launch-faq-how-do-i-migrate-from-containe
...
 to the cloud-init/Ignition transition at least. Since Fedora CoreOS will
 be using rpm-ostree technology, it may be possible to rebase from Fedora
-Atomic Host to Fedora CoreOS. This will be part of a "migrating from
-Fedora Atomic Host" guide which will be published as soon as an actual
-version of Fedora CoreOS is available.
+Atomic Host to Fedora CoreOS, but it will not be recommended. It will
+be preferable to gain experience deploying systems using Ignition so
+that they can be re-provisioned easily if needed. This will all be part
+of a "migrating from Fedora Atomic Host" guide which will be published as
+soon as an actual version of Fedora CoreOS is available.
```
bgilbert: An actual FCOS version is available, so this is no longer accurate.

dustymabe: ok modified it to be less specific

dustymabe: ok modified it to be less specific

keithy:
The reason that I raised this issue was the process of deciding whether to wait and rebase my exisitng Fedora Atomic server, or take the plunge with a new install. Sincethe above text still says "may be", I am none the wiser.

bgilbert:
@keithy There isn't a really great answer right now.  We don't know whether rebasing will be feasible, but it will certainly not be the preferred approach.  On the other hand, Fedora CoreOS is in preview and shouldn't be used for production workloads.

PR #9, opened 2019-08-30
Boot docs

ajeddeloh:
Very much a WIP, wanted to throw these up before I went on vacation


ashcrow:
Is this still planned to be done?

ajeddeloh:
I'm not going to get to this before I leave, it's vrey out of date now anyway.

dustymabe:
looks like some good stuff in there. would be nice if somebody picked it up

jlebon:
Yeah, agreed. Filed https://github.com/coreos/fedora-coreos-docs/issues/24 so we don't forget about it.

PR #10, opened 2019-10-01
scripts: fix interpreter

lucab:
The scripts in here use bash-specific extensions. This fixes the
shebang to match that.


lucab:
(I'd need somebody to merge this for me, as I don't have the merge button here)

PR #11, opened 2019-10-01
refs: add platforms docs

lucab:
This introduces a "Platforms" reference page to show how the platform ID
is used, and to keep track of all known platforms.

### DIFF
--- a/modules/ROOT/pages/platforms.adoc
+++ b/modules/ROOT/pages/platforms.adoc
@@ -1,0 +1,31 @@
+= Platforms
+
+Fedora CoreOS is provisioned via prebuilt disk images, and configured on first-boot via https://github.com/coreos/ignition[Ignition]. Each platform may require specific logic and components, thus dedicated images are provided for each supported environment. Additionally, a unique platform ID is available in the host environment for runtime introspection.
+
+== Runtime introspection
+
+Each Fedora CoreOS image boots with a platform-specific identifier, available on the kernel command-line. The name of the parameter is `ignition.platform.id` and the list of supported platform is documented here.
+
+Platform ID can be introspected at runtime, as follows:
+
+.CLI example of platform introspection
+[source, bash]
+----
+$ grep -o ignition.platform.id='[[:alnum:]]*' /proc/cmdline
+
+ignition.platform.id=aws
+----
+
+Platform ID is consumed by OS components such as https://github.com/coreos/ignition[Ignition] and https://github.com/coreos/afterburn[Afterburn]. Additionally, it can be used in systemd units via https://www.freedesktop.org/software/systemd/man/systemd.unit.html#ConditionKernelCommandLine=[`ConditionKernelCommandLine=`].
+
+== Well-known IDs
+
+Here is a list of all supported platforms and their identifier:
+
+ * `aws`: Amazon Web Services (cloud platform)
+ * `azure`: Microsoft Azure (cloud platform)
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/platforms.adoc
```diff
@@ -0,0 +1,31 @@
...
+$ grep -o ignition.platform.id='[[:alpha:]]*' /proc/cmdline
+
+ignition.platform.id=aws
+----
+
+Platform ID is consumed by OS components such as https://github.com/coreos/ignition[Ignition] and https://github.com/coreos/afterburn[Afterburn]. Additionally, it can be used in systemd units via https://www.freedesktop.org/software/systemd/man/systemd.unit.html#ConditionKernelCommandLine=[`ConditionKernelCommandLine=`].
+
+== Well-known IDs
+
+Here below is a list of all supported platforms and their identifier:
```
dustymabe: I would remove "below"

modules/ROOT/pages/platforms.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+ * `aws`: Amazon Web Services (cloud platform)
+ * `azure`: Microsoft Azure (cloud platform)
+ * `gcp`: Google Cloud Platform (cloud platform)
+ * `metal`: bare metal with BIOS or UEFI boot
+ * `openstack`: OpenStack (cloud platform)
+ * `qemu`: QEMU (hypervisor)
```
dustymabe: Should we mention KVM here at all?

dustymabe: Maybe `QEMU/KVM (hypervisor)`

lucab: May I ask why you want to mention KVM here?
Technically, I don't think we depend on KVM in any way. I think (but never tried) that images would also work with software-emulation only. To the best of my knowledge, on this platform we only depend on `fw_cfg`.

dustymabe: Mostly because that's the stack that's typically used and the wording people are familiar with. I'm not married to it.

modules/ROOT/pages/platforms.adoc
```diff
@@ -0,0 +1,31 @@
...
+== Runtime introspection
+
+Each Fedora CoreOS image boots with a platform-specific identifier, available on the kernel command-line. The name of the parameter is `ignition.platform.id` and the list of supported platform is documented here.
+
+Platform ID can be introspected at runtime, as follows:
+
+.CLI example of platform introspection
+[source, bash]
+----
+$ grep -o ignition.platform.id='[[:alpha:]]*' /proc/cmdline
```
ajeddeloh: We had ec2 in the past and I think having numbers in platform id's isn't inherently bad, so maybe `alnum` instead of `alpha`?

modules/ROOT/pages/platforms.adoc
lucab: Ok, then if you don't mind I'll stick to the current wording. `QEMU/KVM` is a subset of this, so it is implicitly covered too.

lucab:
Rebased. I'd need somebody to merge this, as I don't have the button for this repo.

lucab:
@arithx at some point later on I'd like to grow a "Platform-specific notes" section in here to document the baseline/quirks. I think I'll need your brain for the existing ones :)

PR #12, opened 2019-10-02
pages/platforms: Add aliyun to platform list

ashcrow:
See: https://github.com/coreos/fedora-coreos-tracker/issues/276


ashcrow:
cc @coreos/fedora-coreos-docs

PR #15, opened 2019-10-15
pages/platforms: Add ibmcloud to platform list

ashcrow:
See: https://github.com/coreos/fedora-coreos-tracker/issues/277

### DIFF
--- a/modules/ROOT/pages/platforms.adoc
+++ b/modules/ROOT/pages/platforms.adoc
@@ -26,6 +26,7 @@ Here is a list of all supported platforms and their identifier:
  * `aws`: Amazon Web Services (cloud platform)
  * `azure`: Microsoft Azure (cloud platform)
  * `gcp`: Google Cloud Platform (cloud platform)
+ * `ibmcloud`: IBM Cloud (cloud platform)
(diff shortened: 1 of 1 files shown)


ajeddeloh:
LGTM though maybe we want to add something to these showing how supported it currently is? Like aliyun is in development, not ready for end users.

ashcrow:
@ajeddeloh I think that would be a good addition. I'm keen to add a new column in a follow up PR.

lucab:
My preference would be to keep this page as a plain reference list, and detail any caveats/unsupported bits on each relevant "Booting on XYZ" pages (which we should start writing soon-ish).

ashcrow:
This PR is ready to merge if it is :+1:

ashcrow:
bump :slightly_smiling_face:

lucab:
I took my own dive into `ibmcloud` (notes [here](https://github.com/coreos/fedora-coreos-tracker/issues/277#issuecomment-545916088)).
I do fear we have to take additional details into here as the platform is really a 3-in-1 beast with different semantics and it's not clear to me which part of that we are targeting.

ashcrow:
@lucab I'll verify which sub environment it is today

PR #18, opened 2019-12-03
pages/auto-updates: add section on configuring rollout_wariness

dustymabe:


modules/ROOT/pages/auto-updates.adoc
```diff
@@ -7,6 +7,38 @@ By default, the OS performs continuous auto-updates via two components:
...
+storage:
+  files:
+    - path: /etc/zincati/config.d/51-rollout-wariness.toml
+      contents:
+        inline: |
+          [identity]
+          rollout_wariness = 0.0
+----
+
+The FCC can then be converted to an Ignition config by using FCCT:
```
lucab: I would drop this and the shell-line below. It's already covered in the getting started and we don't want to duplicate it for every single snippet we show. Longer we should have a single page showing FCCT usage, and possible an inline FCCT->Ignition converter.

modules/ROOT/pages/auto-updates.adoc
```diff
@@ -7,6 +7,38 @@ By default, the OS performs continuous auto-updates via two components:
  * https://github.com/coreos/rpm-ostree[rpm-ostree] handles multiple on-disk OSTree deployments, and it can switch between them at boot-time.
  * https://github.com/coreos/zincati[Zincati] continuously checks for OS updates and applies them via rpm-ostree.
 
+== Configuring Update/Rollout Wariness
+
+The Zincati service runs on a Fedora CoreOS node and checks with the remote Cincinnati service to see when updates are available. As part of the request it can send along a configuration value known as `rollout_wariness` (see https://github.com/coreos/zincati/blob/master/docs/usage/auto-updates.md#phased-rollouts-client-wariness-canaries[documentation]) to let the server know how eager, or how risk-averse the node is to receiving updates. `rollout_wariness` can be a floating point value between `0.0` (most eager) and `1.0` (most conservative). To configure a node to receive an update as soon as it is available you can set the `rollout_wariness` to `0.0` in a configuration file within `/etc/zincati/config.d/` using https://github.com/coreos/fcct[FCCT]:
```
lucab: Let's use versioned links instead of `master`: https://github.com/coreos/zincati/blob/0.0.6/docs/usage/auto-updates.md#phased-rollouts-client-wariness-canaries

lucab: `It can be configured with a custom "rollout wariness" value to let the server...`

lucab: This is for users, so I'd rather not point them at `0.0`. Zincati docs explicitly recommends not doing that for normal workloads. It also bypasses our normal rollout window.

My suggestion would be:
```
In order to receive updates very early in the phased rollout cycle, a node can be configured with a low `rollout_wariness` value (e.g. `0.001`) .
This can be done by provisioning the configuration snippet shown below:
```

modules/ROOT/pages/auto-updates.adoc
```diff
@@ -7,6 +7,38 @@ By default, the OS performs continuous auto-updates via two components:
...
+    - name: core
+      ssh_authorized_keys:
+        - ssh-rsa AAA...
+storage:
+  files:
+    - path: /etc/zincati/config.d/51-rollout-wariness.toml
+      contents:
+        inline: |
+          [identity]
+          rollout_wariness = 0.0
```
lucab: I'd prefer showing `0.001` here.

dustymabe: 👍

modules/ROOT/pages/auto-updates.adoc
dustymabe: 👍 I tried to incorporate the suggestion.

dustymabe: +1

dustymabe: I updated the links, but I don't know which is better TBH. I do know that if the docs change then these links will be old and no one will update them.

modules/ROOT/pages/auto-updates.adoc
dustymabe: I dropped the example and linked to the getting started guide

modules/ROOT/pages/auto-updates.adoc
```diff
@@ -17,4 +42,4 @@ The following finalization strategies are available:
  * immediately reboot to apply an update, as soon as it is downloaded and staged locally
  * use an external lock-manager to reboot a fleet of machines in a coordinated way
 
-For further documentation on how to configure updates finalization, check Zincati documentation on https://github.com/coreos/zincati/blob/master/docs/usage/updates-strategy.md[updates strategy].
```
dustymabe: note that I also changed this link to be versioned

modules/ROOT/pages/auto-updates.adoc
lucab: Both have downsides, I agree. The versioned links rationale is mostly to ensure we know *some useful* content is there.

lucab:
@dustymabe can you maybe have a quick look at the permissions of this repo? I think reviewers set is different (and smaller) here compared to all other FCOS ones.

dustymabe:
> @dustymabe can you maybe have a quick look at the permissions of this repo? I think reviewers set is different (and smaller) here compared to all other FCOS ones.

I'm not an admin so I can't see/change any of those settings :(

PR #21, opened 2019-12-20
guides: add sysctl page

lucab:
This adds a user-guide for custom sysctl settings.

### DIFF
--- a/modules/ROOT/pages/sysctl.adoc
+++ b/modules/ROOT/pages/sysctl.adoc
@@ -1,0 +1,23 @@
+= Kernel tunables (sysctl)
+
+The Linux kernel offers a plethora of knobs under `/proc/sys` to control the availability of different features and tune performance parameters.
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/sysctl.adoc
```diff
@@ -0,0 +1,23 @@
+= Kernel tunables (sysctl)
+
+The Linux kernel offers a plethora of knobs under `/proc/sys` to control the availability of different features and tune performance parameters.
+
+Values under `/proc/sys` can be changed directly at runtime, but such changes will not be persisted across reboots.
+Persistent settings should be written under `/etc/sysctl.d/` during provisioning, in order to be appliead on each boot.
```
jlebon: applied

lucab: Good catch, fixed!

PR #26, opened 2020-01-14
bare-metal: use CLI instead of kargs for ISO path

jlebon:
I find using the CLI more elegant. Plus, you get better progress
reporting. Plus it introduces them to the live ISO in a more natural way
(I didn't spell it out, but users can infer from the output that they've
just booted into a working FCOS instance and they can play around as
they want).

### DIFF
--- a/modules/ROOT/pages/bare-metal.adoc
+++ b/modules/ROOT/pages/bare-metal.adoc
@@ -8,7 +8,7 @@ This guide provides instructions to install Fedora CoreOS to bare metal. Two opt
 
 == Prerequisite
 
-Before installing FCOS, you must have an Ignition configuration file. If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File].
+Before installing FCOS, you must have an Ignition configuration file and host it somewhere (e.g. using `python3 -m http.server`). If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File].
 
 == Installing from ISO
 
@@ -18,15 +18,17 @@ To install FCOS onto bare metal with an ISO, follow these steps:
 +
 NOTE: You can install in either legacy boot (BIOS) mode or in UEFI mode, regardless of what mode the OS will use once installed.
 +
-. Burn the ISO to disk and boot it.
-. From the boot menu, press `<TAB>` (isolinux) or `e` (grub) to edit the kernel command line.
-. Add the following parameters to the kernel command line:
-* `coreos.inst.install_dev=/dev/sda`
-* `coreos.inst.stream=stable`
-* `coreos.inst.ignition_url=http://example.com/config.ign`
-. Press `<ENTER>` (isolinux) or `<CTRL-x>` (grub) to start the installation.
+. Burn the ISO to disk and boot it on the target system. The ISO is capable of bringing up a fully functioning FCOS system purely from memory (i.e. without using any disk storage). Once booted, you will have access to a bash command prompt.
+. You can now fetch your Ignition config and run `coreos-installer`:
+[source, bash]
+----
+curl -LO http://example.com/example.ign
+sudo coreos-installer install /dev/sda --ignition example.ign
+----
 
-Once the installation is complete, the machine reboots. After rebooting, the first boot process begins. It is at this time that Ignition ingests the configuration file and provisions the system as specified.
+Once the installation is complete, you can simply `sudo reboot`. After rebooting, the first boot process begins. It is at this time that Ignition ingests the configuration file and provisions the system as specified.
+
+TIP: Check out `coreos-installer install --help` for more options on how to install Fedora CoreOS.
 
 == Installing from PXE
 


jlebon:
Works on top of https://github.com/coreos/fedora-coreos-docs/pull/25.

dustymabe:
yeah I think we discussed this briefly the other day. We should probably see how this fits in with the strategy we wrote down. I do agree this looks nice. Let's talk tomorrow.

dustymabe:
@lucab @bgilbert - in the context of what we discussed the other day, WDYT?

dustymabe:
@jlebon - rebase?

jlebon:
Rebased!

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -19,14 +19,14 @@ To install FCOS onto bare metal with an ISO, follow these steps:
...
-. Press `<ENTER>` (isolinux) or `<CTRL-x>` (grub) to start the installation.
-
-Once the installation is complete, the machine reboots. After rebooting, the first boot process begins. It is at this time that Ignition ingests the configuration file and provisions the system as specified.
+. Once booted, you can fetch your Ignition config and run `coreos-installer`:
+[source, bash]
+----
+curl -LO http://example.com/example.ign
+sudo coreos-installer install /dev/sda --ignition example.ign
+----
+
```
dustymabe: It might be worth adding a **Tip** for running `coreos-install install --help` to explore more options that can be passed to the installer.

dustymabe:
One possible problem with this approach is the fact that it doesn't carry forward the 'networking kargs' into the first boot, but I think it's worth exploring. I do like this approach as it seems simple and it does encourage the user to use the live system. Ultimately, though, we should probably document both ways.

jlebon:
Just to put this in more context, this page is the first introduction to installing on bare metal. Before going the full automated, bells-and-whistles path (with high upfront cost), we need the vanilla/manual path (with low upfront cost) that most users will be willing to work through to try out.

> One possible problem with this approach is the fact that it doesn't carry forward the 'networking kargs' into the first boot, but I think it's worth exploring.

Yeah, we definitely need to document this, and the fully-automated-via-Ignition case. But I don't think we should jump straight to that.

miabbott:
> Yeah, we definitely need to document this, and the fully-automated-via-Ignition case. But I don't think we should jump straight to that.

Do we have a running list of other things like this we'd like to get documented?

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -19,14 +19,16 @@ To install FCOS onto bare metal with an ISO, follow these steps:
...
 . Burn the ISO to disk and boot it.
-. From the boot menu, press `<TAB>` (isolinux) or `e` (grub) to edit the kernel command line.
-. Add the following parameters to the kernel command line:
-* `coreos.inst.install_dev=/dev/sda`
-* `coreos.inst.stream=stable`
-* `coreos.inst.ignition_url=http://example.com/config.ign`
-. Press `<ENTER>` (isolinux) or `<CTRL-x>` (grub) to start the installation.
-
-Once the installation is complete, the machine reboots. After rebooting, the first boot process begins. It is at this time that Ignition ingests the configuration file and provisions the system as specified.
+. Once booted, you can fetch your Ignition config and run `coreos-installer`:
```
dustymabe: I think maybe we should indicate a bit more about what is going on.

*The machine will boot and run Fedora CoreOS live from memory and give you a bash prompt. No hard disks are mounted so you can interactively fetch your ignition config and run coreos-installer:

lucab:
From my notes, I think we wanted to keep these cases:
 1. simple case via kargs (same node)
 2. simple case via container (for arbitrary disks/nodes)
 3. complex case via Ignition

If we keep kargs and we add this on top, we have 1&2 covered. Thus, I'd probably rework this PR to introduce two sections, and have content in both.
(We can show direct CLI invocation or via podman, as you wish)

dustymabe:
> From my notes, I think we wanted to keep these cases:
> 
>     1. simple case via kargs (same node)
> 
>     2. simple case via container (for arbitrary disks/nodes)
> 
>     3. complex case via Ignition
> 
> 
> If we keep kargs and we add this on top, we have 1&2 covered. Thus, I'd probably rework this PR to introduce two sections, and have content in both.
> (We can show direct CLI invocation or via podman, as you wish)

Slight proposed change:

We only document one way to install here and create a dedicated page for "other ways to install". Here we document the "interactive coreos-installer via Live ISO" method. On the other page we document:

1. simple case via kargs (same node)
    - for Live ISO
    - for PXE (automation)
2. simple case via container (for arbitrary disks/nodes)
3. complex case via Ignition

jlebon:
> We only document one way to install here and create a dedicated page for "other ways to install". Here we document the "interactive coreos-installer via Live ISO" method. On the other page we document:

Yup, makes sense to me overall.

What I'm questioning is the value of documenting the kargs via live ISO path. If it's not automated, then why not use the CLI? Or is it possible to use the live ISO in an automated fashion in some situations? If so, we should just document the kargs in that context only (automation), just like PXE.

dustymabe:
> Or is it possible to use the live ISO in an automated fashion in some situations? If so, we should just document the kargs in that context only (automation), just like PXE.

I think it is automatable via tools like virt-install but I think you're right. We should just make the "topic" of this section about automation and mention PXE as the primary use case and leave it open to other use cases as well.

modules/ROOT/pages/bare-metal.adoc
dustymabe: This extra clarification is probably useful for users who have never booted a live ISO before and might not understand what it is.

jlebon: I took your suggestion in spirit, but tweaked it. ;)  I think it's probably worth a "live system" section in the future we can link to on more information about that.

PR #27, opened 2020-01-15
nav.adoc: fix broken link to other adoc files

zonggen:
Not sure if this is intentional but the links to the adoc files are broken..
Feel free to close this if the behavior is indeed intentional.

Signed-off-by: Allen Bai <abai@redhat.com>


jlebon:
Hmm, that's weird. It works fine for me (and online too: https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/). Did you follow the steps in https://github.com/coreos/fedora-coreos-docs#local-preview?

zonggen:
@jlebon My apologies, didn't notice it was built and run inside a container. Now it's working.

Closing

PR #28, opened 2020-01-20
getting-started: provide path for running without Ignition config

jlebon:
To lower the barrier a bit more on trying FCOS, tweak the AWS docs to
clarify that one doesn't *have* to specify an Ignition config. Provide
an example using `--key-name`.

There's also the live image... we should document that too at some
point.

This came out of feedback from a Reddit comment:
https://www.reddit.com/r/Fedora/comments/epyzbx/fedora_coreos_out_of_preview/feoysmk

### DIFF
--- a/modules/ROOT/pages/getting-started.adoc
+++ b/modules/ROOT/pages/getting-started.adoc
@@ -1,9 +1,9 @@
 :experimental:
 = Fedora CoreOS - Getting Started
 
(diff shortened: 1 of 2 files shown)


lucab:
LGTM. At some point however we should start growing cloud-specific pages, and leave the getting-started to a bare minimum "ignition snippet plus node start".

PR #34, opened 2020-01-28
getting-started: various tweaks and enhancements

jlebon:
1. Consistently use "FCOS".
2. Drop `chown core:kvm` and instead add a tip about having access to
   `/dev/kvm`.
3. Add `-snapshot` to the QEMU command so people don't mess up the
   golden image they just downloaded.
4. Add a second QEMU example that creates a new qcow2 for persistent
   storage.
5. Similarly, tweak the `virt-install` command to use the downloaded
   qcow2 as a backing file, not the main disk directly.
6. Simplify the `qemu` and `virt-install` commands. E.g. using
   `qemu-kvm`, and dropping args which already have reasonable defaults.
   We should detail more complex args on a separate page if wanted.

### DIFF
--- a/modules/ROOT/pages/getting-started.adoc
+++ b/modules/ROOT/pages/getting-started.adoc
@@ -11,17 +11,17 @@ NOTE: For more information on configuration, refer to the documentation for xref
 
 To launch FCOS on AWS:
 
-. Go to the https://getfedora.org/coreos/download/[download page] to identify the correct AMI.
+. Identify the correct AMI image ID for your region from https://getfedora.org/coreos/download/[the download page]
 
-. Specify the Ignition configuration file as the https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-add-user-data[user-data]. You can skip this step, but you will want to specify a key using `--key-name` so that you can still access the node. This provides an easy way to test out FCOS without first creating an Ignition configuration.
+. Launch the VM using `aws ec2 run-instances`. The Ignition file can be passed to the VM as its https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-add-user-data[user data]. You can skip this step if you just want SSH access; simply pass a https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-key-pairs.html[registered key name] via `--key-name`. This provides an easy way to test out FCOS without first creating an Ignition configuration.
 
-.Example launching Fedora CoreOS on AWS using an Ignition configuration file
+.Example launching FCOS on AWS using an Ignition configuration file
 [source, bash]
 ----
 aws ec2 run-instances <other options> --image-id <ami> --user-data file://example.ign
 ----
 
-.Example launching Fedora CoreOS on AWS using a registered SSH key pair
+.Example launching FCOS on AWS using a registered SSH key pair
 [source, bash]
 ----
 aws ec2 run-instances <other options> --image-id <ami> --key-name my-key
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/getting-started.adoc
```diff
@@ -11,61 +11,73 @@ NOTE: For more information on configuration, refer to the documentation for xref
...
-. Ensure the correct ownership of the file:
-+
-`chown core:kvm fedora-coreos-qemu.qcow2`
+[source, bash]
+----
+unxz fedora-coreos-qemu.qcow2.xz
+----
 +
-TIP: The above example assumes that you have created the user named `core` on your local machine. If you did not, replace `core` with the user that will boot FCOS on the local host.
+TIP: Make sure that your user has access to `/dev/kvm` either by doing `chmod 666 /dev/kvm` (this matches modern distros), or if `/dev/kvm` is owned by the `kvm` group, adding yourself to that group with `usermod -aG kvm $USER`.
```
lucab: Nit: I'd drop the `chmod` and just keep the reference to the device and the group.

modules/ROOT/pages/getting-started.adoc
```diff
@@ -11,61 +11,73 @@ NOTE: For more information on configuration, refer to the documentation for xref
...
 
-. Specify the Ignition file to the hypervisor. In the case of QEMU, this is done with the `-fw_cfg` parameter, which sets the `opt/com.coreos/config` key in the QEMU firmware configuration device.
+. Launch the VM using `qemu-kvm` or `virt-install`. The Ignition file is passed to the VM via the `-fw_cfg` parameter, which sets the `opt/com.coreos/config` key in the QEMU firmware configuration device. For libvirt, we must pass this through the `--qemu-commandline` argument.
```
lucab: s/we/you/

modules/ROOT/pages/getting-started.adoc
```diff
@@ -11,61 +11,73 @@ NOTE: For more information on configuration, refer to the documentation for xref
 
 To launch FCOS on AWS:
 
-. Go to the https://getfedora.org/coreos/download/[download page] to identify the correct AMI.
+. Identify the correct AMI image ID for your region from https://getfedora.org/coreos/download/[the download page]
 
-. Specify the Ignition configuration file as the https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-add-user-data[user-data]. You can skip this step, but you will want to specify a key using `--key-name` so that you can still access the node. This provides an easy way to test out FCOS without first creating an Ignition configuration.
+. Launch the VM using `aws ec2 run-instances`. The Ignition file can be passed to the VM as its https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-add-user-data[user data]. You can skip this step, but you will want to specify a key using `--key-name` so that you can still access the node. This provides an easy way to test out FCOS without first creating an Ignition configuration.
```
miabbott: Might be worth mentioning that the key used by `--key-name` has to be previously imported into AWS and is not a local file.

jlebon: Added!

mike-nguyen:
LGTM

modules/ROOT/pages/getting-started.adoc
```diff
@@ -11,61 +11,73 @@ NOTE: For more information on configuration, refer to the documentation for xref
...
 	-fw_cfg name=opt/com.coreos/config,file=path/to/example.ign
 ----
 +
 .Example launching FCOS with virt-install
 [source, bash]
 ----
-virt-install -n fcos --vcpus 2 -r 2048 --os-variant=fedora31 --import --network bridge=virbr0 --disk=/var/lib/libvirt/images/fedora-coreos-30.20190905.0-qemu.qcow2,format=qcow2,bus=virtio --noautoconsole --qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=/path/to/example.ign"
+virt-install -n fcos -r 2048 --os-variant=fedora31 --import --noautoconsole \
+	--disk size=10,backing_store=/path/to/fedora-coreos-qemu.qcow2 \
+	--qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=/path/to/example.ign"
```
dustymabe: let's remove `--noautoconsole` and add `--graphics=none` so we will get attached to a serial console

modules/ROOT/pages/getting-started.adoc
```diff
@@ -11,61 +11,73 @@ NOTE: For more information on configuration, refer to the documentation for xref
...
+qemu-kvm -m 2048 -cpu host -nographic \
+	-drive if=virtio,file=my-fcos-vm.qcow2 \
 	-fw_cfg name=opt/com.coreos/config,file=path/to/example.ign
 ----
 +
 .Example launching FCOS with virt-install
 [source, bash]
 ----
-virt-install -n fcos --vcpus 2 -r 2048 --os-variant=fedora31 --import --network bridge=virbr0 --disk=/var/lib/libvirt/images/fedora-coreos-30.20190905.0-qemu.qcow2,format=qcow2,bus=virtio --noautoconsole --qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=/path/to/example.ign"
+virt-install -n fcos -r 2048 --os-variant=fedora31 --import --noautoconsole \
```
dustymabe: let's remove `--noautoconsole` and add `--graphics=none` so we will get attached to a serial console

modules/ROOT/pages/getting-started.adoc
dustymabe: if run as a normal user I think this will use user mode networking by default, which means the `ssh core@<ip address>` from below won't work

modules/ROOT/pages/getting-started.adoc
```diff
@@ -11,61 +11,73 @@ NOTE: For more information on configuration, refer to the documentation for xref
...
+----
 +
-TIP: The above example assumes that you have created the user named `core` on your local machine. If you did not, replace `core` with the user that will boot FCOS on the local host.
+TIP: Make sure that your user has access to `/dev/kvm` (this should already be the case on modern distros), or if `/dev/kvm` is owned by the `kvm` group, adding yourself to that group with `usermod -aG kvm $USER`.
 
-. Specify the Ignition file to the hypervisor. In the case of QEMU, this is done with the `-fw_cfg` parameter, which sets the `opt/com.coreos/config` key in the QEMU firmware configuration device.
+. Launch the VM using `qemu-kvm` or `virt-install`. The Ignition file is passed to the VM via the `-fw_cfg` parameter, which sets the `opt/com.coreos/config` key in the QEMU firmware configuration device. For libvirt, you must pass this through the `--qemu-commandline` argument.
 +
-You must use absolute file paths here. QEMU cannot resolve relative file paths for these options.
+.Example launching FCOS with QEMU (no persistent storage)
```
dustymabe: is there a better way to say this? `no persistent storage` is misleading in this context IMHO. When I hear "no persistent storage" I think of a live/ephemeral system, which is not the case here. If I reboot my changes will still be there.

This is more of a "create snapshot and boot into snapshot" than it is "no persistent storage".

modules/ROOT/pages/getting-started.adoc
jlebon: I personally rarely not use `--noautoconsole` since for libvirt VMs you usually don't authenticate over serial with a password, but via SSH. Though it's useful for FCOS since it prints the IP address for you.

modules/ROOT/pages/getting-started.adoc
jlebon: It will use `--network default`, which is more generic than `bridge=virbr0`. AFAIK, user-mode networking is only if you're using `qemu:///session`.

modules/ROOT/pages/getting-started.adoc
jlebon: I tweaked this bit!

modules/ROOT/pages/getting-started.adoc
dustymabe: > user-mode networking is only if you're using `qemu:///session`.

right, which I think is becoming more common. `vagrant-libvirt` in Fedora now uses `qemu:///session` by default. I think we should probably either be more explicit here or possibly add a **TIP:** that talks about the differing behavior in the different modes.

modules/ROOT/pages/getting-started.adoc
dustymabe: Unfortunately I still think *temporary* is the wrong term. It's not actually temporary either. Maybe we should concentrate on the "snapshot" terminology. I'll grab you in IRC.

dustymabe: ok, I was a bit mistaken here as I thought this was the heading for the `virt-install --disk backing_store=` invocation. This is resolved.

modules/ROOT/pages/getting-started.adoc
jlebon: Yeah you're right. I unset `LIBVIRT_DEFAULT_URI` inherited from my bash config, and indeed `virsh`/`virt-install` defaults to the user daemon. I just switched this now to use `--connection qemu:///system`.

PR #35, opened 2020-01-29
Various fcct usage updates

bgilbert:
The last commit depends on https://github.com/coreos/fcct/pull/72 and the second-to-last one should probably not happen without https://github.com/coreos/fcct/pull/70.

### DIFF
--- a/modules/ROOT/pages/producing-ign.adoc
+++ b/modules/ROOT/pages/producing-ign.adoc
@@ -47,11 +47,11 @@ An easy way to use `fcct` is to run it in a container with `podman`:
 
 . Pull the container for `fcct`:
 +
-`podman pull quay.io/coreos/fcct:v0.2.0`
+`podman pull quay.io/coreos/fcct:release`
 +
 . Run fcct on the FCC file:
 +
-`podman run -i --rm quay.io/coreos/fcct:v0.2.0 -pretty -strict < example.fcc > example.ign`
+`podman run -i --rm quay.io/coreos/fcct:release -pretty -strict < example.fcc > example.ign`
 +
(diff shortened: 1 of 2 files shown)


lucab:
It looks like the first two commits make sense on their own, and would be good to merge quickly to prepare for the rest of changes (esp. the single-dash breaking).

Should we maybe split this PR in order to land the first part right now? Other than that, LGTM.

modules/ROOT/pages/using-fcct.adoc
```diff
@@ -15,24 +15,22 @@ You can run `fcct` as a container with docker or podman or download it as a stan
 
 This example uses podman, but you can use docker in a similar manner.
 
-In this example, substitute `v0.2.0` with the desired version, as the `latest` tag corresponds with `master` and not with the latest release.
```
lucab: Nit: maybe add an explicit note that `release` always tracks latest release, and `latest` tracks git development?

bgilbert:
Updated.  The three commits in this PR should be good to merge now, and the last commit has been moved to https://github.com/coreos/fedora-coreos-docs/pull/38.

modules/ROOT/pages/producing-ign.adoc
```diff
@@ -51,7 +51,7 @@ An easy way to use `fcct` is to run it in a container with `podman`:
 +
```
jlebon: Optional: a link back to https://github.com/coreos/fcct/pull/70 in this commit's message for posterity would be good.

(Grrr... on GitHub for not allowing commenting on a commit message. I'm talking about 8f4c3f200768662a61d0ef3481d1438d780fe25c here.)

bgilbert:
Updated!

PR #37, opened 2020-01-29
platforms: allocate ID for vultr

lucab:
This adds `vultr` as the platform ID for Vultr cloud provider.

Homepage: https://www.vultr.com/
Metadata: https://www.vultr.com/metadata
User data: https://web.archive.org/web/20190513194756/https://www.vultr.com/metadata/#user


lucab:
This is ready to go.

/cc @dustymabe @jlebon

PR #39, opened 2020-02-04
migrate-ah: tweak wording for translating configs instruction

dustymabe:
I believe "Manually configure" is a little less clear about
what needs to be done. Where "This can be manually translated"
seems to be a little better.

### DIFF
--- a/modules/ROOT/pages/migrate-ah.adoc
+++ b/modules/ROOT/pages/migrate-ah.adoc
@@ -23,7 +23,7 @@ ssh_authorized_keys:
   - ssh-rsa ...
 ----
 
-Manually configure with a xref:ign-passwd.adoc[`passwd`] node in the FCC file:
+This can be manually translated into a xref:ign-passwd.adoc[`passwd`] node within the FCC file:
 
 .Example of users:
 [source, yaml]


modules/ROOT/pages/migrate-ah.adoc
```diff
@@ -23,7 +23,7 @@ ssh_authorized_keys:
   - ssh-rsa ...
 ----
 
-Manually configure with a xref:ign-passwd.adoc[`passwd`] node in the FCC file:
+This can be manually translated into a xref:ign-passwd.adoc[`passwd`] node within the FCC file:
```
jlebon: Minor: Hmm, not sure if we should even refer to `ign-passwd.adoc`. We mention higher up already that the conversion should be to FCCs and link to that specification, so I think it'd be less confusing to just say something like

> The equivalent FCC for this would be:

?

dustymabe: if you click on the generated link: [passwd](https://github.com/coreos/fedora-coreos-docs/blob/master/modules/ROOT/pages/ign-passwd.adoc) that document mostly describes the options and then gives example FCC configs. Is the link really not useful?

jlebon: Ohh I see, I thought the `ign-` nodes documented Ignition instead of FCC. But that's not the case (well, it does also document Ignition since those bits match, but the examples are of FCC).

jlebon: I wonder if those files should be named `fcc-<...>.adoc` instead?

dustymabe: maybe :)

PR #44, opened 2020-02-12
faq: add blurb about uploading to private AWS regions

jlebon:
This is mostly adapted from the RHCOS FAQ:
https://github.com/openshift/os/blob/master/FAQ.md

But it applies equally well to FCOS:
https://github.com/coreos/fedora-coreos-tracker/issues/383

### DIFF
--- a/modules/ROOT/pages/faq.adoc
+++ b/modules/ROOT/pages/faq.adoc
@@ -193,3 +193,19 @@ Kubernetes cluster, you will be able to use upcoming tools to coordinate
 updates and reboots.
 
 https://discussion.fedoraproject.org/t/launch-faq-how-do-i-coordinate-cluster-wide-os-updates-is-locksmith-or-the-container-linux-update-operator-available-for-fedora-coreos/56[discuss]
+
+== How do I upload Fedora CoreOS to private AWS EC2 regions?
+
+Fedora CoreOS today is only uploaded to the standard AWS regions. For regions
+in other AWS partitions like GovCloud and AWS China, you must upload the images
+yourself.
+
+Note that Fedora CoreOS uses a unified BIOS/UEFI partition layout. As such, it
+is not compatible with the `aws ec2 import-image` API (for more information,
+see https://github.com/openshift/os/pull/396[related discussions]). Instead,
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/faq.adoc
```diff
@@ -193,3 +193,20 @@ Kubernetes cluster, you will be able to use upcoming tools to coordinate
 updates and reboots.
 
 https://discussion.fedoraproject.org/t/launch-faq-how-do-i-coordinate-cluster-wide-os-updates-is-locksmith-or-the-container-linux-update-operator-available-for-fedora-coreos/56[discuss]
+
+== How do I upload Fedora CoreOS to private AWS EC2 regions?
+
+Fedora CoreOS today is only uploaded to the public AWS regions. For private
+regions like GovCloud and AWS China, you must upload the images yourself.
```
bgilbert: s/public/standard/
`For regions in other AWS partitions like GovCloud and AWS China`...

dustymabe:
This is great! Do you think it would be better if we had a real world example of running those commands with an FCOS release?

modules/ROOT/pages/faq.adoc
```diff
@@ -193,3 +193,20 @@ Kubernetes cluster, you will be able to use upcoming tools to coordinate
...
+
+Note that Fedora CoreOS uses a unified BIOS/UEFI partition layout. As such, it
+is not compatible with the default `aws ec2 import-image` API (for more
```
bgilbert: How is it the default?  Can we drop the word `default`?

jlebon: It's the "default" insofar as it's what I understand a more common workflow than `import-snapshot` for people who want to create AMIs from local disk images. But not against dropping the word entirely. :)

jlebon:
> This is great! Do you think it would be better if we had a real world example of running those commands with an FCOS release?

Thanks! I think having commands would make sense. Though I wouldn't feel comfortable typing those up without trying them out myself, and I don't have the cycles to try it out now. We can probably leave it as a follow-up?

PR #46, opened 2020-02-13
bare-metal: document live ISO memory requirement

lucab:
This documents the minimum memory requirement for booting a live ISO
system. The current actual usage is slightly larger than 2 GiB, so
I just round up to the closest integer for a reasonable margin.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/388

### DIFF
--- a/modules/ROOT/pages/bare-metal.adoc
+++ b/modules/ROOT/pages/bare-metal.adoc
@@ -3,26 +3,26 @@
 
 This guide provides instructions to install Fedora CoreOS to bare metal. Two options are available:
 
-* Installing from ISO
+* Installing from live ISO
 * Installing from PXE
 
 == Prerequisite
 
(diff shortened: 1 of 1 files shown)


bgilbert:
1 GB is expected to be too small.  If 2 GB isn't enough, though, we should make sure we understand why.

bgilbert:
Now that I think about it, it's not obvious that 1 GB shouldn't work either.

lucab:
It looks like we'd prefer holding this for a bit then.

dustymabe:
> The current actual usage is slightly larger than 2 GiB, so
> I just round up to the closest integer for a reasonable margin.

I did notice recently on a VM with 2048 MiB that I was getting errors during bootup because it was running out of disk space, which means it was running out of RAM.

dustymabe:
We should investigate "why" we need more than we think we do, but since the siutation isn't improving right now let's merge this and lower the recommendation when we figure out what the problem is.

dustymabe:
> We should investigate "why" we need more than we think we do, 

opened https://github.com/coreos/fedora-coreos-tracker/issues/407 for this

dustymabe:
merging since no one else has chimed in

PR #48, opened 2020-02-14
Add some troubleshooting steps for access recovery

jlebon:
The `init=/bin/sh` trick is known, but let's give a clear step-by-step
to make this easier. (The SELinux step in particular is easy to miss.)

See related user question:
https://discussion.fedoraproject.org/t/recommended-password-recovery-procedure/17034

### DIFF
--- a/modules/ROOT/pages/access-recovery.adoc
+++ b/modules/ROOT/pages/access-recovery.adoc
@@ -1,0 +1,10 @@
+= Access Recovery
+
+If you've lost the private key of an SSH keypair used to log into Fedora CoreOS, and do not have any password logins set up to use at the console, you can gain access back to the machine using the `init=/bin/sh` trick:
+
+. When booting the system, intercept the GRUB menu and edit the entry to append `init=/bin/sh` to the kernel argument list, then press Ctrl-X to resume booting.
+. Once the system has booted into a shell prompt, load the SELinux policy with `/sbin/load_policy -i`. If you miss this step before running `passwd`, you'll want to run `/sbin/restorecon -v /etc/{passwd,shadow}` before rebooting.
+. Set or reset the password for the target user using the `passwd` utility.
+. And finally, reboot the system with `/sbin/reboot -f`.
+
+You should now be able to log back into the system at the console. From there, you can e.g. fetch a new public SSH key to add to `~/.ssh/authorized_keys` and delete the old one. You may also want to lock the password you've set (using `passwd -l`). Note that Fedora CoreOS by default does not allow SSH login via password authentication.
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -15,5 +15,6 @@
 *** xref:auto-updates.adoc[Auto-Updates]
 ** Troubleshooting
 *** xref:manual-rollbacks.adoc[Manual Rollbacks]
+*** xref:access-recovery.adoc[Access Recovery]
 * Reference pages
  ** xref:platforms.adoc[Platforms]
--- a/modules/ROOT/pages/migrate-ah.adoc
+++ b/modules/ROOT/pages/migrate-ah.adoc
@@ -37,6 +37,8 @@ passwd:
       groups: [ sudo, docker ]
 ----
 
+NOTE: Fedora CoreOS disables password login over SSH by default. It is strongly recommended to only use key authentication. Setting passwords can be useful however for logging into the console directly.
+
 == Converting storage definitions
 
 With FAH, you could configure additional storage for the system with either `cloud-init` or  `docker-storage-setup` via the `/etc/sysconfig/docker-storage-setup` file. With FCOS, you should configure additional storage at provisioning time via Ignition in the xref:ign-storage.adoc[`storage`] node of the FCC file.
--- a/modules/ROOT/pages/migrate-cl.adoc
+++ b/modules/ROOT/pages/migrate-cl.adoc
@@ -14,7 +14,7 @@ To migrate from CL to FCOS, you must convert your old Container Linux Configs, I
 The following changes have been made to the installation process:
 
 * The `coreos-install` script has been replaced with https://github.com/coreos/coreos-installer[`coreos-installer`]. It offers similar functionality.
-* The `coreos.autologin` kernel command-line parameter is not currently supported in FCOS.
+* The `coreos.autologin` kernel command-line parameter is not currently supported in FCOS. For access recovery purposes, there are instructions available xref:access-recovery.adoc[here].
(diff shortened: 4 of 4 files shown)


lucab:
@jlebon can we maybe remove all the password-setting references and just stick to show the SSH flow? I'm worried users could end with forgotten weak temporary passwords.

Also, I'm somehow surprised that init=sh still manages to boot into rootfs given all the missing things from systemd and rpm-ostree. Did you try this on a recent FCOS?

jlebon:
> @jlebon can we maybe remove all the password-setting references and just stick to show the SSH flow? I'm worried users could end with forgotten weak temporary passwords.

I think the only places it's showing up right now are where it makes sense. Once in the specifications for FCC, and once about migrating from AH: https://docs.fedoraproject.org/en-US/fedora-coreos/migrate-ah/. Hmm, I guess we should probably add a warning there though about SSH password logins being disabled by default and discouraged?

> Also, I'm somehow surprised that init=sh still manages to boot into rootfs given all the missing things from systemd and rpm-ostree. Did you try this on a recent FCOS?

Bash doesn't need systemd nor rpm-ostree though :)  It's functional enough to at least reset passwords and do other basic things.

austinnichols101:
I just tried init=/bin/sh with FCOS 31 and it just hangs at boot.  Possibly I am not following the correct procedure...

image

jlebon:
@austinnichols101 What message does it hang on?

I'll admit I tested it on a local build. Will sanity-check against the latest release.

austinnichols101:
@jlebon Here's the last of the boot log (running VMware Workstation 15).

image

modules/ROOT/pages/access-recovery.adoc
```diff
@@ -0,0 +1,10 @@
+= Access Recovery
+
+If you've lost the private key of an SSH keypair used to log into Fedora CoreOS, and do not have any password logins set up to use at the console, you can gain access back to the machine using the `init=/bin/sh` trick:
+
+. Launch the system using `init=/bin/sh`
```
jamescassell: Needs further explanation. (I understand, but it's best to be as clear as possible.)

austinnichols101:
@jlebon - here's an MP4 video of the boot failure.
[RCOS Boot 2020-02-14_12-59-32.zip](https://github.com/coreos/fedora-coreos-docs/files/4206002/RCOS.Boot.2020-02-14_12-59-32.zip)

modules/ROOT/pages/access-recovery.adoc
austinnichols101: Procedure should also be linked from this page:
 https://docs.fedoraproject.org/en-US/fedora-coreos/migrate-cl/ 

e.g.
```
The coreos.autologin kernel command-line parameter is not currently supported in FCOS.
See modules/ROOT/pages/access-recovery.adoc for access recovery procedure.
```

jlebon:
@austinnichols101 Ahhh, you're doing this over tty0, right? Try adding `console=tty0` at the end.

modules/ROOT/pages/access-recovery.adoc
jlebon: Good point, will tweak.

austinnichols101:
@jlebon - Adding `console=tty0` helped.  Then at the prompt I 

- Loaded the SELinux policy: `/sbin/load_policy -i`
- Reset the password for `root` and `core`: `passwd <user>`
. Reboot: `/sbin/reboot -f`

However, after I reboot I'm unable to log on as either `core` or `root`.

image

jlebon:
> Note that I did make a first attempt WITHOUT loading the SELinux policy so that might have screwed things up.

Yeah, judging from the AVC denials, that seems to be the issue. You can fix it using `/sbin/restorecon -v /etc/{passwd,shadow}` from the shell boot again.

modules/ROOT/pages/access-recovery.adoc
jlebon: Done below! :arrow_down:

austinnichols101:
That solved the problem.  Thanks!

@jamescassell - I think the `/sbin/restorecon -v /etc/{passwd,shadow}` should be included in the procedure with a note/warning to load the selinux policy as the first step after logging in.

jlebon:
I added a bit about `restorecon` to be explicit.

modules/ROOT/pages/access-recovery.adoc
```diff
@@ -0,0 +1,10 @@
+= Access Recovery
+
+If you've lost the private key of an SSH keypair used to log into Fedora CoreOS, and do not have any password logins set up to use at the console, you can gain access back to the machine using the `init=/bin/sh` trick:
+
+. When booting the system, intercept the GRUB menu and edit the entry to append `init=/bin/sh` to the kernel argument list, then press Ctrl-X to resume booting.
+. Once the system has booted into a shell prompt, load the SELinux policy with `/sbin/load_policy -i`. If you miss this step before running `passwd`, you'll want to run `/sbin/restorecon -v /etc/{passwd,shadow}` before rebooting.
+. Set or reset the password for the target user using the `passwd` utility.
+. And finally, reboot the system with `/sbin/reboot -f`.
+
+You should now be able to log back into the system at the console. From there, you can e.g. fetch a new public SSH key to add to `~/.ssh/authorized_keys` and delete the old one.
```
bgilbert: I think it's confusing that we're telling users to set a password and then leaving it set afterward.  It'd be good to say something about this: either locking out the password afterward, or explicitly pointing out that the password will not be accepted over SSH by default.

jlebon: I've added a bit about this.

modules/ROOT/pages/access-recovery.adoc
```diff
@@ -0,0 +1,10 @@
+= Access Recovery
+
+If you've lost the private key of an SSH keypair used to log into Fedora CoreOS, and do not have any password logins set up to use at the console, you can gain access back to the machine using the `init=/bin/sh` trick:
+
+. When booting the system, intercept the GRUB menu and edit the entry to append `init=/bin/sh` to the kernel argument list, then press Ctrl-X to resume booting.
```
jamescassell: By holding the left shift key or sequentially pressing the up and down arrows to interrupt the grub timeout, then press `e` to edit the selected entry, entering the grub user (usually root) and password if necessary. (Do we support grub passwords?)

jlebon: Maybe let's keep further tweaks as follow-ups and just get this one in? It does more good as is on the docs site, than not-quite-perfect sitting here. :)

PR #51, opened 2020-02-20
ign-storage.adoc: add example for creating a /var partition

jlebon:
We need to have this stuff better documented because it's pretty easy to
mess up right now. (Of course, the next step is implementing fcct sugar
to make this nicer).

### DIFF
--- a/modules/ROOT/pages/ign-storage.adoc
+++ b/modules/ROOT/pages/ign-storage.adoc
@@ -114,6 +114,41 @@ storage:
 <4> We use the previously defined partitions as devices in a RAID1 md array.
 <5> The resulting md array is used to create an EXT4 file system.
 
+.Example for adding a /var partition to the primary disk
+[source,yaml]
+----
+variant: fcos
+version: 1.0.0
+storage:
+  disks:
+  - device: /dev/vda <1>
+    wipe_table: false <2>
+    partitions:
+    - size_mib: 0
+      start_mib: 0
+      label: var <3>
+  filesystems:
+    - path: /var
+      device: /dev/disk/by-partlabel/var
+      format: xfs
+systemd:
+  units:
+    - name: var.mount <4>
+      enabled: true
+      contents: |
+        [Unit]
+        Before=local-fs.target
+        [Mount]
+        Where=/var
+        What=/dev/disk/by-partlabel/var
+        [Install]
+        WantedBy=local-fs.target
+----
+<1> The name of the primary block device. In virtio-based setups, this is likely `/dev/vda`. Elsewhere, it's likely `/dev/sda`.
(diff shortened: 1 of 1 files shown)


jlebon:
Related discussions in https://github.com/coreos/coreos-assembler/pull/924.

modules/ROOT/pages/ign-storage.adoc
```diff
@@ -114,6 +114,43 @@ storage:
...
+      contents: |
+        [Unit]
+        Before=local-fs.target
+        [Mount]
+        Where=/var
+        What=/dev/disk/by-partlabel/var
+        [Install]
+        WantedBy=local-fs.target
+----
+<1> The name of the primary block device. In virtio-based setups, this is likely `/dev/vda`. Elsewhere, it's likely `/dev/sda`.
```
bgilbert: Rather than footnotes, let's just use comments within the YAML doc.  (Should probably fix the other examples to do the same.)

modules/ROOT/pages/ign-storage.adoc
```diff
@@ -114,6 +114,43 @@ storage:
...
+variant: fcos
+version: 1.0.0
+storage:
+  disks:
+  - device: /dev/vda <1>
+    wipe_table: false <2>
+    partitions:
+    - number: 5 <3>
+      size_mib: 0
+      start_mib: 0
```
bgilbert: Should we mention how to leave space for expanding the root partition?

modules/ROOT/pages/ign-storage.adoc
```diff
@@ -114,6 +114,43 @@ storage:
...
+[source,yaml]
+----
+variant: fcos
+version: 1.0.0
+storage:
+  disks:
+  - device: /dev/vda <1>
+    wipe_table: false <2>
+    partitions:
+    - number: 5 <3>
```
bgilbert: Not `0`?

modules/ROOT/pages/ign-storage.adoc
jlebon: I think we should do this, though I think it's more important to be consistent and I don't want to tackle the other examples right now.

modules/ROOT/pages/ign-storage.adoc
jlebon: Ahh thanks, that's better!

modules/ROOT/pages/ign-storage.adoc
jlebon: Not sure. I think we should discuss the strategy further before encoding it here.

jlebon:
Updated! :arrow_up:

jlebon:
Can we get this in?

Issue #50, opened 2020-02-18
user-guides: add platform docs for vmware


lucab:
We currently don't have any documentation on how to prepare, customize and first-boot FCOS on the `vmware` platform.

We can start by porting over the docs from Container Linux at https://coreos.com/os/docs/latest/booting-on-vmware.html#vmware-guestinfo-interface.

More interestingly, now there seems to be an official [PowerShell based container](https://hub.docker.com/r/vmware/powerclicore/) which should even support [setting guestinfo](https://github.com/vmware/PowerCLI-Example-Scripts/pull/231).

lucab:
Self note: [Import-VApp](https://powercli-core.readthedocs.io/en/latest/cmd_import.html#import-vapp) does not seem to support remote HTTP source. An upload from a local source path seems required.

lucab:
Self note: I got a simple flow working with OVA importing (after https://github.com/coreos/coreos-assembler/pull/1141), steps here below.

```
# OVA is locally at ./templates/fcos.ova
podman run --rm -it -v ./templates:/root/templates:z vmware/powerclicore:latest

# Interactive authentication
Connect-VIServer -Server vsphere.example.com

# Host machine
$vmwHost = 'host1.example.com'
# Path to OVA (in container)
$ovaPath = '/root/templates/fcos.ova'
# Ignition config (after `base64 -w0`)
$nodeConfig = @{}
$nodeConfig["guestinfo.ignition.config.data.encoding"] = "base64"
$nodeConfig["guestinfo.ignition.config.data"] = "<Ignition config in base64>"

Import-VApp -VMHost $vmwHost -Name fcos -OvfConfiguration $nodeConfig -Source $ovaPath
```

lucab:
Before committing to this PowerCLI approach, I'd like to explore https://github.com/vmware/govmomi/blob/master/govc/README.md too. On paper, it may integrate better with the rest of our Golang world.

lucab:
Equivalent steps using `govc` and storing the OVA as a library template:
```
mkdir templates
coreos-installer download -s stable -p vmware -f ova -C ./templates/

govc session.login -u 'user:pass@vsphere.example.com'
govc library.create fedora-coreos
govc library.import -n fcos-stable fedora-coreos ./templates/fedora-coreos-<VER>-vmware.x86_64.ova

govc library.deploy fedora-coreos/fcos-stable fcos-stable-worker01
[...]

PR #53, opened 2020-02-24
platforms: allocate ID for Exoscale

pierre-emmanuelJ:
This adds `exoscale` as the platform ID for Exoscale cloud provider

Homepage: https://www.exoscale.com/
User data and meta data: https://community.exoscale.com/documentation/compute/cloud-init/#querying-the-user-data-and-meta-data-from-the-instance

https://github.com/coreos/fedora-coreos-tracker/issues/384

### DIFF
--- a/modules/ROOT/pages/platforms.adoc
+++ b/modules/ROOT/pages/platforms.adoc
@@ -25,6 +25,7 @@ Here is a list of all supported platforms and their identifier:
  * `aliyun`: Aliyun/Alibaba Cloud (cloud platform)
  * `aws`: Amazon Web Services (cloud platform)
  * `azure`: Microsoft Azure (cloud platform)
+ * `exoscale`: Exoscale (cloud platform)
  * `gcp`: Google Cloud Platform (cloud platform)
  * `ibmcloud`: IBM Cloud, VPC Generation 2 (cloud platform)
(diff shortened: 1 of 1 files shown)


jlebon:
Re. this one and https://github.com/coreos/fedora-coreos-docs/pull/37, shouldn't we wait until we're actually producing the images and they're available on the download page before publishing on the docs page?

pierre-emmanuelJ:
> Re. this one and #37, shouldn't we wait until we're actually producing the images and they're available on the download page before publishing on the docs page?

Yes good idea! We can put the pull request in hold if you want

pierre-emmanuelJ:
you mean this one (https://github.com/coreos/coreos-assembler/pull/1127) instead of #37 ?

lucab:
@jlebon I'm considering the platform page more or less like developer documentation to keep track of all the platforms.
For user documentation, I have things like https://github.com/coreos/fedora-coreos-docs/issues/50 in mind. The latter should wait until we have images for users, indeed.

jlebon:
> @jlebon I'm considering the platform page more or less like developer documentation to keep track of all the platforms.
For user documentation, I have things like #50 in mind. The latter should wait until we have images for users, indeed.

OK, gotcha. WFM!

PR #61, opened 2020-03-19
ign-storage: convert FCC footnotes to inline comments

bgilbert:
cc @jlebon


modules/ROOT/pages/ign-storage.adoc
```diff
@@ -96,23 +110,20 @@ storage:
       number: 1
       size_mib: 65536
       start_mib: 0
-  raid:  <4>
+  # We use the previously defined partitions as devices in a RAID1 md array.
+  raid:
```
jlebon: Minor: indentation here and in the next one looks off.

PR #62, opened 2020-04-10
getting-started: add Azure and DigitalOcean instructions

bgilbert:


lucab:
@bgilbert there is a "Provisioning Machines" section which already contains VMware instructions. Can we move these two new sections there as "Booting on Azure" and "Booting on DigitalOcean" pages?

bgilbert:
@lucab I can move these to separate pages, but then we should probably have a separate PR to move the existing getting-started sections out to their own files as well.  I don't think we should treat some cloud platforms differently than others.  WDYT?

bgilbert:
Updated.

modules/ROOT/pages/provisioning-digitalocean.adoc
```diff
@@ -0,0 +1,49 @@
...
+Fedora CoreOS is designed to be updated automatically, with different schedules per stream.
+
+. Once you have picked the relevant stream, find the corresponding DigitalOcean image on the https://getfedora.org/coreos/download?tab=cloud_operators[download page] and copy the URL of the Download link.
+
+. Create the custom image:
++
+.Example uploading FCOS to a DigitalOcean custom image
+[source, bash]
+----
+doctl compute image create my-fcos-image --region sfo2 --image-url <download-url>
```
lucab, line 22: Unrelated: this looks like another usecase for https://github.com/coreos/coreos-installer/issues/202#issuecomment-606812958.

lucab:
@bgilbert I agree with you. I think that longer term we want all of:
 * a getting-started page with simple instructions for a popular cloud (AWS) and for a popular bare hypervisor (qemu)
 * a dedicated page for AWS, which can contain further details
 * a dedicated page for qemu (libvirt too?), which can contain further details

modules/ROOT/pages/provisioning-azure.adoc
```diff
@@ -0,0 +1,62 @@
...
+az storage container create --connection-string "$cs" -n my-container
+----
+
+. Create an FCOS image:
++
+.Example creating Azure image
+[source, bash]
+----
+# Upload image blob
+az storage blob upload --connection-string "$cs" -c my-container -f <downloaded-image-file.vhd> -n my-fcos-image.vhd
```
dustymabe, line 47: Since we used variables in other places maybe we should make one here:

```
file=<downloaded-image-file.vhd>
az storage blob upload --connection-string "$cs" -c my-container -f "$file" -n my-fcos-image.vhd
```

modules/ROOT/pages/provisioning-azure.adoc
```diff
@@ -0,0 +1,62 @@
...
+# Create the image
+az image create -n my-fcos-image -g my-group --source https://mystorageacct.blob.core.windows.net/my-container/my-fcos-image.vhd --location westus2 --os-type Linux
+# Delete the uploaded blob
+az storage blob delete --connection-string "$cs" -c my-container -n my-fcos-image.vhd
+----
+
+== Launching a VM instance
+
+. Launch a VM. Your Ignition configuration can be passed to the VM as custom data, or you can skip passing custom data if you just want SSH access. Your SSH public key from `~/.ssh` will automatically be added to the VM. This provides an easy way to test out FCOS without first creating an Ignition config.
```
dustymabe, line 56: Interesting. The az command line transparently picks up your ssh pubkey and passes it to the instance without you explicitly requesting it to do so? I assume that goes in through the metadata service (i.e.  not userdata/ignition).

modules/ROOT/pages/provisioning-digitalocean.adoc
miabbott, line 22: I've been using DO for a couple years now and somehow this is the first time I've seen `doctl`

Could we explicitly link to the GH repo as part of the requirements?

modules/ROOT/pages/provisioning-azure.adoc
bgilbert: Correct.

bgilbert:
@lucab I'm okay with special-casing qemu on the getting-started page, but I don't think we should emphasize one particular cloud offering.  Maybe the getting-started page can link to the docs pages for the various supported clouds?

bgilbert:
Updated.

bgilbert:
https://github.com/coreos/fedora-coreos-docs/pull/63 splits AWS and qemu out to their own pages.

PR #63, opened 2020-04-14
Split AWS/qemu instructions out of getting-started to their own pages

bgilbert:
Prompted by the discussion in #62.

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -5,6 +5,7 @@
 ** xref:faq.adoc[FAQ]
 ** xref:migrate-cl.adoc[Migrating from Container Linux]
 ** Provisioning Machines
+*** xref:provisioning-aws.adoc[Booting on AWS]
 *** xref:provisioning-azure.adoc[Booting on Azure]
 *** xref:provisioning-digitalocean.adoc[Booting on DigitalOcean]
 *** xref:provisioning-vmware.adoc[Booting on VMware]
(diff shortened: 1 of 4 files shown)


lucab:
I don't have strong opinions here. This is definitely more consistent, but now getting-started lacks any immediate actionable snippet. I was thinking of keeping the "aws with cloud keys" and "qemu with temporary storage" snippets and then linking to further pages, but I'm also ok with this version. Either directions, PR content looks good.

I'll leave some space for @jlebon who initially wrote this.

dustymabe:
> I don't have strong opinions here. This is definitely more consistent, but now getting-started lacks any immediate actionable snippet. I was thinking of keeping the "aws with cloud keys" and "qemu with temporary storage" snippets and then linking to further pages, but I'm also ok with this version. Either directions, PR content looks good.

I wonder if asciidoc has the ability to source content from other places. i.e. we would be able to include the content in both places but not duplicate the text, with is a maintainability problem.

lucab:
@dustymabe it can do that: https://github.com/coreos/fedora-coreos-docs/blame/master/modules/ROOT/pages/auto-updates.adoc#L63

jlebon:
Lukewarm on this myself. I agree with Luca it's a bit anticlimactic for a "Getting Started" page to only link to more links containing instructions. But maybe it's OK... Sharing content for the basic scenarios as described above sounds like an interesting compromise.

lucab:
Middleground proposal for this:
 * split the current getting-started steps out of there, into four dedicated files:
   1. getting-started-aws
   2. getting-started-qemu
   3. provisioning-aws
   4. provisioning-qemu
 * setup includes so that the getting-started-* sections are sourced by both the getting-started page and by their relevant provisioning-* one

I think @bgilbert prefers keeping only the getting-started-qemu steps as an introduction, I'm fine either ways (anecdotally I think this may drag people into qemu networking woes, though).

lucab:
Closing, https://github.com/coreos/fedora-coreos-docs/pull/97 took care of this.

PR #64, opened 2020-04-23
getting started: add tip for running QEMU with AppArmor

nrvnrvn:


lucab:
@nrvnrvn thanks for the PR! Would you mind holding this for a bit? We are re-shuffling the getting-started and qemu bits at https://github.com/coreos/fedora-coreos-docs/pull/63.

nrvnrvn:
@lucab sure no problem.

lucab:
Hey @nrvnrvn, thanks for bearing with us. There is now a dedicated libvirt page, I think you can add your notes there: https://docs.fedoraproject.org/en-US/fedora-coreos/provisioning-libvirt/

travier:
This also probably needs to use the "local" override in `/etc/apparmor.d/local/`. See at the end of: https://ubuntu.com/server/docs/security-apparmor

PR #66, opened 2020-04-26
Fix broken rendering of Update Streams

pbokoc:
This PR fixes the broken Update Streams page which was reported in https://pagure.io/fedora-infrastructure/issue/8862. It also makes one other smaller fix (start each item in a numbered list with a `.` instead of `1.` - the latter still works but you get a warning on build), and it updates `build.sh` and `preview.sh` to the latest version found in the Template repo.


dustymabe:
Thanks @pbokoc looks good to me except a few small changes.

dustymabe:
Here are the commits I added on top:

- 9fc3b6b  -  build.sh: remove mixed spaces and tabs
- 2b9a681 - update-streams: rm `\` from blockquote

PR #67, opened 2020-04-29
platforms: add DigitalOcean identifier

lucab:
This records the `digitalocean` identifier, which is currently used
by DigitalOcean custom images but was missing from the list.


bgilbert:
LGTM

PR #69, opened 2020-05-05
provisioning/vmware: minor fixes

lucab:
This contains minor wording fixes for the VMware user-guide.


dustymabe:
I don't know much about VMWare but I trust you two.

PR #71, opened 2020-05-08
provisioning: Add getting started on Vultr tutorial

dghubble:
* Add Provisioning Fedora CoreOS on Vultr docs that shows uploading a snapshot image and creating an instance using the `vultr-cli`
* Ignition config must include SSH authorized key setup

Related: https://github.com/coreos/fedora-coreos-tracker/issues/355

### DIFF
--- a/modules/ROOT/pages/provisioning-vultr.adoc
+++ b/modules/ROOT/pages/provisioning-vultr.adoc
@@ -1,0 +1,37 @@
+= Provisioning Fedora CoreOS on Vultr
+
+This guide shows how to provision new Fedora CoreOS (FCOS) nodes on Vultr.  Fedora CoreOS images are currently not published directly on Vultr, so you must download a Fedora CoreOS Vultr image and upload it to your Vultr account as a https://www.vultr.com/docs/requirements-for-uploading-an-os-iso-to-vultr[custom image].
+
+== Prerequisites
+
+Before provisioning a FCOS machine, you must have an Ignition configuration file that sets SSH authorized keys for the core user. If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File]. While the Vultr documentation mentions cloud-init and scripts, FCOS does not support cloud-init or the ability to run scripts from user-data. It accepts only Ignition configuration files.
+
+
+You also need to have access to a Vultr account.  The examples below use the https://github.com/vultr/vultr-cli[vultr-cli] command-line tool.
+
+== Creating a Vultr custom snapshot
+
+Fedora CoreOS is designed to be updated automatically, with different schedules per stream.
+
+. Once you have picked the relevant stream, find the corresponding Vultr image on the https://getfedora.org/coreos/download?tab=cloud_operators[download page] and copy the URL of the Download link.
+
+. Create the custom image:
++
+.Example uploading FCOS to a Vultr custom snapshot
+[source, bash]
+----
+export VULTR_API_KEY=<token>
+vultr-cli snapshot create-url -u <download-url>
+----
+
+=== Launching an instance
+
+Create a FCOS Vultr instance using the snapshot id. This example creates a 1 vCPU, 1GB RAM instance in NYC. Use `vultr-cli regions list` and `vultr plans list` for other options.
+
+[source, bash]
+----
+vultr-cli server create --snapshot <snapshot-id> --region 1 --plan 201 --userdata "$(cat example.ign)"
+----
+
+NOTE: Vultr firewall rules must be adjusted if you wish to ssh to the instance.
+
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -8,6 +8,7 @@
 *** xref:provisioning-azure.adoc[Booting on Azure]
 *** xref:provisioning-digitalocean.adoc[Booting on DigitalOcean]
 *** xref:provisioning-vmware.adoc[Booting on VMware]
+*** xref:provisioning-vultr.adoc[Booting on Vultr]
 ** System Configuration
 *** xref:producing-ign.adoc[Producing an Ignition File]
 *** xref:fcct-config.adoc[FCCT Specification]


dghubble:
I don't typically create instances via CLI, but I did test it out for this using my Vultr image (hosted elsewhere, pending FCOS pipeline publishing) and a small Ignition config with just an SSH authorized key setup. It provisioned and I'm able to login (with cloud firewall rule added).

modules/ROOT/pages/provisioning-vultr.adoc
```diff
@@ -0,0 +1,37 @@
...
+vultr-cli snapshot create-url -u <download-url>
+----
+
+=== Launching an instance
+
+Create a FCOS Vultr instance using the snapshot id. This example creates a 1 vCPU, 1GB RAM instance in NYC. Use `vultr-cli regions list` and `vultr plans list` for other options.
+
+[source, bash]
+----
+vultr-cli server create --snapshot <snapshot-id> --region 1 --plan 201 --userdata "$(cat example.ign)"
```
dustymabe, line 33: does vultr support providing a file to `--userdata`? Or another option that allows the user to specify a file?

modules/ROOT/pages/provisioning-vultr.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+. Once you have picked the relevant stream, find the corresponding Vultr image on the https://getfedora.org/coreos/download?tab=cloud_operators[download page] and copy the URL of the Download link.
+
+. Create the custom image:
++
+.Example uploading FCOS to a Vultr custom snapshot
+[source, bash]
+----
+export VULTR_API_KEY=<token>
+vultr-cli snapshot create-url -u <download-url>
```
dustymabe, line 24: is there any option here to do verification of the image (maybe via providing a checksum)?

modules/ROOT/pages/provisioning-vultr.adoc
```diff
@@ -0,0 +1,37 @@
...
+=== Launching an instance
+
+Create a FCOS Vultr instance using the snapshot id. This example creates a 1 vCPU, 1GB RAM instance in NYC. Use `vultr-cli regions list` and `vultr plans list` for other options.
+
+[source, bash]
+----
+vultr-cli server create --snapshot <snapshot-id> --region 1 --plan 201 --userdata "$(cat example.ign)"
+----
+
+Note, Vultr firewall rules must be adjusted if you wish to ssh to the instance.
```
dustymabe, line 36: can you change this to

```
NOTE: Vultr firewall rules must be adjusted if you wish to ssh to the instance.
```

It will make a pretty text box with the information if you do that.

jlebon:
We should probably hold this until it's actually downloadable from the getfedora.org page. (I added a corresponding checkbox in https://github.com/coreos/fedora-coreos-tracker/issues/355).

modules/ROOT/pages/provisioning-vultr.adoc
dghubble: I looked for one for the CLI to read the file itself, but looks like it only has an option for the string input.

```
  ...
  -t, --tag string               The tag to assign to this instance
  -u, --userdata string          userdata you want to give this instance
```

modules/ROOT/pages/provisioning-vultr.adoc
dghubble: That would be sweet, but the CLI just takes any URL to download, same as their UI. No checksum verification.

```
Usage:
  vultr-cli snapshot create-url [flags]

Flags:
  -h, --help         help for create-url
  -u, --url string   Remote URL from where the snapshot will be downloaded.
```

modules/ROOT/pages/provisioning-vultr.adoc
dustymabe: Thanks!

dustymabe:
> We should probably hold this until it's actually downloadable from the getfedora.org page

agree.

Thanks @dghubble - we'll merge this when the next round of releases comes out.

modules/ROOT/pages/provisioning-vultr.adoc
summatix: How do you access `userdata` from within the snapshot without embedding the Vultr API key? AFAIK `userdata` was a feature that Vultr once used alongside `cloud-init` but has since essentially been deprecated and is now undocumented.

Just wondering if you know something I don't know.

summatix: I see from https://github.com/coreos/fedora-coreos-tracker/issues/355 you reference using the old metadata API from this archived page: https://web.archive.org/web/20190513194756/https://www.vultr.com/metadata/#user

I wasn't aware that `http://169.254.169.254/user-data/user-data` still worked since it has long been removed from Vultr's documentation. I just tested it and it is in fact working. I assumed this old metadata API was replaced by `v1`, which no longer includes userdata: https://www.vultr.com/metadata/

dghubble: Several things here. Like most cloud providers, Vultr runs a metadata service to provide guest VMs network access to "metadata" about themselves (IP address, disks, userdata). Access doesn't use a providers API tokens because (1) its metadata about the asking node (not secret) and (2) because it cannot, machines fetch userdata early in their lifecycle. Cloud providers serve metadata such as opaue userdata and OSes interpret that as cloud-init, Ignition, whatever.

I suspect Vultr doesn't mention the endpoint becuase for most users, their OS consumes from it already. Cloud-init relies on the same mechanism.

dghubble: Ignition implementation that fetches userdata on many platforms is https://github.com/coreos/ignition/tree/master/internal/providers.

summatix: I'm waiting to hear back from the Vultr development team about this. AFAIK the method being used (https://github.com/coreos/ignition/blob/master/internal/providers/vultr/vultr.go#L32) is deprecated and hence is not present in their documentation, but the dev team will know for sure the status of that endpoint.

summatix: I've confirmed with the Vultr dev team that this is simply an undocumented but not a deprecated feature:

> the /user-data/user-data endpoint is still supported and no it will not be going away.
>
> The reason you do not see this in v1 is it is a undocumented/hidden field. There are tools that rely on user-data/user-data which is why this still exists. 
> So when you send create an instance with user-data you will have to call  169.254.169.254/user-data/user-data this is where the data will get populated not anywhere in the v1.

dustymabe:
hmm was just about to re-look at merging this but I still don't see vultr on the download page.

dghubble:
Vultr images show up in meta.json, but it seems a few more pieces are missing https://github.com/coreos/fedora-coreos-tracker/issues/355#issuecomment-640963164

lucab:
The website is now showing Vultr artifacts: https://getfedora.org/en/coreos/download?tab=cloud_operators&stream=stable.

summatix:
Following the instructions here: https://docs.fedoraproject.org/en-US/fedora-coreos/provisioning-vultr/

I'm seeing the error: "Expected a raw disk image, but found: xz compressed data"

So the instructions may need to be updated a little (unless I'm misreading the instructions).

lucab:
Ah yes, there is a decompress step missing. We can probably switch to use `coreos-installer` too, which can internally do the verify and decompress steps too.

summatix:
I see in the Vultr panel there is a new "Fedora CoreOS" server type option that requests for ignition content. So it seems like the Vultr team have added native support for Fedora CoreOS already.

PR #73, opened 2020-05-11
provisioning/vmware: minor examples cleanup

lucab:
This drops a few unnecessary `export` statements from examples.


dustymabe:
Thanks @lucab

PR #74, opened 2020-05-17
pages: emergency-shell

web-refinery:
Adds documentation to describe how to access the emergency shell with tty0.

Resolves (by documentating) https://github.com/coreos/fedora-coreos-tracker/issues/453

### DIFF
--- a/modules/ROOT/pages/emergency-shell.adoc
+++ b/modules/ROOT/pages/emergency-shell.adoc
@@ -1,6 +1,6 @@
 = Access the emergency shell on boot.
 
-If there is an error while booting Fedora CoreOS and emergency.target is reached a shell prompt will open by default on ttyS0 (i.e. serial console). This usually only happens during first-boot provisioning..
+If there is an error while booting Fedora CoreOS and emergency.target is reached a shell prompt will open by default on ttyS0 (i.e. serial console). This usually only happens during first-boot provisioning.
 
 To access the emergency shell when not using a serial console, you should intercept the GRUB menu and edit the entry to drop all 'console' parameters from the kernel line that you don't need, then press Ctrl+X to resume booting.
 
--- a/modules/ROOT/pages/emergency-shell.adoc
+++ b/modules/ROOT/pages/emergency-shell.adoc
@@ -1,0 +1,5 @@
+= Access the emergency shell on boot.
+
+If there is an error while booting Fedora CoreOS and emergency.target is reached a shell prompt will open by default on ttyS0 (i.e. serial console).
+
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/emergency-shell.adoc
```diff
@@ -0,0 +1,5 @@
+= Access the emergency shell on boot.
+
+If there is an error while booting Fedora CoreOS and emergency.target is reached a shell prompt will open by default on ttyS0 (i.e. serial console).
+
+To access the emergency shell when using VGA intercept the GRUB menu and edit the entry to append 'console=tty0' to the kernel argument list, then press Ctrl-X to resume booting.
```
lucab, line 5: I think that parameter is already on the command-line (both should be there).
If the machine does not have/use a serial console, it should be enough to drop the unwanted `console=` argument.
Can you confirm that works for you too?

lucab:
@web-refinery thanks for the PR! I've left a comment with a minor doubt I had.

In general, we should mention somewhere that this only affects the first-boot provisioning, i.e. when Ignition fails to setup a new machine.

web-refinery:
@lucab You were correct, thanks for the feedback! There are several console parameters by default. I had only noticed the serial console one. How's the PR looking now?

It is unfortunate that we can only output to all console's but not mirror them so the emergency shell would be on all of them. Oh well.

modules/ROOT/pages/emergency-shell.adoc
```diff
@@ -0,0 +1,7 @@
+= Access the emergency shell on boot.
+
+If there is an error while booting Fedora CoreOS and emergency.target is reached a shell prompt will open by default on ttyS0 (i.e. serial console). This usually only happens during first-boot provisioning.
+
+To access the emergency shell when not using a serial console, you should intercept the GRUB menu and edit the entry to drop all 'console' parameters from the kernel line that you don't need, then press Ctrl+X to resume booting.
+
+For VGA, you would keep 'console=tty0'
```
lucab, line 7: This is either missing the final dot, or a whole part of a sentence.

I'd probably move this as the second paragraph in the page, explaining that there are two `console=` parameters by default and what each is for (perhaps as a bullet list).

web-refinery:
@lucab Thanks again for the feedback. I've made some more modifications, to (hopefully) improve readability. 

If I understood your feedback correctly, this should cover it. Can you let me know or edit my fork directly?

lucab:
@web-refinery yes, this looks good. I'll squash and merge. Thanks again!

web-refinery:
Woo! My first ever pull request (ever), merged! Thanks!

(I was away so I hadn't noticed it yet!)

PR #80, opened 2020-06-04
Document SSH key and password authentication

bgilbert:
Hold until Fedora 32 reaches FCOS stable.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/138.

### DIFF
--- a/modules/ROOT/pages/authentication.adoc
+++ b/modules/ROOT/pages/authentication.adoc
@@ -1,0 +1,63 @@
+= Configuring Authentication
+
+== Using an SSH key
+
+To configure an SSH key for a local user, you can use a Fedora CoreOS Config:
+
+[source,yaml]
+----
+variant: fcos
+version: 1.1.0
+passwd:
+  users:
+    - name: core
+      ssh_authorized_keys:
+        - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDC5QFS...
+----
+
+== Using password authentication
+
+Fedora CoreOS ships with no default passwords. You can use a Fedora CoreOS Config to set a password for a local user:
+
+[source,yaml]
+----
+variant: fcos
+version: 1.1.0
+passwd:
+  users:
+    - name: core
+      password_hash: "$y$j9T$A0Y3wwVOKP69S.1K/zYGN.$S596l11UGH3XjN..."
+----
+
+To generate a secure password hash, use the `mkpasswd` command:
+
+[source]
+----
+$ mkpasswd --method=yescrypt
+Password: 
+$y$j9T$A0Y3wwVOKP69S.1K/zYGN.$S596l11UGH3XjN...
+----
+
+The `yescrypt` hashing method is recommended for new passwords. For more details on hashing methods, see `man 5 crypt`.
+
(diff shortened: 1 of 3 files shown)


dustymabe:
This looks great!

Any chance you could throw in an example of how to create a hashed password and an FCC that uses it? I know when I come to look at this documentation those two pieces of info (along with the SSH password auth enablement bit) are going to be exactly what I'm looking for to try to reproduce a problem or what not and having them together is going to make my life way easier.

modules/ROOT/pages/authentication.adoc
```diff
@@ -0,0 +1,21 @@
...
+version: 1.1.0
+storage:
+  files:
+    - path: /etc/ssh/sshd_config.d/02-enable-passwords.conf
+      mode: 0644
+      contents:
+        inline: |
+          # Fedora CoreOS disables SSH password login by default.
+          # Enable it.
+          # This file must sort before 04-disable-passwords.conf.
```
dustymabe, line 19: I wonder if since SSHD uses a "first one wins" strategy for config merging we should make the files we write easier to override by putting them at a higher number. i.e., `80-disable-passwords.conf`

bgilbert: We can't.  The Fedora package ships an `05-redhat.conf` that includes `PasswordAuthentication yes`.

bgilbert:
Updated!

modules/ROOT/pages/authentication.adoc
dustymabe: ahh - grr. I guess my same question would apply to the rpm maintainers. Started a discussion there: https://src.fedoraproject.org/rpms/openssh/pull-request/9

dustymabe:
This is ready for merge, but I'm converting to draft so no one merges yet as recommended in https://github.com/coreos/fedora-coreos-docs/pull/80#issue-427527491

dustymabe:
Should we link to this new documentation from the "By default, FCOS does not allow password logins via SSH. We recommend configuring SSH keys instead."  in https://docs.fedoraproject.org/en-US/fedora-coreos/migrate-cl/#_configuration_changes ?

bgilbert:
Done!

bgilbert:
Hold for [sshd packaging changes](https://bodhi.fedoraproject.org/updates/FEDORA-2020-a85cb3c327).

bgilbert:
Added a separate commit to renumber the `sshd_config.d` fragment more sensibly after https://github.com/coreos/fedora-coreos-config/pull/480 lands, but that PR is blocked on https://github.com/coreos/fedora-coreos-tracker/issues/540.  Do we want to land the rest of this now, or wait until we can document a long-term stable filename?

dustymabe:
> Do we want to land the rest of this now, or wait until we can document a long-term stable filename?

IIUC if someone implements `02-enable-passwords.conf` file now, things will still work even after the priority number of the other files are changed (in sshd and fcos configs). I think we can land the rest of it now.

bgilbert:
OK, moved that part to https://github.com/coreos/fedora-coreos-docs/pull/85.

Issue #6, opened 2019-08-14
Clarify docs on SSHing into FCOS running on QEMU


rfairley:
When following the sections [Booting with QEMU](https://docs.fedoraproject.org/en-US/fedora-coreos/getting-started/#_booting_with_qemu) and [Generating Ignition configs](https://docs.fedoraproject.org/en-US/fedora-coreos/getting-started/#_generating_ignition_configs), using the example config provided adding my SSH key to the `core` user, and SSHing in with `ssh core@<ip>` (where `<ip>` is the address assigned by the QEMU built-in DHCP server) does not appear to work.

After editing `/etc/ssh/sshd` config in the FCOS VM to comment out `PasswordAuthentication no`, set `PasswordAuthentication yes`, and `PubkeyAuthentication yes`, and then `systemctl restart sshd`, SSHing into the VM gives the following (full log under **Details** below):

```
$ ssh -p 22 -vvv core@10.0.2.15
...
debug3: authmethod_is_enabled publickey
debug1: Next authentication method: publickey
debug1: Offering public key: /home/rfairley/.ssh/id_rsa RSA SHA256:dwilTJGh02bpSp0U7BLPrmC4wn2wJ40lxOpmkf0fsbY agent
debug3: send packet: type 50
debug2: we sent a publickey packet, wait for reply
debug3: receive packet: type 51
debug1: Authentications that can continue: publickey,gssapi-keyex,gssapi-with-mic,password
...
debug2: we did not send a packet, disable method
debug3: authmethod_lookup password
debug3: remaining preferred: ,password
debug3: authmethod_is_enabled password
debug1: Next authentication method: password
core@10.0.2.15's password: 
debug3: send packet: type 50
debug2: we sent a password packet, wait for reply
debug3: receive packet: type 51
debug1: Authentications that can continue: publickey,gssapi-keyex,gssapi-with-mic,password
Permission denied, please try again.
```

(Entering the same password I set for `core` through an Ignition config, and used to log in through the serial console).

From looking at what `cosa run` [does for SSH](https://github.com/coreos/coreos-assembler/blob/32b95541b4fddbd03242415cca871d7bded75e36/src/cmd-run#L256), pubkey authentication worked using `hostfwd`:

```
[...]

PR #86, opened 2020-06-20
getting-started: add host port forwarding to QEMU example

rfairley:
See details in commit message.

---

Closes: https://github.com/coreos/fedora-coreos-docs/issues/6

### DIFF
--- a/modules/ROOT/pages/getting-started.adoc
+++ b/modules/ROOT/pages/getting-started.adoc
@@ -32,14 +32,15 @@ unxz fedora-coreos-qemu.qcow2.xz
 +
 TIP: Make sure that your user has access to `/dev/kvm` (this should already be the case on modern distros), or if `/dev/kvm` is owned by the `kvm` group, adding yourself to that group with `usermod -aG kvm $USER`.
 
-. Launch the VM using `qemu-kvm` or `virt-install`. The Ignition file is passed to the VM via the `-fw_cfg` parameter, which sets the `opt/com.coreos/config` key in the QEMU firmware configuration device. For libvirt, you must pass this through the `--qemu-commandline` argument. You can use `-snapshot` to make `qemu-kvm` allocate temporary storage for the VM, or `qemu-img create` to first create a layered qcow2.
+. Launch the VM using `qemu-kvm` or `virt-install`. The Ignition file is passed to the VM via the `-fw_cfg` parameter, which sets the `opt/com.coreos/config` key in the QEMU firmware configuration device. For libvirt, you must pass this through the `--qemu-commandline` argument. You can use `-snapshot` to make `qemu-kvm` allocate temporary storage for the VM, or `qemu-img create` to first create a layered qcow2. For QEMU usermode networking, the assigned IP address is not reachable from the host, so `hostfwd` is used to forward a port on `localhost` to the default SSH port on the guest machine.
 +
 .Example launching FCOS with QEMU (temporary storage)
 [source, bash]
 ----
 qemu-kvm -m 2048 -cpu host -nographic -snapshot \
 	-drive if=virtio,file=fedora-coreos-qemu.qcow2 \
-	-fw_cfg name=opt/com.coreos/config,file=path/to/example.ign
+	-fw_cfg name=opt/com.coreos/config,file=path/to/example.ign \
+	-nic user,model=virtio,hostfwd=tcp::2222-:22
 ----
 +
 .Example launching FCOS with QEMU (persistent storage)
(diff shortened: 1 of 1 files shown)


rfairley:
ref #63 for when rebasing, in case this gets merged beforehand

modules/ROOT/pages/getting-started.adoc
```diff
@@ -48,28 +49,32 @@ qemu-kvm -m 2048 -cpu host -nographic -snapshot \
...
 ----
 +
 .Example launching FCOS with virt-install
 [source, bash]
 ----
 virt-install --connect qemu:///system -n fcos -r 2048 --os-variant=fedora31 --import \
 	--graphics=none --disk size=10,backing_store=/path/to/fedora-coreos-qemu.qcow2 \
-	--qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=/path/to/example.ign"
+	--qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=/path/to/example.ign" \
+	--qemu-commandline="-device virtio-net,netdev=user.0 -netdev user,id=user.0,hostfwd=tcp::2222-:22"
```
jlebon, line 62: By default, `virt-install` should already be attaching the default network. Is that not the case on your box? But more generally, I think we should let libvirt be aware of the network config here instead of passing it "behind its back" via `--qemu-commandline`.

modules/ROOT/pages/getting-started.adoc
```diff
@@ -48,28 +49,32 @@ qemu-kvm -m 2048 -cpu host -nographic -snapshot \
...
 ----
 
 NOTE: When using `virt-install`, you must specify absolute paths to the image and the Ignition file.
 
 TIP: If running with SELinux enabled, you may need to change the label of the Ignition file to allow access: `chcon -t svirt_home_t path/to/example.ign`
 
-Once the VM has finished booting, its IP addresses will appear on the serial console. You can then SSH into the FCOS VM as the `core` user:
+Once the VM has finished booting, its IP addresses will appear on the serial console. You can then SSH into the FCOS VM as the `core` user (replacing `localhost` and `2222` with the IP address and port of the VM, if different values from the suggested values mentioned earlier were chosen):
```
jlebon, line 69: Once we simplify the `virt-install` one, I think we can reword this to:

> (on QEMU, use `localhost` and pass `-p 2222` or whichever port you've forwarded)

or something like that?

modules/ROOT/pages/getting-started.adoc
```diff
@@ -48,28 +49,32 @@ qemu-kvm -m 2048 -cpu host -nographic -snapshot \
...
 
 [source, bash]
 ----
-ssh core@<ip address>
+ssh core@localhost -p 2222
```
jlebon, line 73: I'd keep this unchanged since it should work as is for libvirt, and for QEMU we'll have the notice from the previous paragraph.

modules/ROOT/pages/getting-started.adoc
```diff
@@ -48,28 +49,32 @@ qemu-kvm -m 2048 -cpu host -nographic -snapshot \
...
 ----
 
+NOTE: With QEMU/libvirt, the IP addresses are assigned by the QEMU usermode networking backend, and will not be directly visible from your host machine. However, via host port forwarding through `localhost` (shown above), the machine is accessible.
```
jlebon, line 76: Hmm, feels like we can drop this paragraph given that it's explained in the sentence just before the `qemu-kvm` command-line?

modules/ROOT/pages/getting-started.adoc
lucab: Better yet, move to a `${LOCALPORT}` variable which is `2222` on qemu and `22` on libvirt.

modules/ROOT/pages/getting-started.adoc
lucab: Though I think we should have a note here to mention that the address shown in the console in qemu is not reachable from the host, and SSH port is forwarded to an higher port on localhost instead.

modules/ROOT/pages/getting-started.adoc
rfairley: Ahh, I had misunderstood how `virt-install` set up the network previously, thinking it used a similar usermode networking setup like QEMU. Makes a lot more sense to keep the networking set up via libvirt, and just use the IP address shown on the FCOS serial console.

modules/ROOT/pages/getting-started.adoc
rfairley: SGTM, added!

modules/ROOT/pages/getting-started.adoc
rfairley: :+1: to keeping this unchanged, and have the note above in https://github.com/coreos/fedora-coreos-docs/pull/86#discussion_r443684364. `${LOCALPORT}` sounds good, but  if we did this I think it'd look neater to have an `${IPADDRESS}` variable too (choosing localhost or the one assigned by libvirt), IMHO, to avoid a variable substitution in one part of the line but deleting/overwriting `<ip address>` in another. Since the user would need to enter the IP address manually anyway in the unchanged version or with a variable, and it wouldn't be too much more typing for `-p 2222`, I've left this as it is. Though really this is just minor preference, and I'd still be happy using the variables mentioned with a nudge (either both the IPADDRESS and LOCALPORT or just LOCALPORT).

modules/ROOT/pages/getting-started.adoc
rfairley: :+1: I condensed this note a bit, and kept the explanation at the above explanation (mentioning briefly the IP address not being reachable).

jlebon:
Thanks for this!

rfairley:
Updated - good to merge is all is good!

rfairley:
Thanks for the reviews!

PR #90, opened 2020-06-26
Corrected example path to disk device symlink

dmarsh19:


lucab:
@dmarsh19 good catch, thanks for fixing it as well!

PR #92, opened 2020-06-29
fcct: point to the container image

lucab:
This tweaks the FCCT docpage, prominently pointing to the container image.

### DIFF
--- a/modules/ROOT/pages/producing-ign.adoc
+++ b/modules/ROOT/pages/producing-ign.adoc
@@ -1,20 +1,22 @@
-:experimental:
-= Fedora CoreOS - Producing an Ignition File
+= Producing an Ignition File
 
-This guide provides easy, step-by-step instructions to produce a basic Ignition file.
+This guide provides step-by-step instructions on how to produce a basic Ignition file.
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/producing-ign.adoc
```diff
@@ -1,20 +1,22 @@
...
 == A simple example
+
 Create a basic Ignition file that modifies the default FCOS user `core` to allow this user login with an SSH key.
 
 The overall steps are as follows:
 
 . Write the Fedora CoreOS Configuration (FCC) file in the YAML format.
-. Download and install the Fedora CoreOS Configuration Transpiler (`fcct`).
-. Use `fcct` to convert the FCC file into an Ignition (JSON) file.
+. Use the `quay.io/coreos/fcct:release` container image to convert the FCC file into an Ignition (JSON) file.
```
jlebon, line 18: Maybe in parentheses here, we could add " (on Fedora, you can also install the `fcct` package)" or something?

lucab: Hmm, after your comment I realized we already have a dedicated page for getting `fcct` https://docs.fedoraproject.org/en-US/fedora-coreos/using-fcct/.

I think there is some room to deduplicate the content and merge the two pages.

jlebon: Yeah, probably worth just linking to the docs page and adding the `fcct` package there as another way for users who already have a Fedora environment?

PR #93, opened 2020-06-29
migrate-cl: update link to Afterburn attribute docs

bgilbert:
For https://github.com/coreos/afterburn/pull/444.

cc @lucab

### DIFF
--- a/modules/ROOT/pages/migrate-cl.adoc
+++ b/modules/ROOT/pages/migrate-cl.adoc
@@ -1,7 +1,7 @@
 :experimental:
 = Migrating from CoreOS Container Linux (CL) to Fedora CoreOS (FCOS)
 
-Fedora CoreOS is the official successor of CoreOS Container Linux, which will https://coreos.com/os/eol/[reach its end of life] on May 26, 2020. This page attempts to document the differences between CL and FCOS to ease the transition to FCOS.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/migrate-cl.adoc
```diff
@@ -37,7 +37,7 @@ When writing FCC files, note the following changes:
 ** `EC2` is now `AWS`
 ** `GCE` is now `GCP`
 +
-For more info, see the https://github.com/coreos/afterburn/blob/master/README.md[Afterburn readme].
+For more info, see the https://github.com/coreos/afterburn/blob/master/docs/usage/attributes.md[Afterburn documentation].
```
lucab, line 40: What about linking this to the upcoming 4.4.1 release to ensure the target stays in place?

The line of reasoning is: it's easier to find a resource that moved somewhere else (or got updated) if you are at least getting an handle to when/where it used to exist.

bgilbert: Wouldn't it make more sense to prefer linking to current documentation, rather than optimizing for the rare case that we have to update a dead link?

lucab: Eh, I don't agree it's a rare case, but I agree that I should stop fighting windmills at this point :)

lucab:
https://github.com/coreos/afterburn/pull/444 merged, we can land this one too.

PR #96, opened 2020-06-30
bare-metal: drop coreos.inst.stream=stable karg for PXE installs

jlebon:
Otherwise, osmet won't kick in.


jlebon:
Still need to test this. Noticed this when looking at https://github.com/coreos/fedora-coreos-tracker/issues/558.

jlebon:
I didn't test this, but the reporter in https://github.com/coreos/fedora-coreos-tracker/issues/558 did, so marking as ready. :)

PR #101, opened 2020-07-20
provisioning/vultr: add decompression steps

lucab:
This tweaks Vultr provisioning guide to download, decompress and
upload a FCOS image to a storage bucket. This is in order to expose
the raw image on a public URL, due to platform limitations in the
importing logic.

Ref: https://github.com/coreos/coreos-assembler/pull/1595

### DIFF
--- a/modules/ROOT/pages/provisioning-vultr.adoc
+++ b/modules/ROOT/pages/provisioning-vultr.adoc
@@ -1,37 +1,64 @@
 = Provisioning Fedora CoreOS on Vultr
 
-This guide shows how to provision new Fedora CoreOS (FCOS) nodes on Vultr.  Fedora CoreOS images are currently not published directly on Vultr, so you must download a Fedora CoreOS Vultr image and upload it to your Vultr account as a https://www.vultr.com/docs/requirements-for-uploading-an-os-iso-to-vultr[custom image].
+This guide shows how to provision new Fedora CoreOS (FCOS) nodes on Vultr. FCOS images are currently not published directly on Vultr, but they can be uploaded as https://www.vultr.com/docs/requirements-for-uploading-an-os-iso-to-vultr[custom images].
 
 == Prerequisites
 
 Before provisioning a FCOS machine, you must have an Ignition configuration file that sets SSH authorized keys for the core user. If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File]. While the Vultr documentation mentions cloud-init and scripts, FCOS does not support cloud-init or the ability to run scripts from user-data. It accepts only Ignition configuration files.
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/provisioning-vultr.adoc
```diff
@@ -1,37 +1,64 @@
...
+
+. Fetch the latest image suitable for your target stream (or https://getfedora.org/coreos/download/[download and verify] it from the web).
++
+[source, bash]
+----
+STREAM='stable'
+coreos-installer download -s "${STREAM}" -p vultr -f raw.xz --decompress
+----
+
+. Upload the raw image to your bucket, and note its public URL.
```
dustymabe, line 31: I know you link to https://www.vultr.com/docs/vultr-object-storage earlier in this doc, but I think the user (including me just now) is going to skim over that. Might be worth adding the same or similar information in this step as a `NOTE` or something.

dustymabe:
@bh7cw - would you want to try out these steps and see if it works for you?

modules/ROOT/pages/provisioning-vultr.adoc
lucab: Yes, some redundancy doesn't hurt.
Is it fine to tweak this as "Use s3cmd to upload the raw ..." and link that to [Vultr s3cmd example](https://www.vultr.com/docs/vultr-object-storage#s3cmd__Example_CLI_tool)? Or were you missing some specific detail here?

bh7cw:
> @bh7cw - would you want to try out these steps and see if it works for you?

I will try.

modules/ROOT/pages/provisioning-vultr.adoc
dustymabe: I think the part that is slightly confusing to me is that we need a separate set of credentials just for the s3 access, which makes sense, though was hoping there was a way to just to the upload with the vultr CLI and re-use the same creds. The link to the docs helped hammer that point home to me.

lucab: I was looking for a `vultr-cli` subcommand too, but at this point I think it simply does not exist (and their docs point to s3cmd directly).

lucab: Amended.

dustymabe:
That final change LGTM. Will wait to hear from @bh7cw!

bh7cw:
The steps are very clear. My test goes well except took a long time to upload fcos image to S3.
And if we have `AWS CLI` installed, we have one more option to use `AWS CLI` besides `s3cmd`. The following commands are the same:
`aws s3 cp fedora-coreos-${FCOS_VERSION}-vultr.x86_64.raw s3://{BUCKET}/`
`s3cmd put -P "fedora-coreos-${FCOS_VERSION}-vultr.x86_64.raw" "s3://{BUCKET}/`
Overall, LGTM.

dustymabe:
Thanks @bh7cw for verifying! I'll merge now.

Issue #100, opened 2020-07-18
Give the user more information about UEFI/BIOS boot compatibility for our disk image


dustymabe:
Let's add some information to the docs addressing https://discussion.fedoraproject.org/t/are-fedora-coreos-disk-images-hybrid-bios-uefi-bootable/21911/2

- add a FAQ entry for this
- add some helpful text in https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/
- not sure if there is anything about this we can add to https://docs.fedoraproject.org/en-US/fedora-coreos/platforms/

nasirhm:
I would like to work on it.

dustymabe:
Thanks!

dustymabe:
fixed by #102

PR #102, opened 2020-07-20
faq: add an entry for BIOS/UEFI boot modes

nasirhm:
According to issue: https://github.com/coreos/fedora-coreos-docs/issues/100

- A faq entry added for Hybrid boot on Metal.
- On the /bare-metal page, there's already a note stating it.
- On the /platforms page, It's already written there with `metal` as well.

### DIFF
--- a/modules/ROOT/pages/faq.adoc
+++ b/modules/ROOT/pages/faq.adoc
@@ -236,6 +236,6 @@ systemd:
 
 == Are Fedora CoreOS x86_64 disk images hybrid BIOS+UEFI bootable?
 
-The x86_64 images we provide can be used for either BIOS (legacy) boot or UEFI boot. They contain a hybrid BIOS/UEFI partition setup that allows them to be used for either. The exception to that is the `metal4k` 4k native image, which is targeted at disks with 4k sectors and [does not have a BIOS boot partition](https://github.com/coreos/coreos-assembler/blob/12029fea7798fa5d3535eafcf8c3d02f9a6095e4/src/cmd-buildextend-metal#L200-L202) because 4k native disks are [only supported with UEFI](https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/hard-drives-and-partitions#advanced-format-drives).
+The x86_64 images we provide can be used for either BIOS (legacy) boot or UEFI boot. They contain a hybrid BIOS/UEFI partition setup that allows them to be used for either. The exception to that is the `metal4k` 4k native image, which is targeted at disks with 4k sectors and https://github.com/coreos/coreos-assembler/blob/12029fea7798fa5d3535eafcf8c3d02f9a6095e4/src/cmd-buildextend-metal#L200-L202[does not have a BIOS boot partition] because 4k native disks are https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/hard-drives-and-partitions#advanced-format-drives[only supported with UEFI].
 
 https://discussion.fedoraproject.org/t/are-fedora-coreos-disk-images-hybrid-bios-uefi-bootable/21911[Discuss]
--- a/modules/ROOT/pages/faq.adoc
+++ b/modules/ROOT/pages/faq.adoc
@@ -236,6 +236,6 @@ systemd:
 
 == Are Fedora CoreOS x86_64 disk images hybrid BIOS+UEFI bootable?
 
-The bare metal images we provide can be used for either BIOS (legacy) boot or UEFI boot,
+The x86_64 images we provide can be used for either BIOS (legacy) boot or UEFI boot. They contain a hybrid BIOS/UEFI partition setup that allows them to be used for either. The exception to that is the `metal4k` 4k native image, which is targeted at disks with 4k sectors and [does not have a BIOS boot partition](https://github.com/coreos/coreos-assembler/blob/12029fea7798fa5d3535eafcf8c3d02f9a6095e4/src/cmd-buildextend-metal#L200-L202) because 4k native disks are [only supported with UEFI](https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/hard-drives-and-partitions#advanced-format-drives).
 
 https://discussion.fedoraproject.org/t/are-fedora-coreos-disk-images-hybrid-bios-uefi-bootable/21911[Discuss]
--- a/modules/ROOT/pages/faq.adoc
+++ b/modules/ROOT/pages/faq.adoc
@@ -234,7 +234,7 @@ systemd:
       mask: true
 ----
 
-== Are Fedora CoreOS disk images hybrid BIOS+UEFI bootable?
+== Are Fedora CoreOS x86_64 disk images hybrid BIOS+UEFI bootable?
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/faq.adoc
```diff
@@ -233,3 +233,9 @@ systemd:
     - name: docker.service
       mask: true
 ----
+
+== Are Fedora CoreOS disk images hybrid BIOS+UEFI bootable?
+
+The bare metal images we provide can be used for either BIOS (legacy) boot or UEFI boot, You can boot it in either BIOS (legacy) or UEFI mode regardless of what mode the OS will use once installed.
```
lucab, line 239: I think you mixed up a bunch of statements here, the second part is usually found in the context of booting the live-ISO (in order to image a bare-metal disk).

lucab:
@nasirhm thanks for the patch! I've left a comment in review, I think you can make the entry simpler without the second part.

nasirhm:
Hey @lucab , here's the path: https://github.com/coreos/fedora-coreos-docs/pull/102/commits/dec125e0b647d7c9d61fe6ed1f3453f5f419d75c

modules/ROOT/pages/faq.adoc
```diff
@@ -233,3 +233,9 @@ systemd:
     - name: docker.service
       mask: true
 ----
+
+== Are Fedora CoreOS disk images hybrid BIOS+UEFI bootable?
+
+The bare metal images we provide can be used for either BIOS (legacy) boot or UEFI boot,
```
dustymabe, line 239: Hey @nasirhm. I did a little more digging on this topic and updated my forum reponse: https://discussion.fedoraproject.org/t/are-fedora-coreos-disk-images-hybrid-bios-uefi-bootable/21911/2?u=dustymabe

```suggestion
The x86_64 images we provide can be used for either BIOS (legacy) boot or UEFI boot. They contain a hybrid BIOS/UEFI partition setup that allows them to be used for either. The exception to that is the `metal4k` 4k native image, which is targeted at disks with 4k sectors and [does not have a BIOS boot partition](https://github.com/coreos/coreos-assembler/blob/12029fea7798fa5d3535eafcf8c3d02f9a6095e4/src/cmd-buildextend-metal#L200-L202) because 4k native disks are [only supported with UEFI](https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/hard-drives-and-partitions#advanced-format-drives).
```

modules/ROOT/pages/faq.adoc
```diff
@@ -233,3 +233,9 @@ systemd:
     - name: docker.service
       mask: true
 ----
+
+== Are Fedora CoreOS disk images hybrid BIOS+UEFI bootable?
```
dustymabe, line 237:
```suggestion
== Are Fedora CoreOS x86_64 disk images hybrid BIOS+UEFI bootable?
```

nasirhm:
Thank You @dustymabe for the review and updates.

dustymabe:
Sigh, I forgot this was adoc and not markdown.. New suggestion to fix the links..

modules/ROOT/pages/faq.adoc
```diff
@@ -233,3 +233,9 @@ systemd:
     - name: docker.service
       mask: true
 ----
+
+== Are Fedora CoreOS x86_64 disk images hybrid BIOS+UEFI bootable?
+
+The x86_64 images we provide can be used for either BIOS (legacy) boot or UEFI boot. They contain a hybrid BIOS/UEFI partition setup that allows them to be used for either. The exception to that is the `metal4k` 4k native image, which is targeted at disks with 4k sectors and [does not have a BIOS boot partition](https://github.com/coreos/coreos-assembler/blob/12029fea7798fa5d3535eafcf8c3d02f9a6095e4/src/cmd-buildextend-metal#L200-L202) because 4k native disks are [only supported with UEFI](https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/hard-drives-and-partitions#advanced-format-drives).
```
dustymabe, line 239:
```suggestion
The x86_64 images we provide can be used for either BIOS (legacy) boot or UEFI boot. They contain a hybrid BIOS/UEFI partition setup that allows them to be used for either. The exception to that is the `metal4k` 4k native image, which is targeted at disks with 4k sectors and https://github.com/coreos/coreos-assembler/blob/12029fea7798fa5d3535eafcf8c3d02f9a6095e4/src/cmd-buildextend-metal#L200-L202[does not have a BIOS boot partition] because 4k native disks are https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/hard-drives-and-partitions#advanced-format-drives[only supported with UEFI].
```

nasirhm:
Oops, I didn't see it as well. Thanks @dustymabe .

modules/ROOT/pages/faq.adoc
```diff
@@ -233,3 +233,9 @@ systemd:
     - name: docker.service
       mask: true
 ----
+
+== Are Fedora CoreOS x86_64 disk images hybrid BIOS+UEFI bootable?
+
+The x86_64 images we provide can be used for either BIOS (legacy) boot or UEFI boot. They contain a hybrid BIOS/UEFI partition setup that allows them to be used for either. The exception to that is the `metal4k` 4k native image, which is targeted at disks with 4k sectors and https://github.com/coreos/coreos-assembler/blob/12029fea7798fa5d3535eafcf8c3d02f9a6095e4/src/cmd-buildextend-metal#L200-L202[does not have a BIOS boot partition] because 4k native disks are https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/hard-drives-and-partitions#advanced-format-drives[only supported with UEFI].
```
jlebon, line 239: Not sure if it's worth linking to cosa code, but meh. Hmm, also wish we had a better link than the Microsoft one, but I can't find one offhand, so WFM! (I found https://lwn.net/Articles/377895/ though that might be too detailed.)

jlebon, line 239: Maybe for the first sentence we can add the words "bare metal" since that's how it's labeled in the download page, e.g.:

```
The x86_64 bare metal images we provide can be used for either BIOS (legacy) boot or UEFI boot.
```

?

dustymabe: I think we should keep it generic. I would have claimed otherwise, but the truth is that this isn't limited to our bare metal images. From looking at the code the only x86_64 image we create that isn't hybrid is the 4kn disk image. Am I reading it correctly?

dustymabe: Yeah the existing link is fine. I'm sure it'll break at some point and we'll fix it.

jlebon: Sure, that's true. Though OTOH it's not like it matters much for most other platforms, and although I love the consistency, it's possible we may change it in the future (see e.g. https://github.com/coreos/fedora-coreos-config/pull/407) and should reverse the right to break that.

No strong opinion either way though.

dustymabe: Thanks. Let's merge as-is then.

dustymabe:
Thanks @nasirhm

nasirhm:
Thank you for the opportunity @dustymabe =)

PR #110, opened 2020-07-26
Change eth0 to enp1s0 to reference Predictable Interface Names

ccamacho:
Now we use predictable interface names, by using eth0 this can lead to misconfigurations in the guests.


dustymabe:
As part of reviewing this PR I realized there was some obsolete information in this file so I deleted it and re-organized the links so hopefully they are more obvious to the user. See #111

dustymabe:
Thanks for the contribution @ccamacho

PR #112, opened 2020-07-27
bare-metal: add section on PXE rootfs migration

bgilbert:
Don't update the PXE instructions yet, since we're not yet producing `rootfs` images.

Also simplify PXE kargs.

For https://github.com/coreos/fedora-coreos-tracker/issues/390.


jlebon:
Shouldn't the last commit also change the example PXE configs to use either `,fedora-coreos-32.20200715.3.0-live-rootfs.x86_64.img` or `coreos.live.rootfs_url=`?

bgilbert:
@jlebon Not for another two weeks.

Issue #108, opened 2020-07-23
document time based updates strategy


dustymabe:
ref: https://discussion.fedoraproject.org/t/zincati-update-strategy-time-based/21929/2

zincati 0.0.11 seems to be in all streams now.

PR #118, opened 2020-07-30
auto-updates: show Zincati periodic strategy

lucab:
This refreshes the auto-updates docs, adding a reference to the new
"periodic" strategy and showing it directly in the example.

Closes: https://github.com/coreos/fedora-coreos-docs/issues/108

### DIFF
--- a/modules/ROOT/pages/auto-updates.adoc
+++ b/modules/ROOT/pages/auto-updates.adoc
@@ -10,7 +10,7 @@ By default, the OS performs continual auto-updates via two components:
 == Wariness to updates
 
 The local Zincati agent periodically checks with a remote service to see when updates are available.
-A custom "rollout wariness" value (see https://github.com/coreos/zincati/blob/0.0.6/docs/usage/auto-updates.md#phased-rollouts-client-wariness-canaries[documentation]) can be provided to let the server know how eager, or how risk-averse, the node is to receiving updates.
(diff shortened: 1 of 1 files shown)


dustymabe:
should we take this as an opportunity to update the FCCT `version: 1.0.0` ?

modules/ROOT/pages/auto-updates.adoc
```diff
@@ -54,10 +56,12 @@ storage:
       contents:
         inline: |
           [updates]
-          strategy = "fleet_lock"
-          [updates.fleet_lock]
-          base_url = "https://example.com/lock-manager/"
+          strategy = "periodic"
+          [[updates.periodic.window]]
```
dustymabe, line 60: I need to learn more TOML. The `[[ ]]`` syntax is new to me.

dustymabe:
any plans for `strategy = "fleet_lock"` in the future?

lucab:
> any plans for strategy = "fleet_lock" in the future?

Do you mean showing both strategies in the docs here?

I'm not planning to duplicate the whole Zincati docs here, but only to show what exists and how to apply customization via fcct snippets. There is a pointer at the bottom to the whole Zincati docs.

As there are many more moving parts involved in the `fleet_lock` flow, perhaps we can fit that into the new Tutorial section.

modules/ROOT/pages/auto-updates.adoc
lucab: That's an entry in an [array of tables](https://github.com/toml-lang/toml/blob/master/toml.md#array-of-tables).

PR #119, opened 2020-07-30
faq: explain differences between Ignition and fcct configs

lucab:
This adds a FAQ entry explaining that Ignition and fcct configurations
are different interfaces (not just JSON-vs-YAML for the same content).

### DIFF
--- a/modules/ROOT/pages/faq.adoc
+++ b/modules/ROOT/pages/faq.adoc
@@ -239,3 +239,20 @@ systemd:
 The x86_64 images we provide can be used for either BIOS (legacy) boot or UEFI boot. They contain a hybrid BIOS/UEFI partition setup that allows them to be used for either. The exception to that is the `metal4k` 4k native image, which is targeted at disks with 4k sectors and https://github.com/coreos/coreos-assembler/blob/12029fea7798fa5d3535eafcf8c3d02f9a6095e4/src/cmd-buildextend-metal#L200-L202[does not have a BIOS boot partition] because 4k native disks are https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/hard-drives-and-partitions#advanced-format-drives[only supported with UEFI].
 
 https://discussion.fedoraproject.org/t/are-fedora-coreos-disk-images-hybrid-bios-uefi-bootable/21911[Discuss]
+
+== What's the difference between Ignition and FCCT configurations?
+
+Ignition configuration is a low-level interface used to define the whole set of customizations for an instance.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/faq.adoc
```diff
@@ -239,3 +239,20 @@ systemd:
...
+Many high-level tools exist that can produce an Ignition configuration starting from their own specific input formats,
+such as `terraform`, `matchbox`, `openshift-installer`, and `fcct`.
+
+Fedora CoreOS Configuration Transpiler (`fcct`) is one of such high-level tools.
+It is primarily meant as a human-friendly interface, thus defining its own richer configuration entries and using YAML documents as input.
+This YAML configuration is never directly processed by FCOS instances (only the resulting Ignitiong configuration is).
+
+Although similar, structure and entries differ between Ignition configurations and FCCT ones.
+That is, those YAML and JSON configurations contain different information and data-structures, which are not interchangeable.
+Additionally, the different formats (YAML vs JSON) make it easier to avoid mixing up inputs by mistake.
```
dustymabe, line 258: This paragraph twisted my mind a bit. Maybe something like:

_Although they look like they have the same structure, converting from an FCCT configuration to an Ignition configuration is not as simple as converting YAML to JSON. FCCT has extra user friendly and distribution specific helpers that are not present in Ignition and make the formats not interchangeable. Additionally, keeping FCCT in YAML and Ignition in JSON helps to avoid mixing up inputs by mistake._

You don't need to take that paragraph as is. It's just an example of something I find easier to comprehend what you're saying.

dustymabe:
I wonder if we should mention in here the relationship between spec versions. Somebody was asking me recently.

I think Ignition spec 3.x.y corresponds to fcct spec 1.x.y (3.0.0 ~ 1.0.0, 3.1.0 ~ 1.1.0, etc). Is that right?

lucab:
@dustymabe good point. I'm not sure where we note down fcct<->Ignition mapping, but I can try to track that down. However I'd prefer to keep FAQ straight to the point. Let's add another "how do the map" entry after this.

modules/ROOT/pages/faq.adoc
lucab: I see. I'll try to reword that. To be honest, your suggestion is not too verbose and I could basically just blend that in.

bgilbert:
There's no general mapping rule, but we can document a mapping table.  Each FCC spec doc mentions the corresponding Ignition version.

lucab:
Amended to reword the last paragraph. Moved the schema mapping topic to https://github.com/coreos/fcct/issues/121.

Issue #117, opened 2020-07-29
Live PXE/ISO improvements


bgilbert:
- [ ] Live PXE/ISO is awkwardly documented at the bottom of a page that bills itself as "Installing on Bare Metal".  Maybe it should be a separate page, or maybe the "installing" part should be renamed.

- [ ] We don't document `coreos-installer iso embed` anywhere.

- [ ] We should highlight `coreos-installer download`.

- [ ] We don't document initrd-embedded Ignition configs anywhere except bullet 4 [here](https://github.com/coreos/fedora-coreos-docs/blob/master/modules/ROOT/pages/migrate-cl.adoc#operator-notes).  See https://github.com/coreos/coreos-installer/issues/322.

- [ ] We should document hooking or replacing `coreos-installer.service`.  In the latter case, we should figure out what to recommend.  Units should at least pull in `coreos-installer-pre.target`.

lucab:
I was already thinking about the first part, specifically the "boot a live instance by pointing iPXE to our bucket". Overall, my idea was to apply the following structure to the content:

 * Provisioning Machines
   * Booting on AWS
   * ...
 * Provisioning Transient Systems
   * Live-booting from ISO
   * Live-booting via iPXE
   * Live-booting via PXELINUX
 * Imaging Bootable Disks
   * Installing from a container
   * Installing from live-ISO
   * Installing from live-PXE

travier:
- [ ] Document: `coreos-installer install --{append,delete}-kargs`

- [ ] Document automated installation via PXE
- [ ] Document automated installation via ISO

lucab:
I started following up on this with https://github.com/coreos/fedora-coreos-docs/issues/117.

bgilbert:
Some updates in https://github.com/coreos/fedora-coreos-docs/pull/355.

PR #121, opened 2020-07-31
provisioning: add instructions for live-PXE via iPXE

lucab:
This adds a new guide showing how to provision a transient FCOS
instance by live-PXE booting it via iPXE.

Ref: https://github.com/coreos/fedora-coreos-docs/issues/117

### DIFF
--- a/modules/ROOT/pages/live-booting-ipxe.adoc
+++ b/modules/ROOT/pages/live-booting-ipxe.adoc
@@ -1,0 +1,42 @@
+= Booting live Fedora CoreOS via iPXE
+
+This guide shows how to boot a transient Fedora CoreOS (FCOS) system via iPXE. By default this will run FCOS in a stateless way, completely out of RAM. Separately, FCOS can also be installed to disk.
+
+== Prerequisite
+
+Before booting FCOS, you must have an Ignition configuration file and host it somewhere (e.g. on a reachable web server). If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File].
+
+NOTE: The live system requires at least 3 GiB of RAM.
+
+== PXE images
+
+include::pxe-artifacts.adoc[]
+
+== Setting up the Boot Script
+
+An iPXE-capable machine needs to be provided with a relevant Boot Script to fetch and load FCOS artifacts.
+
+The example below shows how to load those directly from Fedora infrastructure. For performance and reliability reasons it is recommended to mirror them on the local infrastructure, and then tweak the `BASEURL` as needed.
+
+[source]
+----
+#!ipxe
+
+set STREAM stable
+set VERSION 32.20200726.3.0
+set CONFIGURL https://example.com/config.ign
+
+set BASEURL https://builds.coreos.fedoraproject.org/prod/streams/${STREAM}/builds/${VERSION}/x86_64
+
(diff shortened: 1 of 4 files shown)


lucab:
I manually tested this with testing-devel artifacts on Vultr infrastructure in preparation for https://github.com/coreos/fedora-coreos-tracker/issues/390.

The example is referencing the next stable release https://github.com/coreos/fedora-coreos-streams/issues/158, which is the first one with a the initramfs/rootfs split. The VERSION variable may need to adjusted if my guesstimation turns out to be off.

modules/ROOT/pages/live-booting-ipxe.adoc
```diff
@@ -0,0 +1,57 @@
...
+initrd ${BASEURL}/fedora-coreos-${VERSION}-live-rootfs.x86_64.img
+
+boot
+----
+
+== Update process
+
+Since the traditional FCOS upgrade process requires a disk, live-PXE systems are not able to auto-update in place. For this reason, Zincati is not running there.
+
+Instead, it is reccomended that images references in the PXE configuration are regularly refreshed. Once infrastructure and configurations are updated, the live-PXE instance simply needs to be rebooted in order to start running the new FCOS version.
```
dustymabe, line 57: Is there any way in the ipxe environment/config to do some logic to figure out what the latest is and use that? That would allow people to follow the latest without having to update their configs. The flip side of that is that someone could automate generating the ipxe config and just substitute in the latest version.

modules/ROOT/pages/live-booting-ipxe.adoc
```diff
@@ -0,0 +1,57 @@
...
+- September 22: `rootfs` image required in `testing` stream
+- October 6: `rootfs` image required in `stable` stream
+
+To boot with the `rootfs` artifact, make one of the following changes to your PXE config:
+
+- Specify both `initramfs` and `rootfs` files as initrds in your PXE configuration.  In PXELINUX, put both file paths in the `initrd` directive, separated by commas, or specify two `initrd=` parameters in your `append` line.
+- Concatenate the `initramfs` and `rootfs` files together, and specify the combined file as the initrd.
+- Specify only the `initramfs` file as the initrd, and pass an HTTP(S) URL for the `rootfs` using the `coreos.live.rootfs_url=` kernel argument.
+
+For more information on this change, see the https://github.com/coreos/fedora-coreos-tracker/issues/390[tracker issue].
```
dustymabe, line 28: This looks similar to the content from https://github.com/coreos/fedora-coreos-docs/pull/112. Should we just have it in one place and reference it?

lucab: Yes, it is exactly the same content, I've been a bit lazy.

modules/ROOT/pages/live-booting-ipxe.adoc
lucab: Not that I've seen. Plus I'm not sure that would be a very good approach. We somehow consciously decided to steer away from the Container Linux way (having a `current` directory and not encoding version into file name), and this mostly follows from it.

What we could add here as a followup is showing how to mirror the current tip of a stream via `coreos-installer download`. That should make it slightly easier to automate and refresh (not fully though).

lucab:
Rebased to avoid duplicated content.

PR #136, opened 2020-08-06
Tutorials update

travier:
First draft for the tutorial updates:

  * Splits the initial setup into its own page
  * Keep updates enabled by default
  * General rewording and update
  * Rename tutorials to remove "complexity" and mention content instead

PR #139, opened 2020-08-07
Fix systemd tty service name typo

paulnivin:
Fix typo in example FCCT config such that the resulting ign produced by fcct works as intended


lucab:
I'm not sure this is a typo, the TTY name depends on the node setup (i.e. serial console vs graphic console) if I remember correctly.

/cc @dustymabe

travier:
Thanks, this is a typo in this case. This will be updated as part of my PR to update the tutorials: https://github.com/coreos/fedora-coreos-docs/pull/136

travier:
Fixed by #136. Thanks for the PR!

dustymabe:
Thanks @paulnivin !

PR #144, opened 2020-08-07
Document installing from the container and the metal image itself

jlebon:
```
commit 94e659e5e0d4947cb593403209bffacce343730e
Date:   Fri Aug 7 11:27:11 2020 -0400

    bare-metal: use `--ignition-url`

    No need to separately `curl` the Ignition config now.

commit fdb5a88db254bb914ea8e2a3e249cdc8ad17511d
Date:   Fri Aug 7 11:27:50 2020 -0400

    bare-metal: document installing from the container

commit f8521b71004774b34ca5045a2f764c893d670ff5
Date:   Fri Aug 7 11:28:19 2020 -0400

    bare-metal: document the metal image

    E.g. downloading it directly for mirroring or for offline environments
    where they don't to/can't use our live environments for some reason.
    Also add docs about 512b vs 4k native disks and a note about dual
    UEFI/BIOS.

    Fixes: #78
```

### DIFF
--- a/modules/ROOT/pages/bare-metal.adoc
+++ b/modules/ROOT/pages/bare-metal.adoc
@@ -1,10 +1,11 @@
 :experimental:
 = Installing CoreOS on Bare Metal
 
-This guide provides instructions to install Fedora CoreOS to bare metal. Two options are available:
+This guide provides instructions to install Fedora CoreOS to bare metal. Three options are available:
 
 * Installing from live ISO
 * Installing from PXE
+* Installing from the container
 
 == Prerequisite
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/bare-metal.adoc
```diff
@@ -53,6 +54,34 @@ IPAPPEND 2
...
 For more details on how to use this information, see this https://dustymabe.com/2019/01/04/easy-pxe-boot-testing-with-only-http-using-ipxe-and-libvirt/[blog post] for testing a PXE installation via a local VM and `libvirt`.
 
+== Installing from the container
+
+You can use the `coreos-installer` https://quay.io/repository/coreos/coreos-installer[container] from an existing system to install to an attached block device. For example (substitute `docker` for `podman` if needed):
+
+[source, bash]
+----
+sudo podman run --pull=always --privileged --rm \
+    -v /dev:/dev -v /run/udev:/run/udev -v .:/data \
```
dustymabe, line 64:
```suggestion
    -v /dev:/dev -v /run/udev:/run/udev -v ${PWD}:/data \
```

`${PWD}` is a little more explicit.

dustymabe:
>  E.g. downloading it directly for mirroring or for offline environments
>     where they don't to/can't use our live environments for some reason.
>     Also add docs about 512b vs 4k native disks and a note about dual
>     UEFI/BIOS.

`where they don't to/can't` - typo?

modules/ROOT/pages/bare-metal.adoc
jlebon: Fixed!

jlebon:
> I feel like maybe we should have a section at the top of the page that briefly mentions BIOS vs UEFI and 4kn vs 512b disks and that for the most part the installer will do the right thing for you.

I think I agree, though we should probably hold off on that until we refactor as discussed in https://github.com/coreos/fedora-coreos-docs/issues/117#issuecomment-666220590.

dustymabe:
> > I feel like maybe we should have a section at the top of the page that briefly mentions BIOS vs UEFI and 4kn vs 512b disks and that for the most part the installer will do the right thing for you.
> 
> I think I agree, though we should probably hold off on that until we refactor as discussed in [#117 (comment)](https://github.com/coreos/fedora-coreos-docs/issues/117#issuecomment-666220590).

👍 - as long as that is tracked somewhere.

PR #146, opened 2020-08-07
networking: Document how to customize NIC names

kelvinfan001:
Document how to customize NIC names using link files, udev rules, and kernel arguments.

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -23,6 +23,7 @@
 *** xref:running-containers.adoc[Running Containers]
 *** xref:authentication.adoc[Configuring Authentication]
 *** xref:hostname.adoc[Setting a Hostname]
+*** xref:customize-nic.adoc[How to Customize a NIC Name]
 ** OS updates
 *** xref:update-streams.adoc[Update Streams]
 *** xref:auto-updates.adoc[Auto-Updates]
--- a/modules/ROOT/pages/customize-nic.adoc
+++ b/modules/ROOT/pages/customize-nic.adoc
@@ -1,0 +1,49 @@
+:experimental:
+= How to Customize a NIC Name
+
+== Using a systemd Link File
+You can create a systemd xref:https://www.freedesktop.org/software/systemd/man/systemd.link.html[link file] with Ignition configs.
+
+For example, to name NIC with the MAC address `12:34:56:78:9a:bc` to "infra", place a systemd link file at `/etc/systemd/network/25-infra.link` using the xref:fcct-config.adoc[FCCT] configuration snippet shown below:
+
+.Example: Customize NIC via systemd Link File
+[source,yaml]
+----
+storage:
+  files:
+    - path: /etc/systemd/network/25-infra.link
+      mode: 0644
+      contents:
+        inline: |
+          [Match]
+          MACAddress=12:34:56:78:9a:bc
+          [Link]
+          Name=infra
+----
+
+== Using Udev Rules
+Similarly, also through Ignition configs, to name NIC with the MAC address `12:34:56:78:9a:bc` to "infra", create a xref:https://linux.die.net/man/8/udev[udev rule] at `/etc/udev/rules.d/80-ifname.rules` using the xref:fcct-config.adoc[FCCT] configuration snippet shown below:
+
+.Example: Customize NIC via Udev Rules
+[source,yaml]
+----
+storage:
+  files:
+    - path: /etc/udev/rules.d/80-ifname.rules
+      mode: 0644
+      contents:
+        inline: |
+          SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="12:34:56:78:9a:bc", ATTR{type}=="1", NAME="infra"'
+----
+
+== Networking in the Initramfs via Kernel Arguments
+If networking in the initramfs is required, the kernel argument `ifname=` will dynamically create a udev rule to change the name of a NIC.
+
+Currently, unlike other parts of the networking config from the initramfs (e.g. static IPs, hostnames, etc.), these udev rules are not persisted into the real root. If the custom name needs to be applied to the real root, either a link file or udev rule must be created, as shown above. See xref:https://github.com/coreos/fedora-coreos-tracker/issues/553[this issue] for more details.
+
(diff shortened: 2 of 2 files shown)


kelvinfan001:
Should this be under the "Networking" page?

modules/ROOT/nav.adoc
```diff
@@ -22,6 +22,7 @@
 *** xref:sysctl.adoc[Kernel Tuning]
 *** xref:running-containers.adoc[Running Containers]
 *** xref:authentication.adoc[Configuring Authentication]
+*** xref:customize-nic[How to Customize a NIC Name]
```
lucab, line 25: This is missing the `.adoc` extension.

modules/ROOT/pages/customize-nic.adoc
```diff
@@ -0,0 +1,48 @@
...
+  files:
+    - path: /etc/udev/rules.d/80-ifname.rules
+      mode: 0644
+      contents:
+        inline: |
+          SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="12:34:56:78:9a:bc", ATTR{type}=="1", NAME="infra"'
+----
+
+== Networking in the `initramfs` via Kernel Arguments
+If networking in the `initramfs` is required, the `dracut` argument `ifname=` dynamically creates a udev rule to change the name of a NIC. This gets applied via `ifname-genrules.sh` from the `40network` dracut module which creates a rule in the `initramfs`.
```
lucab, line 40: The second sentence here is an implementation detail (it can slightly change at any point without effects on the feature). It can be dropped, also to avoid overloading the reader.

modules/ROOT/nav.adoc
kelvinfan001: oops!

modules/ROOT/pages/customize-nic.adoc
```diff
@@ -0,0 +1,48 @@
+:experimental:
+= How to Customize a NIC Name
+
+== Using a `systemd` Link File
+You can create a `systemd` xref:https://www.freedesktop.org/software/systemd/man/systemd.link.html[link file] with `ignition` configs.
```
sohankunkerkar, line 5: `s/ignition/Ignition`

sohankunkerkar: For your reference: https://github.com/coreos/fedora-coreos-config/pull/344#discussion_r435419740

modules/ROOT/pages/customize-nic.adoc
```diff
@@ -0,0 +1,48 @@
...
+      contents:
+        inline: |
+          [Match]
+          MACAddress=12:34:56:78:9a:bc
+          [Link]
+          Name=infra
+----
+
+== Using Udev Rules
+Similarly, also through `ignition` configs, to name NIC with the MAC address `12:34:56:78:9a:bc` to "infra", create a xref:https://linux.die.net/man/8/udev[udev rule] at `/etc/udev/rules.d/80-ifname.rules` using the xref:fcct-config.adoc[FCCT] configuration snippet shown below:
```
sohankunkerkar, line 25: same as above

modules/ROOT/pages/customize-nic.adoc
kelvinfan001: thanks!

modules/ROOT/pages/customize-nic.adoc
```diff
@@ -0,0 +1,49 @@
...
+.Example: Customize NIC via Udev Rules
+[source,yaml]
+----
+storage:
+  files:
+    - path: /etc/udev/rules.d/80-ifname.rules
+      mode: 0644
+      contents:
+        inline: |
+          SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="12:34:56:78:9a:bc", ATTR{type}=="1", NAME="infra"'
```
dustymabe, line 36: off topic: is `ATTR{type}=="1"` a way to match based on the type of networking interface?

kelvinfan001: :open_mouth: 
I believe it is. This would help a lot with supporting custom names in networking devices in console-login-helper-messages!

modules/ROOT/pages/customize-nic.adoc
```diff
@@ -0,0 +1,49 @@
...
+----
+
+== Networking in the Initramfs via Kernel Arguments
+If networking in the initramfs is required, the Dracut argument `ifname=` dynamically creates a udev rule to change the name of a NIC.
```
dustymabe, line 40: It feels like we might need to give the user some more background information. Maybe something like:

```suggestion
If networking in the initramfs is required, the kernel argument `ifname=` will dynamically create a udev rule to change the name of a NIC in the initramfs. If the custom name needs to be applied to the real root then a systemd .link file or udev rule, as discussed earlier on this page, will need to get created via Ignition. See [this issue](https://github.com/coreos/fedora-coreos-tracker/issues/553) for more details.
```

dustymabe: ahh. I see this information is in the last paragraph. Maye we can add some of this down there and add a link to the issue down there so people can get more context on why they need to do both.

modules/ROOT/pages/customize-nic.adoc
```diff
@@ -0,0 +1,49 @@
...
+        inline: |
+          SUBSYSTEM=="net", ACTION=="add", DRIVERS=="?*", ATTR{address}=="12:34:56:78:9a:bc", ATTR{type}=="1", NAME="infra"'
+----
+
+== Networking in the Initramfs via Kernel Arguments
+If networking in the initramfs is required, the kernel argument `ifname=` will dynamically create a udev rule to change the name of a NIC.
+
+Currently, unlike other parts of the networking config from the initramfs (e.g. static IPs, hostnames, etc.), these udev rules are not persisted into the real root. If the custom name needs to be applied to the real root, either a link file or udev rule must be created, as shown above. See xref:https://github.com/coreos/fedora-coreos-tracker/issues/553[this issue] for more details.
+
+For example, to name NIC with the MAC `12:34:56:78:9a:bc` to "infra", provide the argument `ifname=infra:12:34:56:78:9a:bc`. A udev rule like the following would be created in the initramfs:
```
dustymabe, line 44:
```suggestion
For example, to give the NIC with the MAC `12:34:56:78:9a:bc` a name of "infra", provide a `ifname=infra:12:34:56:78:9a:bc` kernel argument. A udev rule would be created in the initramfs like:
```

Issue #43, opened 2020-02-11
user-guides: document wireguard setup via Ignition


lucab:
This is splitting off my comment from https://github.com/coreos/fedora-coreos-tracker/issues/362#issuecomment-583320187:

> Specifically, I'm interested in reaching a final UX flow where Ignition writes NM wireguard profiles (keyfiles) and secret material, and tunnel provisioning works without further custom scripting.

We should have a doc page showing how to perform this with `fcct` snippets.

(JL: Put under System Configuration in new "Configuring Network" subheading)

jdoss:
I'll take this one.

jlebon:
@jdoss, are you still planning to work on this issue?

jdoss:
Yep I am going to do it tomorrow during the documentation hackfest at Nest.

jdoss:
This can be closed as the docs were merged.

PR #150, opened 2020-08-07
Configuring WireGuard Documentation

jdoss:
This adds documentation for configuring WiredGuard on Fedora CoreOS.

### DIFF
--- a/modules/ROOT/pages/sysconfig-configure-wireguard.adoc
+++ b/modules/ROOT/pages/sysconfig-configure-wireguard.adoc
@@ -1,0 +1,261 @@
+:experimental:
+= Configuring FCOS to use WireGuard
+
+== Introduction
+
+https://www.wireguard.com/[WireGuard] is a novel VPN that runs inside the Linux Kernel and uses state-of-the-art cryptography. It aims to be faster, simpler, leaner, and more useful than IPSec, while avoiding the massive headache. It intends to be considerably more performant than OpenVPN. WireGuard is designed as a general purpose VPN for running on embedded interfaces and super computers alike, fit for many different circumstances. It runs over UDP.
+
+- https://www.wireguard.com/[Conceptual Overview]
+- https://www.wireguard.com/quickstart/[Quickstart]
+- https://www.wireguard.com/papers/wireguard.pdf[Whitepaper]
+
+FCOS has full support for WireGuard out of the box. This guide is going to demonstrate how to setup a single connection between a FCOS server and one client computer. It goes over the basic client configuration, but it does not cover installing WireGuard on your clients.
+
+== Generate Keys
+
+You will need to generate some keys to configure WireGuard. For this guide, the keys should be pre-generated on your workstation. First, let's create the FCOS WireGuard keys:
+
+.Generate FCOS WireGuard keys
+[source,bash]
+----
+$ umask 077
+$ wg genkey | tee privatekey | wg pubkey > publickey
+----
+
+These keys will be referenced as `fcos_public_key` and `fcos_private_key` from here on out in this guide.
+
+Now let's generate the client keys:
+
+.Generate Client One WireGuard keys
+[source,bash]
+----
+$ umask 077
+$ wg genkey | tee privatekey | wg pubkey > publickey
+----
+
+These keys will be referenced as `client_one_public_key` and `client_one_private_key` from here on out in this guide.
+
+Now create a PresharedKey:
+
+.Generate a Preshared key per peer pair
+[source,bash]
+----
+$ wg genpsk > fcos_client_one_psk
+----
+
+The PresharedKey will be referenced as `fcos_client_one_psk` from here on out in this guide.
+
+NOTE: The `wg genpsk` command generates a PresharedKey that can only be used once per peer pair. Every peer you add to the FCOS server will need to generate a unique PresharedKey.
+
+== Configure WireGuard on FCOS
+
+You can now configure your Ignition config to create the `wg0` configuration file:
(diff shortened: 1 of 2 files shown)


jdoss:
Fixes #43

modules/ROOT/pages/wireguard-config.adoc
```diff
@@ -0,0 +1,259 @@
...
+
+- https://www.wireguard.com/[Conceptual Overview]
+- https://www.wireguard.com/quickstart/[Quickstart]
+- https://www.wireguard.com/papers/wireguard.pdf[Whitepaper]
+
+This guide is going to demonstrate how to setup a single connection between a FCOS server and one client computer. It goes over the basic client configuration, but it does not cover installing WireGuard on your clients.
+
+== Generate Keys
+
+You will need to generate some keys to configure WireGuard. First create the FCOS WireGuard keys:
```
miabbott, line 16: Should these keys be generated on the FCOS machine?  On your workstation?  Somewhere else?

modules/ROOT/pages/wireguard-config.adoc
```diff
@@ -0,0 +1,259 @@
...
+
+== Introduction
+
+https://www.wireguard.com/[WireGuard] is a novel VPN that runs inside the Linux Kernel and uses state-of-the-art cryptography (the "Noise" protocol). It aims to be faster, simpler, leaner, and more useful than IPSec, while avoiding the massive headache. It intends to be considerably more performant than OpenVPN. WireGuard is designed as a general purpose VPN for running on embedded interfaces and super computers alike, fit for many different circumstances. It runs over UDP.
+
+- https://www.wireguard.com/[Conceptual Overview]
+- https://www.wireguard.com/quickstart/[Quickstart]
+- https://www.wireguard.com/papers/wireguard.pdf[Whitepaper]
+
+This guide is going to demonstrate how to setup a single connection between a FCOS server and one client computer. It goes over the basic client configuration, but it does not cover installing WireGuard on your clients.
```
miabbott, line 12: I think this could be rephrased to be a bit more clear.  For example, it would be useful to say that WireGuard is already included on FCOS, so no installation is necessary there.  

Maybe frame the rest of the guide where the FCOS system is acting a peer for a WG connection; users will need to setup a second FCOS system (or other system) to act as the other peer for the WG connection.

jdoss: This guide is more geared for setting up a WireGuard VPN with FCOS as the server and you can configure your clients (laptop, phone, etc) to connect to it. 

I will add in a line about FCOS having WireGuard support out of the box.

jdoss:
Let's get this merged eh?

lucab:
This is a useful tutorial, but it goes in the opposite direction of what #43 is asking for.

In particular this bypasses NM, plus it runs a service directly on the host. The former is already know to cause issues, see https://wiki.archlinux.org/index.php/WireGuard#Routes_are_periodically_reset. The latter is a general trend we are pushing *against* with FCOS.

As as already stated, we are really aiming at "making the flow work with NM" and not "bypass NM with an on-host service".

zx2c4:
Not sure what these details are about, but I should note that WireGuard itself does *not* use any services or daemons.

cgwalters:
> In particular this bypasses NM, plus it runs a service (oneshot, wg-quick) directly on the host. 

Hmm it looks like we ended up shipping `wg-quick@.service` as part of FCOS...it's not clear to me we actually debated that detail in the original issue.

From a quick look at this it does seem like it's documenting what we are shipping in FCOS today; though I would probably agree that this should also include the recommendation to tell NM to not manage these interfaces.  (Although `wg-quick@` itself could do that?)

Actually for a long time what I really think Linux should have is generic metadata (like xattrs) on network interfaces so e.g. projects like podman/docker can add an "owner" field on the interface and if present that would tell projects like NetworkManager to ignore it.

cgwalters (approved):
General approval; needs a rebase.

jdoss:
@cgwalters Thanks for the approval. I have rebased as requested.

modules/ROOT/pages/sysconfig-configure-wireguard.adoc
```diff
@@ -0,0 +1,261 @@
...
+version: 1.1.0
+storage:
+  files:
+    - path: /etc/wireguard/wg0.conf
+      mode: 0600
+      contents:
+        inline: |
+          [Interface]
+          Address = 192.168.71.1/24,fdc9:3c6b:21c7:e6bd::1/64
+          PrivateKey = <fcos_private_key>
```
lucab, line 67: Wishlist: this mixes secrets and config in the same file. It would be great to have `FooKey = /my-path.key` supported by `wg-quick` so that content could be sanely split.

lucab:
@jdoss would you mind squashing commits too?

I still believe this kind of "how to setup an on-host service" is not the kind of documentation we want for FCOS, but if people find this content useful I'm ok to merge it now and refine later with NM instructions.

modules/ROOT/pages/sysconfig-configure-wireguard.adoc
zx2c4: You can email the wireguard mailing list with proposals. (Here is not the appropriate venue for feature requests.)

There's also already guidance on how to do this in the wg-quick(8) man page. But again, that's outside the scope of this PR, so I won't follow up further messages on it. Please use the mailing list.

jdoss:
@lucab done.

dustymabe:
Thanks @jdoss!

PR #154, opened 2020-08-09
Fix misspellings

rffontenelle:


build.sh
```diff
@@ -14,7 +14,7 @@ if [ "$(uname)" == "Darwin" ]; then
 
 elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
     # Running on Linux.
-    # Check whether podman is available, else faill back to docker
+    # Check whether podman is available, else fail back to docker
```
travier, line 17: I think that should be 'fall' here.

rffontenelle: Duh. You're totally right. Fixed and force-pushed a change in my commit.

PR #166, opened 2020-08-27
remove quotes from disk size

gabx:
- Setting the disk size between quotes output an error. Remove the quotes around the integer.
- Don't over complicate with environment variables
- `-r ` for ram is depreciated. Use `---memory` instead
- Good practice is to set `--vcpus `explicitly

### DIFF
--- a/modules/ROOT/pages/getting-started-libvirt.adoc
+++ b/modules/ROOT/pages/getting-started-libvirt.adoc
@@ -17,7 +17,7 @@ IGNITION_CONFIG="/path/to/example.ign"
 IMAGE="/path/to/image.qcow2"
 VM_NAME="fcos-test-01"
 RAM_MB="2048"
-DISK_GB="10"
+DISK_GB=10
 
 virt-install --connect qemu:///system -n "${VM_NAME}" -r "${RAM_MB}" --os-variant=fedora32 \
         --import --graphics=none --disk "size=${DISK_GB},backing_store=${IMAGE}" \
--- a/modules/ROOT/pages/getting-started-libvirt.adoc
+++ b/modules/ROOT/pages/getting-started-libvirt.adoc
@@ -19,9 +19,11 @@ VM_NAME="fcos-test-01"
 RAM_MB="2048"
 DISK_GB=10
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/getting-started-libvirt.adoc
```diff
@@ -17,11 +17,13 @@ IGNITION_CONFIG="/path/to/example.ign"
 IMAGE="/path/to/image.qcow2"
 VM_NAME="fcos-test-01"
 RAM_MB="2048"
-DISK_GB="10"
+DISK_GB=10
```
cgwalters, line 20: I don't understand, what broke with this?

modules/ROOT/pages/getting-started-libvirt.adoc
```diff
@@ -17,11 +17,13 @@ IGNITION_CONFIG="/path/to/example.ign"
...
 
-virt-install --connect qemu:///system -n "${VM_NAME}" -r "${RAM_MB}" --os-variant=fedora32 \
-        --import --graphics=none --disk "size=${DISK_GB},backing_store=${IMAGE}" \
-        --qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=${IGNITION_CONFIG}"
+virt-install --connect qemu:///system --name fcos-test-01 --memory 2048 --os-variant=fedora32 \
+        --import --graphics=none vcpus=2,maxcpus=4 --disk size=10 \
+        format=qcow2,backing_store=/path/to/image.qcow2 \
+        --qemu-commandline="-fw_cg
```
cgwalters, line 25: Looks like this lost the `f` in `fw_cfg` at least.

modules/ROOT/pages/getting-started-libvirt.adoc
gabx: I was unable to reproduce my error when using variables. So this commit has to be deleted

modules/ROOT/pages/getting-started-libvirt.adoc
gabx: yes, the format of this line is incorrect in my commit. Let's keep your variables, even if I think it is not needed and simpler to pass arguments directly in the command.

Issue #131, opened 2020-08-04
Have dedicated page to document storage-related modifications


jlebon:
Right now, we have some storage examples nested in the "FCCT Specification" page. We should split that out into a dedicated "Configuring Storage" page. Examples of what this page could have:
- configuring `/var` (or submountpoints) on a separate partition on the primary disk
- configuring `/var` (or submountpoints) on a separate partition on a separate disk
- configuring `/var` on RAID

This will be prep as well for reconfiguring the root filesystem itself once we support that (https://github.com/coreos/fedora-coreos-tracker/issues/94).

lucab:
Prior art: https://coreos.com/os/docs/latest/mounting-storage.html

PR #167, opened 2020-08-27
Add dedicated page to configuring storage

jlebon:
Let's have a page where we list different ways one can modify the
default storage layout on FCOS. This will also include root
reprovisioning examples such as xfs → btrfs, or root on RAID1,
eventually root on LUKS, etc...

Closes: #131

### DIFF
--- a/modules/ROOT/pages/storage.adoc
+++ b/modules/ROOT/pages/storage.adoc
@@ -1,0 +1,89 @@
+= Configuring Storage
+
+Fedora CoreOS ships with a simple default storage layout: the root partition is the last one and expands to take the full size of the disk. Apart from the boot partition, all data is stored on the root partition.
+
+Below, we provide examples of various ways you can customize this.
+
+== Setting up separate /var mounts
+
+Here's an example FCC file to set up `/var` on a separate partition on the same primary disk:
+
+.Adding a /var partition to the primary disk
+[source,yaml]
+----
+variant: fcos
+version: 1.1.0
+storage:
+  disks:
+  - # The name of the primary block device. In virtio-based setups, this is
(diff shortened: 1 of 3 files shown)


jlebon:
Leaving as draft for now until https://github.com/coreos/fedora-coreos-config/pull/503 is in a release and I actually double-check all the example configs.

cgwalters:
Looks sane so far.  One thing we could do is drop things like this in a hackmd.io first and that way people can fix typos and such directly (though I didn't see any yet).

jlebon:
OK, I split out the root reprovisioning bits into https://github.com/coreos/fedora-coreos-docs/pull/178 because it hasn't hit stable yet. The rest here we can merge now.

All tested working and ready to go!

(On that subject, we should hack up a script here which would kola test all the examples.)

modules/ROOT/pages/storage.adoc
```diff
@@ -0,0 +1,87 @@
...
+  disks:
+  - # The name of the primary block device. In virtio-based setups, this is
+    # likely `/dev/vda`. Elsewhere, it's likely `/dev/sda`.
+    device: /dev/vda
+    # We do not want to wipe the partition table since this is the primary
+    # device.
+    wipe_table: false
+    partitions:
+    - size_mib: 0
+      # Start at 5G so that we leave enough space for the root partition.
```
travier, line 26: I would duplicate those comments in the config below to make sure someone only looking at the one they need would not miss them.

travier (approved):
One minor comment but otherwise LGTM. We need this as we have people asking about this on discourse.

modules/ROOT/pages/storage.adoc
jlebon: Good idea, thanks!

PR #168, opened 2020-08-28
Link to stream endpoint and mention it in AWS

cgwalters:
We want people provisioning new systems in an automated fashion
to script it by parsing the stream metadata.  Describe that more
including its URL in docs, and also mention it in the AWS section.

### DIFF
--- a/modules/ROOT/pages/getting-started.adoc
+++ b/modules/ROOT/pages/getting-started.adoc
@@ -6,6 +6,11 @@
 
 There are three Fedora CoreOS (FCOS) update streams available: `stable`, `testing`, and `next`. In general, you will want to use `stable`, but it is recommended to run some machines on `testing` and `next` as well.
 
+Each stream has a canonical URL representing its current state in JSON format.  For example, the URL for the stable stream is: https://builds.coreos.fedoraproject.org/streams/stable.json
+
+While Fedora CoreOS does automatic in-place updates, it is generally a good practice to start provisioning new machines from the latest images.
+The format of this JSON file is stable, and you can parse it to e.g. find the latest ISO/AMI/etc.
+
 For more information on streams and switching between them, see xref:update-streams.adoc[Update Streams].
 
 === Provisioning Philosophy
--- a/modules/ROOT/pages/getting-started-aws.adoc
+++ b/modules/ROOT/pages/getting-started-aws.adoc
@@ -1,6 +1,6 @@
 :page-partial:
 
-New AWS instances can be directly created and booted from public FCOS images. The correct AMI image ID for each region can be found in the https://getfedora.org/coreos/download/[download page].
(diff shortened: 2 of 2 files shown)


miabbott:
This looks fine to me, but I have a hazy memory that we didn't want to point users at `stable.json` or the like.

However, looking at the FCOS download page, we do have links to the various JSON files, so my memory must be bad.

lucab:
This PR does not have any context/reference so it's hard to judge where it's coming from and how was the initial issue framed. I'm fine to merge this as a short-term band-aid, but I have a feeling there is more to uncover behind this.

My guess is that it is the same train of discussion as https://github.com/coreos/coreos-installer/issues/202, whose usecase is to pin **references** to images (instead of the artifacts themselves).

modules/ROOT/pages/getting-started.adoc
```diff
@@ -6,6 +6,10 @@
 
 There are three Fedora CoreOS (FCOS) update streams available: `stable`, `testing`, and `next`. In general, you will want to use `stable`, but it is recommended to run some machines on `testing` and `next` as well.
 
+Each stream has a canonical URL representing its current state in JSON format.  For example, the URL for the stable stream is: https://builds.coreos.fedoraproject.org/streams/stable.json
+While Fedora CoreOS does automatic in-place updates, it is generally a good practice to start provisioning new machines from the latest "bootimages".
```
bgilbert, line 10: The term `bootimage` is not used in FCOS, since (as this points out) everyone should always start from the latest release, so let's not introduce the term here.

bgilbert:
@miabbott We don't want to point people to `meta.json` or the _release_ metadata, but we very much do want them to use the _stream_ metadata.

@lucab I don't follow the "short-term band aid" part.  AFAICT the docs change here is entirely consistent with our updates policy?

lucab:
@bgilbert that refers to the "here's a JSON, figure it out on your own" paragraph as currently worded. I feel that a longer-term/better UX may involve tweaking `coreos-installer` or figure out some explicit flow more apt to a "getting started" page.

modules/ROOT/pages/getting-started.adoc
cgwalters: Do you have an alternative suggestion?  Just ..."from the latest images"?

cgwalters:
>  I feel that a longer-term/better UX may involve tweaking coreos-installer

This is also related strongly to https://github.com/coreos/fedora-coreos-tracker/issues/235

modules/ROOT/pages/getting-started.adoc
cgwalters: Updated to use just "images".

cgwalters:
Since this had some approval, merging but we can always do more followups!

PR #171, opened 2020-09-01
Add description for powershell

benjaminpreiss:
In case of using powershell I think it would be very helpful to give people a hint that `<` will not work.

### DIFF
--- a/modules/ROOT/pages/using-fcct.adoc
+++ b/modules/ROOT/pages/using-fcct.adoc
@@ -21,6 +21,9 @@ This example uses podman, but you can use docker in a similar manner. Note that 
 .Example running `fcct` using standard in and standard out:
 `podman run -i --rm quay.io/coreos/fcct:release --pretty --strict < your_config.fcc > transpiled_config.ign`
 +
+.Same example as above just for powershell (`<` is a reserved char.):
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/using-fcct.adoc
```diff
@@ -21,6 +21,9 @@ This example uses podman, but you can use docker in a similar manner. Note that
 .Example running `fcct` using standard in and standard out:
 `podman run -i --rm quay.io/coreos/fcct:release --pretty --strict < your_config.fcc > transpiled_config.ign`
 +
+.Same example as above just for powershell (`<` is a reserved char.):
+`Get-Content your_config.fcc | podman run -i --rm quay.io/coreos/fcct:release --pretty --strict > transpiled_config.ign`
```
lucab, line 25: There are unfortunately a few more places in the docs where we show this stdin redirection, and I don't think it's good to duplicate all the examples. I'd suggest instead to add a `TIP` at the bottom here with something like:
> PowerShell has a different syntax for input redirection, and `<` is a reserved character there. Instead, you can use the `Get-Content` cmdlet to read the configuration and pass it over a pipe.

bgilbert: I agree, but let's provide a clear example inside that `TIP`.

PR #172, opened 2020-09-02
storage: add LUKS docs

arithx:
Adds two LUKS based examples, a data directory & the root filesystem.


arithx:
Docs themselves are tested but this is blocked behind a new clevis release & `fedora-coreos-config` having the `clevis*` dependencies added.

modules/ROOT/pages/storage.adoc
```diff
@@ -0,0 +1,168 @@
...
+      label: root
+----
+
+NOTE: You don't need the `path` or `with_mount_unit` keys; FCOS knows that the root partition is special and will figure out how to find it and mount it.
+
+== Encrypted storage (LUKS)
+
+Here is an example to configure a LUKS device at `/var/lib/data`.
+
+[source.yaml]
```
travier, line 136: What would the secret be for such a device?

travier (commented):
LGTM also I have not test the examples.

modules/ROOT/pages/storage.adoc
arithx: The key-file is being generated by Ignition (The functionality was added in https://github.com/coreos/ignition/pull/1055)

arithx:
Marked as ready for review now that https://github.com/coreos/fedora-coreos-config/pull/609 is up and approved (merge is waiting on additional debugging in https://github.com/coreos/fedora-coreos-tracker/issues/619)

modules/ROOT/pages/storage.adoc
```diff
@@ -128,3 +128,41 @@ storage:
...
+
+.Moving the root filesystem to LUKS
+[source.yaml]
+----
+variant: fcos
+version: 1.2.0-experimental
+storage:
+  luks:
+    - name: root
+      device: /dev/vdb
```
jlebon, line 160: I think it will be more common to reprovision LUKS in place. So this would be `/dev/disk/by-partlabel/root` (see the added test in https://github.com/coreos/fedora-coreos-config/pull/609).

modules/ROOT/pages/storage.adoc
```diff
@@ -128,3 +128,41 @@ storage:
...
+----
+variant: fcos
+version: 1.2.0-experimental
+storage:
+  luks:
+    - name: data
+      device: /dev/vdb
+  filesystems:
+    - path: /var/lib/data
+      device: /dev/disk/by-id/dm-name-data
```
jlebon, line 146: Any reason for that vs `/dev/mapper/data`? I find the latter much nicer to use. :)

modules/ROOT/pages/storage.adoc
```diff
@@ -128,3 +128,41 @@ storage:
...
+      format: xfs
+      label: DATA
+      with_mount_unit: true
+----
+
```
jlebon, line 151: Should we add some explanation test here to introduce the next example? (And e.g. link to Clevis.)

jlebon (commented):
Thanks for this! Will look at refreshing https://github.com/coreos/fedora-coreos-docs/pull/167 and getting that in next week so we can unblock this!

modules/ROOT/pages/storage.adoc
jlebon: So then we'd need `wipeVolume: true` here too like in https://github.com/coreos/fedora-coreos-config/pull/616.

arithx: Sure, swapped it over to doing an in place override.

modules/ROOT/pages/storage.adoc
arithx: Nope. I don't have a personal preference so switched them to `/dev/mapper/$device`.

jlebon (approved):
LGTM!

I added a `hold` label for the same reason as https://github.com/coreos/fedora-coreos-docs/pull/178.

I think you can rebase this on the latest master and just depend on #178 now that #167 has merged.

travier (approved):

jlebon (approved):
LGTM! Just needs a rebase.

Issue #115, opened 2020-07-29
provisioning: add a "Booting on Aliyun" page


lucab:
The "Provisioning Machines" section is currently missing a "Booting on Aliyun" page.

We do publish qcow2.xz for the `aliyun` platform, but we do need instructions on how to import the image to the cloud and launch a VM based on that.

PR #173, opened 2020-09-04
Added booting documentation for Alibaba Cloud

robertvolkmann:
Fixes: https://github.com/coreos/fedora-coreos-docs/issues/115

### DIFF
--- a/modules/ROOT/pages/provisioning-aliyun.adoc
+++ b/modules/ROOT/pages/provisioning-aliyun.adoc
@@ -1,0 +1,199 @@
+= Provisioning Fedora CoreOS on Alibaba Cloud (Aliyun)
+
+This guide shows how to provision new Fedora CoreOS (FCOS) nodes on Alibaba Cloud.  Fedora currently does not publish Fedora CoreOS images within Alibaba Cloud, so you must download an Alibaba Cloud image from Fedora and upload it to one of your Object Storage Service (OSS) buckets.
+
+== Prerequisites
+
+Before provisioning an FCOS machine, you must have an Ignition configuration file containing your customizations. If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File].
+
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/provisioning-aliyun.adoc
```diff
@@ -0,0 +1,185 @@
...
++
+.Example deleting uploaded blob
+[source, bash]
+----
+aliyun oss rm "${BUCKET_URL}/${IMAGE_BLOB}" --region "${REGION}"
+----
+
+== Creating a VSwitch (Optional)
+
+There exists no default VPCs or VSwitches in Alibaba Cloud. Hence for creating any instances a VSwitch has to exists. Pick some existing or create one with the following steps.
```
lucab, line 81: I seem to remember instances can be created in a standalone way with a floating public IP directly attached to them, am I wrong?
While I acknowledge that a VPC is a better infrastructure, standalone addressable nodes are usually easier to grasp for people that are getting started. What about documenting that first, and leaving the VPC setup to a later advanced section?

(Assuming my memory is not mixing details from other clouds, in which case just disregard this comment)

lucab:
Thanks a lot for this PR!
I don't have an AliCloud account at hand right now, but I'll get back here soon as I was planning to touch this anyway in the next weeks.

dustymabe:
Thanks so much @robertvolkmann ! 🎉🎉🎉

modules/ROOT/pages/provisioning-aliyun.adoc
robertvolkmann: I don't know anything about something like "standalone addressable nodes". I never saw such an option in the Console or in the API documentation. Currently, only VPC networks are supported, see [https://www.alibabacloud.com/help/doc-detail/61651.htm](Network Types).

robertvolkmann:
@lucab Did you get a chance to look into it?

lucab:
@robertvolkmann not yet. The company account I used initially when working on AliCloud is gone. I asked for access to another internal one we have, and I hope to be unblocked next week.

lucab (approved):

PR #177, opened 2020-09-13
Document data persistence in FAQ

### LINKED ISSUE #174: Document /etc persistence across updates and reboots in FAQ
I could not find documentation about persistence when doing updates and reboots.

This could be documented in the FAQ.

So far the information I got from IRC is that updates will preserve changes by default and that `/etc` persist across reboots.

EDIT:
Further details from [machine-config-operator] which links to the [ostree adapting exists] page.

[machine-config-operator]: https://github.com/openshift/machine-config-operator/blob/master/docs/FAQ.md#q-if-i-change-something-manually-on-the-host-will-the-mco-revert-it
[ostree adapting exists]: https://ostree.readthedocs.io/en/latest/manual/adapting-existing/

>  Stated more generally, on an OSTree managed system, all content in /etc and /var is preserved by default across upgrades.

Confusingly, the second link only state that `/var` persists.

tomleb:
Closes #174

### DIFF
--- a/modules/ROOT/pages/faq.adoc
+++ b/modules/ROOT/pages/faq.adoc
@@ -116,6 +116,17 @@ Ignition equivalent.
 
 https://discussion.fedoraproject.org/t/launch-faq-how-are-fedora-coreos-nodes-provisioned-can-i-re-use-existing-cloud-init-configurations/49/1[Discuss on discussion.fedoraproject.org]
 
+== What data persists across upgrades and reboots?
+
+The directories `/etc` and `/var` are mounted as read-write which lets users
+write and modify files.
+
+The directory `/etc` may be changed by deployments, but will not override user
+made changes. The content under `/var` is left untouched by rpm-ostree when
+applying upgrades or rollbacks. For more information, refer to the
+https://rpm-ostree.readthedocs.io/en/latest/manual/administrator-handbook/#filesystem-layout[filesystem
+layout].
+
 == How do I migrate from Container Linux to Fedora CoreOS?
 
 Migrations can be accomplished by re-provisioning the machine with


lucab:
Looks good to to me.

@jlebon @cgwalters is the wording of the `/etc` part simple enough yet still accurate?

modules/ROOT/pages/faq.adoc
```diff
@@ -116,6 +116,15 @@ Ignition equivalent.
 
 https://discussion.fedoraproject.org/t/launch-faq-how-are-fedora-coreos-nodes-provisioned-can-i-re-use-existing-cloud-init-configurations/49/1[Discuss on discussion.fedoraproject.org]
 
+== What data persists across upgrades and reboots?
+
+The directories `/etc` and `/var` are mounted as read-write which lets users
+write and modify files.
+
+The directory `/etc` may be changed by deployments, but will not override user
+made changes.  The directory `/var` is never touched by upgrades or reboots.
```
jlebon, line 125: Hmm, I think "or reboots" might be misleading. `/var` can change on a reboot into a new deployment if the upgrade e.g. added new tmpfiles entries. I think the thing we want to emphasize is that it's not changed during the upgrade operation itself. Maybe something like

```suggestion
made changes.  The directory `/var` is never touched by `rpm-ostree upgrade`.
```

?

jlebon (approved):
One comment, LGTM otherwise. Thanks for writing this!

modules/ROOT/pages/faq.adoc
lucab: Good point. I'd write that down as:
> The content under `/var` is left untouched by rpm-ostree when applying upgrades or rollbacks.

tomleb: Made the changes.

I'm not sure I understand :

> /var can change on a reboot into a new deployment if the upgrade e.g. added new tmpfiles entries.

Do you have an example?

lucab: I sadly don't have a quick example at hand, but the core is: rpm-ostree won't change `/var` directly, but anything else which is part of the upgrade could.

So, if the update comes with a tmpfiles fragment or a migration script that tweaks data under `/var`, some changes may indeed happen when you upgrade/reboot and run the new logic. This is often the case for software going through schema changes for their local data.

lucab (approved):

lucab:
@tomleb thank you very much for following up with both the ticket and this PR, LGTM!

PR #178, opened 2020-09-14
storage: add root reprovisioning examples

jlebon:
We now support reprovisioning the root filesystem. Give some basic
examples which work today. More to come as support is added.

Requires: https://github.com/coreos/fedora-coreos-docs/pull/167

This is all tested working, but we should wait until it hits the stable stream next week before merging.

### DIFF
--- a/modules/ROOT/pages/storage.adoc
+++ b/modules/ROOT/pages/storage.adoc
@@ -87,3 +87,47 @@ storage:
       format: xfs
       with_mount_unit: true
 ----
+
+== Reconfiguring the root filesystem
+
+It is possible to reconfigure the root filesystem itself. You can use the path `/dev/disk/by-label/root` to refer to the original root partition. You must ensure that the new filesystem also has a label of `/root`.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/storage.adoc
```diff
@@ -0,0 +1,131 @@
...
+.Changing the root filesystem to ext4
+[source,yaml]
+----
+variant: fcos
+version: 1.1.0
+storage:
+  filesystems:
+    - device: /dev/disk/by-partlabel/root
+      wipe_filesystem: true
+      format: ext4
```
jlebon, line 106: I had to switch back to `ext4` here from `btrfs`. I think the latter would be a nicer example given the push in Fedora, though it needs https://github.com/coreos/fedora-coreos-config/pull/590. We can switch it once that merges.

travier (approved):

jlebon:
This is in stable now. Merging!

PR #181, opened 2020-09-15
Make the bash script's style more consistent and appropriate for today's resource usage

renich:
Added proper form long options and followed general recommendations here: https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

Also suggesting a biger amount of RAM and the use of 3 virtual CPUs at least.

### DIFF
--- a/modules/ROOT/pages/getting-started-libvirt.adoc
+++ b/modules/ROOT/pages/getting-started-libvirt.adoc
@@ -16,8 +16,8 @@ coreos-installer download -s "${STREAM}" -p qemu -f qcow2.xz --decompress -C ~/.
 IGNITION_CONFIG="/path/to/example.ign"
 IMAGE="/path/to/image.qcow2"
 VM_NAME="fcos-test-01"
-VCPUS="3"
-RAM_MB="3072"
+VCPUS="2"
+RAM_MB="2048"
 DISK_GB="10"
 
 virt-install --connect="qemu:///system" --name="${VM_NAME}" --vcpus="${VCPUS}" --memory="${RAM_MB}" \
--- a/modules/ROOT/pages/getting-started-libvirt.adoc
+++ b/modules/ROOT/pages/getting-started-libvirt.adoc
@@ -21,7 +21,8 @@ RAM_MB="3072"
 DISK_GB="10"
 
 virt-install --connect="qemu:///system" --name="${VM_NAME}" --vcpus="${VCPUS}" --memory="${RAM_MB}" \
-        --os-variant="fedora32" --import --graphics="none" --disk="size=${DISK_GB},backing_store=${IMAGE}" \
+        --os-variant="fedora32" --import --graphics="none" \
+        --disk="size=${DISK_GB},backing_store=${IMAGE}" \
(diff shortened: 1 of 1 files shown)


renich:
Also, it might be a good idea to pump up the proposed RAM usage. Silverblue suggests 3072 right now but I know nothing.

And, maybe suggest 2 or 4 vCPUs?

lucab:
@renich yes thanks! A 3GiB / 2 vCPU instance looks indeed more reasonable, and keeping the actual values in variables makes it straightforward for people to tweak. Feel free to amend your commit with that.

renich (commented):
OK, how's this?

travier (approved):
LGTM

jlebon:
> Also suggesting a biger amount of RAM and the use of 3 virtual CPUs at least.

Hmm, isn't it understood that users can add more vCPUs/RAM as needed for their use case? Or were you seeing degradation of some kind with the current settings and a minimal workload? If not, I think I'd rather stick with the base resources we have now. (Note for example that a lot of our CI tests run FCOS with 1 vCPU and 1G of RAM just fine.)

renich:
> Hmm, isn't it understood that users can add more vCPUs/RAM as needed for their use case? Or were you seeing degradation of some kind with the current settings and a minimal workload? If not, I think I'd rather stick with the base resources we have now. (Note for example that a lot of our CI tests run FCOS with 1 vCPU and 1G of RAM just fine.)

I'm just saying that we should provide a sane minimum. It's a quickstart so stuff should be, more less, ready to run on non-minimal workloads; so that the user can test it without worrying about that. Maybe 2 vCPUs would be enough for this, though. But being 1.5 GiB the minimum "recommended" by libvirt when using the Fedora 32/CentOS 8/RHEL 8.3 profile, I think 1 extra gig isn't insane.

renich:
s/libvirt/virt-manager/g

jlebon:
> Maybe 2 vCPUs would be enough for this, though. But being 1.5 GiB the minimum "recommended" by libvirt when using the Fedora 32/CentOS 8/RHEL 8.3 profile, I think 1 extra gig isn't insane.

Looks like [f32 is 1G min/2G recommended](https://gitlab.com/libosinfo/osinfo-db/-/blob/master/data/os/fedoraproject.org/fedora-32.xml.in#L240-252), while el8 is indeed 1.5G min/recommended. I think it makes sense to match the recommended amount for the rest of Fedora here (and actually in https://github.com/coreos/fedora-coreos-tracker/issues/221 we're talking about eventually adding FCOS to osinfo-db). So how about we meet halfway and bump to 2 vCPUs but leave the RAM at 2G? :)

I don't want to give the impression that FCOS needs a minimum of 3G of RAM to be useful. Part of our messaging is that we're a lightweight OS.

renich:
OK, I've updated the branch.

travier (approved):

lucab:
LGTM too, leaving to @jlebon for the final pass/merge.

lucab (approved):

jlebon (approved):
Thanks for the patch!

PR #185, opened 2020-09-28
provisioning: add instructions for IBM Cloud

dustymabe:


modules/ROOT/pages/provisioning-ibmcloud.adoc
```diff
@@ -0,0 +1,106 @@
...
+ * `ibmcloud plugin install cloud-object-storage`
+ * `ibmcloud plugin install infrastructure-service`
+
+After you've logged in using `ibmcloud login` you can set a target region:
+
+.Target a specific region
+[source, bash]
+----
+REGION='us-east' # run `ibmcloud regions` to view options
+ibmcloud target -r $REGION
```
lucab, line 22: Nit: I think `ibmcloud login -r ${REGION} --apikey @my-key.json` is equivalent here and also covers the API-login step (but please double-check that it also affects commands further below).

modules/ROOT/pages/provisioning-ibmcloud.adoc
```diff
@@ -0,0 +1,106 @@
+= Provisioning Fedora CoreOS on IBM Cloud
+
+This guide shows how to provision new Fedora CoreOS (FCOS) instances in IBM Cloud VPC Generation 2.
+
+== Prerequisites
+
+Before provisioning a FCOS instance, you must have an Ignition configuration file containing your customizations. If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File].
+
+You also need to have access to an https://cloud.ibm.com/login[IBM Cloud account]. The examples below use the https://cloud.ibm.com/docs/cli?topic=cli-getting-started[`ibmcloud`] command-line tool, which must be separately installed and configured beforehand. There are also several pieces that need to be in place first, like a VPC, SSH keys, networks, permissions, etc.. Unfortunately, this guide is not a comprehensive IBM Cloud guide. If you are new to IBM Cloud please familiarize yourself using https://cloud.ibm.com/docs/vpc?topic=vpc-getting-started[the documentation for VPC Gen2] first.
```
lucab, line 9: s/etc../etc./ ?

modules/ROOT/pages/provisioning-ibmcloud.adoc
```diff
@@ -0,0 +1,106 @@
...
+[source, bash]
+----
+IMAGE=${FILE:0:-6}     # pull off .qcow2
+IMAGE=${IMAGE//[._]/-} # replace . and _ with -
+ibmcloud is image-create $IMAGE --file "cos://${REGION}/${BUCKET}/${FILE}" --os-name centos-7-amd64
+----
+
+NOTE: Specifying `--os-name centos-7-amd64` is required for now until IBM expands the list of OS types.
+
+You'll have to wait for the image creation process to and now show up as `pending`. Before you can use the image. Monitor with the following command:
```
lucab, line 66: I think something went wrong with this sentence, possibly some copy-paste mixup?

modules/ROOT/pages/provisioning-ibmcloud.adoc
dustymabe: I was trying to explicitly avoid giving the user the `ibmcloud login` command to use because they could be using an API key or they could be using a username/password. They probably do set the region when they log in, but having this as a separate command does enable us to set the `$REGION` environment varialbe, which we do use later.

modules/ROOT/pages/provisioning-ibmcloud.adoc
dustymabe: indeed :spaghetti:

will fix

dustymabe: New wording in last push. Hopefully that is better.

lucab (approved):

lucab:
I think @dustymabe wants to hold this for a bit and synchronize with the rest of the work at https://github.com/coreos/fedora-coreos-tracker/issues/277.
This LGTM as is (but I haven't tried a full provisioning flow right now, I stopped midway due to lack of ACL/time), I'll let him merge when he prefers.

travier (approved):
Not tested but LGTM.

dustymabe:
Thanks for the reviews. I'll merge after the next round of releases (which should all contain IBM Cloud images IIUC).

PR #188, opened 2020-10-02
Document the usage of toolbox

### LINKED ISSUE #187: Document toolbox
Provide a recommendation and examples for how to run additional administration tools in a privileged container.

tomleb:
Fixes #187.

This is more of a draft asking for suggestion / how it can be improved.

Went for a brief how-to with a link to the official docs.

A few things I'm not sure.. 

- Did we want this to be a section in the tutorial? 
- Should I go more in-depth in details? 
- Do we want the criterias there? I'm not even sure the criterias are official, but I took them from https://hackmd.io/3LlGUzYjT-C3sbkIDBquKQ.

Was a great opportunity to learn about toolbox, it's pretty cool.

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -32,6 +32,7 @@
 *** xref:manual-rollbacks.adoc[Manual Rollbacks]
 *** xref:access-recovery.adoc[Access Recovery]
 *** xref:emergency-shell.adoc[Emergency Shell Access]
+*** xref:debugging-with-toolbox.adoc[Debugging with Toolbox]
 * Migration notes
 ** xref:migrate-ah.adoc[Migrating from Atomic Host]
 ** xref:migrate-cl.adoc[Migrating from Container Linux]
--- a/modules/ROOT/pages/debugging-with-toolbox.adoc
+++ b/modules/ROOT/pages/debugging-with-toolbox.adoc
@@ -1,0 +1,68 @@
+= Debugging with Toolbox
+
+The FCOS image is kept minimal by design to reduce the image size and the
+attack surface. This means that it does not include every troubleshooting tools
+that a normal OS may include. Instead, the recommended approach is to leverage
+containers with the https://github.com/containers/toolbox/[toolbox] utility
+included in the image.
+
+== What is toolbox
+
+Toolbox is a utility that allows you to create privileged containers meant to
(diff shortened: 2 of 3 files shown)


cgwalters (approved):
Broadly speaking I'm OK with this, however we will have a long running tension here around upstream documentation versus ours.

cgwalters:
Also, I think what toolbox really needs is not basic usage docs so much as "war stories" - how one might use specific tools inside it to debug general classes of problems.  You noted `strace` but there's a whole lot of others, for example using the [bcc tooling](https://github.com/iovisor/bcc) or SystemTap, or use disk partitioning tooling, or `tcpdump` for network debugging etc.  Those could probably go upstream, we could just note they're possible and encouraged.

tomleb:
> Also, I think what toolbox really needs is not basic usage docs so much as "war stories" - how one might use specific tools inside it to debug general classes of problems.

I wonder if those "war stories" should be left outside of the fedora coreos docs (eg: toolbox repository/website). Maybe that's what you mean with "Those could probably go upstream".

In that case, would the current docs be sufficient if they could also link to other "war stories" kind of docs?  Or do you simply want me to add more tools example to install in this current doc?

lucab:
> This is more of a draft asking for suggestion / how it can be improved.
> Went for a brief how-to with a link to the official docs.

Thanks, it looks great! Answering your remaining questions:

> Did we want this to be a section in the tutorial?

I'd say this looks fine where it is right now. We may add a tutorial page later with a different spin, e.g. showing how to troubleshoot the network via `tcpdump` in a toolbox. I think this is a similar concept to Colin's war stories above, but we don't have to tackle it here (and maybe not in this repo at all, not sure).

> Should I go more in-depth in details?

I'd say this level of the details is fine, main goal is to show that there exists a way to perform debugging even if the packages are not directly on host, and that seems covered.

> Do we want the criterias there?

I'd personally drop that section from here. We definitely need to document them somewhere when agreed upon, but this is a doc-page for users trying to debug things on a host, no need to derail their journey into an OS development discussion.

tomleb:
> I'd personally drop that section from here. We definitely need to document them somewhere when agreed upon, but this is a doc-page for users trying to debug things on a host, no need to derail their journey into an OS development discussion.

Agreed, removed.

lucab (approved):

PR #190, opened 2020-10-05
references: document signing keys and update barriers

lucab:
This adds a reference page with a description and an example covering
signing keys and update barriers.

Refs: https://github.com/coreos/fedora-coreos-streams/pull/201#discussion_r498687448


travier (approved):

dustymabe (approved):
LGTM

Issue #192, opened 2020-10-09
coreos-installer related info missing


SaravanaStorageNetwork:
This issue is part of "Getting started" section - https://docs.fedoraproject.org/en-US/fedora-coreos/getting-started/

There is a mention to use coreos-installer to download the image. 
But, from where to download coreos-installer binary itself is not mentioned. 

Note: The path where it should be saved also needs be mentioned.

lucab:
`coreos-installer` is available as a container (or directly as a binary on FCOS), see 
https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/.

travier:
Also https://coreos.github.io/coreos-installer/getting-started/

jlebon:
Added some docs in https://github.com/coreos/fedora-coreos-docs/pull/193 which should help new users.

PR #193, opened 2020-10-09
getting-started-libvirt: mention how to get coreos-installer

jlebon:
This is the first mention of coreos-installer when reading from the
Getting Started guide. Let's provide more info re. where they can get
coreos-installer from.

Closes: #192

### DIFF
--- a/modules/ROOT/pages/getting-started-libvirt.adoc
+++ b/modules/ROOT/pages/getting-started-libvirt.adoc
@@ -1,11 +1,15 @@
 :page-partial:
 
-. Fetch the latest image suitable for the `qemu` platform (or https://getfedora.org/coreos/download/[download and verify] it from the web).
+. Fetch the latest image suitable for the `qemu` platform using `coreos-installer` (or https://getfedora.org/coreos/download/[download and verify] it from the web). You can use `coreos-installer` https://quay.io/repository/coreos/coreos-installer[as a container], or on Fedora install it from the repos.
 +
 [source, bash]
 ----
 STREAM="stable"
+# as an installed binary:
 coreos-installer download -s "${STREAM}" -p qemu -f qcow2.xz --decompress -C ~/.local/share/libvirt/images/
+# or as a container:
+podman run --pull=always --rm -v $HOME/.local/share/libvirt/images/:/data -w /data \
+    quay.io/coreos/coreos-installer:release download -s "${STREAM}" -p qemu -f qcow2.xz --decompress
 ----
 +
 


travier:
I would recommend linking to https://coreos.github.io/coreos-installer/getting-started/ instead as it has all the options for running coreos-installer in a single page.

jlebon:
> I would recommend linking to coreos.github.io/coreos-installer/getting-started instead

Yeah, I considered that, but I'm hesitant to have the "Getting Started" guide link to other pages where there's yet more information that needs to be consumed and understood.

OTOH, I agree just saying "can be run as a container" might not be enough to go on. Though the upstream docs container example is for installing, not downloading. Maybe let's add an example there for that and then we can link directly to it from the FCOS docs?

travier:
We have that in the tutorial too: https://docs.fedoraproject.org/en-US/fedora-coreos/tutorial-setup/
Maybe we should make it a separated included "page"?

travier:
Or maybe we could create a specific "How to download FCOS" section with all examples.

jlebon:
Maybe let's just get this in for now? I agree it could be improved, though at least it's an improvement on what we currently have. :)

bgilbert:
Let's at least provide an example of running from a container, since there are a couple non-obvious options.

```
STREAM="stable"
podman run --pull=always --rm -v $HOME/.local/share/libvirt/images/:/data -w /data \
    quay.io/coreos/coreos-installer:release  download -s "${STREAM}" -p qemu -f qcow2.xz --decompress
```

jlebon:
OK, done!

modules/ROOT/pages/getting-started-libvirt.adoc
```diff
@@ -1,11 +1,15 @@
...
+. Fetch the latest image suitable for the `qemu` platform using `coreos-installer` (or https://getfedora.org/coreos/download/[download and verify] it from the web). You can use `coreos-installer` https://quay.io/repository/coreos/coreos-installer[as a container], or on Fedora install it from the repos.
 +
 [source, bash]
 ----
 STREAM="stable"
+# as an installed binary:
 coreos-installer download -s "${STREAM}" -p qemu -f qcow2.xz --decompress -C ~/.local/share/libvirt/images/
+# or as a container:
+podman run --pull=always --rm -v $HOME/.local/share/libvirt/images/:/data -w /data \
+    quay.io/coreos/coreos-installer:release  download -s "${STREAM}" -p qemu -f qcow2.xz --decompress
```
bgilbert, line 12: Extra space after `release`, duly copypasted from my example.  :disappointed:

bgilbert (approved):

PR #196, opened 2020-10-15
*: adjust live RAM requirements

bgilbert:
Require 2 GiB of RAM for live PXE.  Live ISO no longer has any unusual RAM requirements, so drop the caveat entirely.

Per https://github.com/coreos/fedora-coreos-tracker/issues/407#issuecomment-680258844 and https://github.com/coreos/fedora-coreos-tracker/issues/407#issuecomment-709463014.

### DIFF
--- a/modules/ROOT/pages/bare-metal.adoc
+++ b/modules/ROOT/pages/bare-metal.adoc
@@ -21,7 +21,7 @@ podman run --privileged --pull=always --rm -v .:/data -w /data \
     quay.io/coreos/coreos-installer:release download -f iso
 ----
 +
-NOTE: The live system requires at least 3 GiB of RAM. You can boot it in either legacy BIOS or UEFI mode, regardless of what mode the OS will use once installed.
+NOTE: You can boot the live ISO in either legacy BIOS or UEFI mode, regardless of what mode the OS will use once installed.
 +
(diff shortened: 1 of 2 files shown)


Siosm (approved):

travier (approved):

dustymabe:
one caveat here. In https://github.com/coreos/fedora-coreos-tracker/issues/407#issuecomment-709463014 I did mention that I can run fine with 2iG *if* I use the `coreos.live.rootfs_url` arg. If I combine the rootfs with the initramfs to make a **fat initramfs** I needed more than 2GiB (2400MiB seemed to be the magic number). With 2GiB I get:

```
[!!!!!!] Failed to isolate default target.
[    1.338876] systemd[1]: Freezing execution.
```

bgilbert:
Ah, okay, I had understood from https://github.com/coreos/fedora-coreos-tracker/issues/407#issuecomment-680258844 that that was not the case.

dustymabe:
> Ah, okay, I had understood from [coreos/fedora-coreos-tracker#407 (comment)](https://github.com/coreos/fedora-coreos-tracker/issues/407#issuecomment-680258844) that that was not the case.

Yeah, I think the requirement may have increased slightly since then.

bgilbert:
Okay, I conditionalized the requirement to 2 GiB with rootfs fetching and 3 GiB with rootfs appending.

travier (approved):

dustymabe (approved):
LGTM

PR #198, opened 2020-10-19
troubleshooting: Add instructions for enabling kdump

kelvinfan001:
Since `kexec-tools` is currently not yet in the base image, I included a step to layer `kexec-tools`. Should I hold this PR until we've included `kexec-tools` in the base image?

### DIFF
--- a/modules/ROOT/pages/debugging-kernel-crashes.adoc
+++ b/modules/ROOT/pages/debugging-kernel-crashes.adoc
@@ -1,0 +1,27 @@
+= Debugging kernel crashes using kdump
+
+. Memory must be reserved for the crash kernel during booting of the first kernel. Kernel arguments can be provided like this:
++
+[source, bash]
+----
+sudo rpm-ostree kargs --append='crashkernel=256M'
+----
+xref:kernel-args.adoc[More information] on how to modify kargs via `rpm-ostree`.
+
+. By default, the path in which the vmcore will be saved is `/var/crash`. It is also possible to write the dump over the network or to some other location on the local system by editing `/etc/kdump.conf`. For additional information, see https://linux.die.net/man/8/mkdumprd[`mkdumprd(8)`] and the comments in `/etc/kdump.conf`.
+
+. Enable the kdump systemd service. 
++
+[source, bash]
+----
+sudo systemctl enable kdump.service
+----
+
+. Reboot your system.
++
+[source, bash]
+----
+sudo systemctl reboot
+----
+
+TIP: For additional information on how to test that kdump is properly armed and how to analyze the dump, refer to the https://fedoraproject.org/wiki/How_to_use_kdump_to_debug_kernel_crashes[kdump documentation for Fedora] and https://www.kernel.org/doc/html/latest/admin-guide/kdump/kdump.html[the Linux kernel documentation on kdump].
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -39,6 +39,7 @@
 *** xref:access-recovery.adoc[Access Recovery]
 *** xref:emergency-shell.adoc[Emergency Shell Access]
 *** xref:debugging-with-toolbox.adoc[Debugging with Toolbox]
+*** xref:debugging-kernel-crashes.adoc[Debugging Kernel Crashes]
 * Migration notes
 ** xref:migrate-ah.adoc[Migrating from Atomic Host]
 ** xref:migrate-cl.adoc[Migrating from Container Linux]


cgwalters:
It is probably simplest to ship it now and then hold the PR until it's reached stable, but you could add a note in the meantime to layer if it's not installed?

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -0,0 +1,34 @@
...
+[source, bash]
+----
+sudo rpm-ostree kargs --append='crashkernel=256M'
+----
+
+. Next, make sure the path in which the vmcore will be saved is somewhere in the `/var` directory of the FCOS stateroot. It is also possible to write the dump over the network; for additional information, consult the mkdumprd man page and the comments in /etc/kdump.conf.
++
+[source, bash]
+----
+sed -i "s/^path.*/path \/sysroot\/ostree\/deploy\/fedora-coreos\/var\/crash/" /etc/kdump.conf
```
cgwalters, line 21: That...looks like a bug?

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -0,0 +1,34 @@
...
+----
+
+. Next, reboot your system.
++
+
+. Finally, enable and activate the kdump systemd service. 
++
+[source, bash]
+----
+sudo systemctl enable kdump.service --now
```
cgwalters, line 31: This is OK but I think it'd be better to do it *before* the reboot (then you don't need `--now`, but it also doesn't hurt).

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -0,0 +1,34 @@
+= Enabling kdump to debug kernel crashes
+
+. First, install `kexec-tools` using `rpm-ostree`.
```
bgilbert, line 3: Since we're using an enumerated list, First/Next/Finally are redundant.

modules/ROOT/pages/debugging-kernel-crashes.adoc
kelvinfan001: Oh I remember why I put this here: if this were done earlier, `kexec-tools` wouldn't be layered on the system yet.

jlebon:
`kexec-tools` is added in https://github.com/coreos/fedora-coreos-config/pull/708. Maybe let's also get https://github.com/coreos/fedora-coreos-docs/pull/199 in and then this document can link to that?

dustymabe:
#199 is in now

kelvinfan001:
Link to the page in #199 added.

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -0,0 +1,40 @@
...
+sudo systemctl reboot
+----
+
+. Memory must be reserved for the crash kernel during booting of the first kernel. Kernel arguments can be provided like this:
++
+[source, bash]
+----
+sudo rpm-ostree kargs --append='crashkernel=256M'
+----
+xref:kernel-args.adoc[More information] on how to modify kargs via `rpm-ostree`.
```
kelvinfan001, line 17: Link to the page in #199 added.

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -0,0 +1,40 @@
+= Enabling kdump to debug kernel crashes
+
+. If `kexec-tools` is not present on the system, install `kexec-tools` using `rpm-ostree`, and reboot as needed.
+[source, bash]
++
+----
+sudo rpm-ostree install kexec-tools
+sudo systemctl reboot
+----
```
jlebon, line 9: I think we can drop this now that we're planning to ship `kexec-tools` in-tree.

kelvinfan001: What about for users that didn't update? But I guess everyone knows how to do an `rpm-ostree install` :)

jlebon: I think in general the docs assume that the user is at least at the latest stable release. But yeah, on that note we should hold this PR until https://github.com/coreos/fedora-coreos-config/pull/708 actually makes it to the stable stream.

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -0,0 +1,32 @@
+= Enabling kdump to debug kernel crashes
+
+. Memory must be reserved for the crash kernel during booting of the first kernel. Kernel arguments can be provided like this:
++
+[source, bash]
+----
+sudo rpm-ostree kargs --append='crashkernel=256M'
```
strigazi, line 7: The doc suggests 128, maybe we should keep that 'default'? Unless there is a reason fcos needs more memory.

https://fedoraproject.org/wiki/How_to_use_kdump_to_debug_kernel_crashes

kelvinfan001: Based off of a local qemu build of the current `testing-devel`, it seems like 128M does not work.

kelvinfan001:
@dustymabe Looks like `kexec-tools` is in the stable stream now; this should be OK to merge?

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -0,0 +1,32 @@
+= Enabling kdump to debug kernel crashes
```
jlebon, line 1: WDYT about rewording this as "Debugging kernel crashes using kdump" so that the title matches the title of the menu item more closely?

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -0,0 +1,32 @@
...
+
+. Memory must be reserved for the crash kernel during booting of the first kernel. Kernel arguments can be provided like this:
++
+[source, bash]
+----
+sudo rpm-ostree kargs --append='crashkernel=256M'
+----
+xref:kernel-args.adoc[More information] on how to modify kargs via `rpm-ostree`.
+
+. Make sure the path in which the vmcore will be saved is somewhere in the `/var` directory of the FCOS stateroot. It is also possible to write the dump over the network; for additional information, consult the mkdumprd man page and the comments in /etc/kdump.conf.
```
jlebon, line 11:
```suggestion
. Make sure the path in which the vmcore will be saved is somewhere in the `/var` directory of the FCOS stateroot. It is also possible to write the dump over the network; for additional information, see `mkdumprd(8)` and the comments in `/etc/kdump.conf`.
```

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -0,0 +1,32 @@
...
++
+[source, bash]
+----
+sed -i "s/^path.*/path \/sysroot\/ostree\/deploy\/fedora-coreos\/var\/crash/" /etc/kdump.conf
```
jlebon, line 15: Why can't we just use `/var/crash`?

jlebon (approved):
LGTM overall, just some minor things.

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -0,0 +1,32 @@
...
+----
+
+. Reboot your system.
++
+[source, bash]
+----
+sudo systemctl reboot
+----
+
+TIP: For additional information on how to test that kdump is properly armed and how to analyze the dump, refer to the https://fedoraproject.org/wiki/How_to_use_kdump_to_debug_kernel_crashes[kdump documentation for Fedora].
```
travier, line 32: Should we also link to https://www.kernel.org/doc/html/latest/admin-guide/kdump/kdump.html?

modules/ROOT/pages/debugging-kernel-crashes.adoc
kelvinfan001: hmm you're right. I just tested and it seems like the default `/var/crash` seems to have no problems. I don't remember why I thought we had to modify it.

Issue #88, opened 2020-06-24
document how to edit/set kernel arguments


dustymabe:
We have some items in the works that will make settting/editing kargs easier but for now let's just get up a page that is a starting point for people to use to configure kernel arguments persistently.

I'm thinking it should be at the same place in the navigation as https://docs.fedoraproject.org/en-US/fedora-coreos/sysctl/ but for kargs.

jdoss:
@jlebon per our conversation on IRC, this seems to work for enabling cgroups v2 on FCOS:

```
systemd:
  units:
  - name: enable-cgroups-v2.service
    enabled: true
    contents: |
      [Unit]
      Description=Enable cgroups v2 (systemd.unified_cgroup_hierarchy)
      ConditionFirstBoot=true
      Wants=basic.target
      Before=multi-user.target mycool-podman-pod.service

      [Service]
      Type=oneshot
      ExecStart=/usr/bin/rpm-ostree kargs --delete systemd.unified_cgroup_hierarchy=0 --reboot

      [Install]
      WantedBy=basic.target
```

I think this unit can be cleaned up more. An edge case I found was where this unit would be called and then before the system could reboot, my unit that sets up my podman pod, would start and it would cause podman to be configured to use cgroups v1. After the system came back up in cgroups v2 mode podman would fail to start my containers via systemd.

A work around to that was adding ` mycool-podman-pod.service` to `Before`. Maybe there is a better way to block systemd from continuing while it starts the reboot process? Having a generic drop in service for FCOS that doesn't have to be modified by an end user seems worth figuring out.

cgwalters:
I think we should extend FCOS (or potentially Ignition) to have a standard target that (if enabled) reboots and that *other* units can be ordered against.

In OpenShift we kind of hack this together by having the MCO inject systemd units that perform an OS upgrade+reboot and are `Before=kubelet.service`, but this is a generic problem.  

Something like `ignition-user-complete-reboot.target` - runs in the real root.  Users would then order units that require "real root configuration" (like `kubelet.service`) to be `After=` that generic service.  

Another way to look at this is extending the concept of Ignition as "runs at most once configuration" to the real root.  One *can* do this now, but having users invent "run at most once" semantics + reboot handling makes things more likely to clash.

bgilbert:
[...]

PR #199, opened 2020-10-23
Document how to modify kargs via `rpm-ostree`

jlebon:
Closes: #88

### DIFF
--- a/modules/ROOT/pages/kernel-args.adoc
+++ b/modules/ROOT/pages/kernel-args.adoc
@@ -1,0 +1,42 @@
+= Modifying Kernel Arguments
+
+Currently to change kernel arguments, you must script a systemd service which runs `rpm-ostree kargs --reboot`. The command supports appending kernel arguments (via `--append KEY[=VAL]`), deleting them (via `--delete KEY[=VAL]`), and replacing them (via `--replace KEY[=VALUE]`).
+
+NOTE: In the future, we will have a more Ignition-friendly method of doing this with stronger guarantees. See https://github.com/coreos/ignition/issues/1051[this upstream issue] for more information.
+
+Here's an example which drops the default `systemd.unified_cgroup_hierarchy=0` kernel argument so that the machine uses cgroups v2:
+
+[source,yaml]
+----
+variant: fcos
+version: 1.1.0
+passwd:
+  users:
+    - name: core
+      ssh_authorized_keys:
+        - $pubkey
+systemd:
+  units:
+    - name: cgroups-v2-karg.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Switch To cgroups v2
+        # We run after `systemd-machine-id-commit.service` to ensure that
+        # `ConditionFirstBoot=true` services won't rerun on the next boot.
+        After=systemd-machine-id-commit.service
+        ConditionKernelCommandLine=systemd.unified_cgroup_hierarchy
+        ConditionPathExists=!/var/lib/cgroups-v2-karg.stamp
+
+        [Service]
+        Type=oneshot
+        RemainAfterExit=yes
+        ExecStart=/bin/rpm-ostree kargs --delete=systemd.unified_cgroup_hierarchy
+        ExecStart=/bin/touch /var/lib/cgroups-v2-karg.stamp
+        ExecStart=/bin/systemctl --no-block reboot
+
+        [Install]
+        WantedBy=multi-user.target
+----
+
+NOTE:  The `After=systemd-machine-id-commit.service` directive is important to avoid some subtle issues. Similarly, any `ConditionFirstBoot=true` services should use `Before=first-boot-complete.target systemd-machine-id-commit.service`. See https://github.com/systemd/systemd/blob/3045c416e1cbbd8ab40577790522217fd1b9cb3b/man/systemd.unit.xml#L1315[the systemd documentation] for more details.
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/kernel-args.adoc
```diff
@@ -0,0 +1,72 @@
+= Modifying Kernel Arguments
+
+Currently to change kernel arguments, you must script a systemd service which runs `rpm-ostree kargs --reboot`. The command supports appending kernel arguments (via `--append KEY[=VAL]`), deleting them (via `--delete KEY[=VAL]`), and replacing them (via `--replace KEY[=VALUE]`).
+
+NOTE: In the future, we will have a more Ignition-friendly method of doing this with stronger guarantees. See https://github.com/coreos/ignition/issues/1051 for more information.
```
dustymabe, line 5: Would you want to make it a hyperlink? 

```suggestion
NOTE: In the future, we will have a more Ignition-friendly method of doing this with stronger guarantees. See the [upstream issue](https://github.com/coreos/ignition/issues/1051] for more information.
```

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -0,0 +1,72 @@
...
+        RemainAfterExit=yes
+        ExecStart=/bin/rpm-ostree kargs --reboot --delete=systemd.unified_cgroup_hierarchy
+
+        [Install]
+        WantedBy=multi-user.target
+----
+
+NOTE: Make sure to use `After=systemd-machine-id-commit.service`. Otherwise there is a chance that systemd services with `ConditionFirstBoot=true` will rerun on the next boot.
+
+The systemd service above will execute on every boot. This means that e.g. manually re-adding the kernel argument later on will cause the service to kick in again on reboot. If this is a concern, you can additionally use a stamp file:
```
dustymabe, line 40: Hmm.. Why don't we just show only one version and make it the stamp file version and tell people if they really want to run it on every boot they can remove the stampfile part?

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -0,0 +1,72 @@
...
+        [Service]
+        Type=oneshot
+        RemainAfterExit=yes
+        ExecStart=/bin/rpm-ostree kargs --reboot --delete=systemd.unified_cgroup_hierarchy
+
+        [Install]
+        WantedBy=multi-user.target
+----
+
+NOTE: Make sure to use `After=systemd-machine-id-commit.service`. Otherwise there is a chance that systemd services with `ConditionFirstBoot=true` will rerun on the next boot.
```
bgilbert, line 38: It seems better to move this to a comment in the unit file itself.

We should also document that any user-provided `ConditionFirstBoot` units must run `Before` this unit.

modules/ROOT/pages/kernel-args.adoc
rugk: Hmm, does AciiDoc support this Markdown-like syntax. Does not it have to be [like this](https://docs.fedoraproject.org/en-US/fedora-docs/asciidoc-fedora/markup/#snippets)?

dustymabe: ahh, yes I used the wrong syntax:

```suggestion
NOTE: In the future, we will have a more Ignition-friendly method of doing this with stronger guarantees. See the https://github.com/coreos/ignition/issues/1051[upstream issue] for more information.
```

modules/ROOT/pages/kernel-args.adoc
jlebon: I just dropped it since the use case for running it on every boot is not clear enough to warrant documenting.

modules/ROOT/pages/kernel-args.adoc
jlebon: Done!

modules/ROOT/pages/kernel-args.adoc
jlebon: Done!

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -0,0 +1,42 @@
...
+        RemainAfterExit=yes
+        ExecStart=/bin/rpm-ostree kargs --delete=systemd.unified_cgroup_hierarchy
+        ExecStart=/bin/touch /var/lib/cgroups-v2-karg.stamp
+        ExecStart=/bin/systemctl --no-block reboot
+
+        [Install]
+        WantedBy=multi-user.target
+----
+
+NOTE:  The `After=systemd-machine-id-commit.service` directive is important to avoid some subtle issues. Similarly, any `ConditionFirstBoot=true` services should use `Before=first-boot-complete.target systemd-machine-id-commit.service`. See https://github.com/systemd/systemd/blob/3045c416e1cbbd8ab40577790522217fd1b9cb3b/man/systemd.unit.xml#L1315[the systemd documentation] for more details.
```
jlebon, line 42: Once the new docs make it to https://www.freedesktop.org/software/systemd/man/systemd.unit.html, we can link to that instead.

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -0,0 +1,42 @@
...
+      ssh_authorized_keys:
+        - $pubkey
+systemd:
+  units:
+    - name: cgroups-v2-karg.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Switch To cgroups v2
+        # We run before to ensure that `ConditionFirstBoot=true` services won't
```
bgilbert, line 25: "We run before to ensure that" seems awkward.  Dropped word?

jlebon: Fixed!

bgilbert (approved):

dustymabe (commented):
LGTM

PR #202, opened 2020-11-07
provisioning: add instructions for OpenStack

dustymabe:


modules/ROOT/pages/provisioning-openstack.adoc
```diff
@@ -0,0 +1,97 @@
+= Provisioning Fedora CoreOS on OpenStack
+
+This guide shows how to provision new Fedora CoreOS (FCOS) nodes on an
+OpenStack cloud environment, either private, or public (like https://vexxhost.com/[VEXXHOST]).
+
+== Prerequisites
+
+Before provisioning a FCOS machine, you must have an Ignition configuration file
```
jlebon, line 8: Aside: looks like we inconsistently use "a FCOS" and "an FCOS" throughout the docs. We should decide on one and clean up the docs in a follow-up.

modules/ROOT/pages/provisioning-openstack.adoc
```diff
@@ -0,0 +1,97 @@
...
+containing your customizations. If you do not have one, see
+xref:producing-ign.adoc[Producing an Ignition File].
+
+You also need to have access to an OpenStack environment and a functioning
+https://docs.openstack.org/python-designateclient/latest/user/shell-v2.htm[`openstack` CLI].
```
jlebon, line 13:
```suggestion
https://docs.openstack.org/python-designateclient/latest/user/shell-v2.html[`openstack` CLI]. 
```

modules/ROOT/pages/provisioning-openstack.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+Before provisioning a FCOS machine, you must have an Ignition configuration file
+containing your customizations. If you do not have one, see
+xref:producing-ign.adoc[Producing an Ignition File].
+
+You also need to have access to an OpenStack environment and a functioning
+https://docs.openstack.org/python-designateclient/latest/user/shell-v2.htm[`openstack` CLI]. 
+Typically you'll https://docs.openstack.org/python-openstackclient/latest/configuration/index.html[configure the client]
+by using a `clouds.yaml` file or via environment variables. If you're starting from scratch, this
+environment may need networks, ssh keypairs, security groups, etc.. set up. Please consult the
```
jlebon, line 16:
```suggestion
environment may need networks, SSH keypairs, security groups, etc.. set up. Please consult the
```

modules/ROOT/pages/provisioning-openstack.adoc
```diff
@@ -0,0 +1,97 @@
...
+https://docs.openstack.org/[OpenStack Documentation] to learn more.
+
```
jlebon, line 18: Should we mention here what version of OpenStack the instructions below are tested against?

modules/ROOT/pages/provisioning-openstack.adoc
```diff
@@ -0,0 +1,97 @@
...
+== Downloading an OpenStack Image
+
+Fedora CoreOS is designed to be updated automatically, with different schedules per stream.
```
jlebon, line 21: Should we link to https://docs.fedoraproject.org/en-US/fedora-coreos/update-streams/ here?

modules/ROOT/pages/provisioning-openstack.adoc
```diff
@@ -0,0 +1,97 @@
...
+https://getfedora.org/coreos/download?tab=cloud_operators[download page].
+Verify the download, following the instructions on that page, and decompress it.
+
+== Uploading the Image to OpenStack
+
+.Create the FCOS image in OpenStack
+[source, bash]
+----
+FCOS_VERSION='...'
+FILE=fedora-coreos-${FCOS_VERSION}-openstack.x86_64.qcow2
```
jlebon, line 41: Minor/optional: meh... I'd just pick a sample version here and roll with that. I think it's understood that they'd sub in the filename they actually have.

modules/ROOT/pages/provisioning-openstack.adoc
```diff
@@ -0,0 +1,97 @@
...
+IMAGE=${FILE:0:-6} # pull off .qcow2
```
jlebon, line 42: I think the canonical "shell way" to strip extensions is e.g. `basename $FILE .qcow2`

modules/ROOT/pages/provisioning-openstack.adoc
```diff
@@ -0,0 +1,97 @@
...
+[source, bash]
+----
+openstack image list --name=$IMAGE
+----
+
+== Launching a VM instance
+
+Now that you have an image created in your account you can launch a VM
+instance. You’ll have to specify several pieces of information in the
+command, such as instance flavor, network information, ssh key, etc.
```
jlebon, line 56:
```suggestion
command, such as instance flavor, network information, SSH key, etc...
```

modules/ROOT/pages/provisioning-openstack.adoc
```diff
@@ -0,0 +1,97 @@
...
+----
+OPENSTACK_NETWORK=public
+openstack floating ip create $OPENSTACK_NETWORK
+
+FLOATING_IP=1.1.1.1  # from just created floating IP
+openstack server add floating ip $INSTANCE_NAME $FLOATING_IP
+----
+
+Now you should be able to SSH into the instance using the IP address
+associated with the floating IP. If you didn't change the defaults the
```
jlebon, line 96: Minor: 
```suggestion
associated with the floating IP. If you didn't change the defaults, the
```

modules/ROOT/pages/provisioning-openstack.adoc
dustymabe: I'm cool with that. Want to open an issue?

modules/ROOT/pages/provisioning-openstack.adoc
dustymabe: Will fix.. Thanks!

modules/ROOT/pages/provisioning-openstack.adoc
dustymabe: Yeah, I copy and pasted this from another guide. I think all of them could probably use a single person going through and updating each of them so the language and common steps are a bit more consistent.

modules/ROOT/pages/provisioning-openstack.adoc
dustymabe: wow. I've run `basename $FILE` in the past many times, but never used it to pull off a suffix. TIL

I still think I'd prefer to keep it the way it is, without calling out to a binary to do the same thing.

modules/ROOT/pages/provisioning-openstack.adoc
dustymabe: Added to new upload.

dustymabe:
ok I think I addressed all comments.

modules/ROOT/pages/provisioning-openstack.adoc
jlebon: Super minor: I noticed you :+1: this, but I don't see it in the latest push.

modules/ROOT/pages/provisioning-openstack.adoc
dustymabe: Added a `NOTE:` to mention it.

modules/ROOT/pages/provisioning-openstack.adoc
dustymabe: Sorry, I did indeed miss making that change. Should be done now!

modules/ROOT/pages/provisioning-openstack.adoc
dustymabe: changed to `an FCOS`

jlebon (approved):

PR #203, opened 2020-11-11
Add a bootupd section

cgwalters:
Let's describe the status quo.

### DIFF
--- a/modules/ROOT/pages/bootloader-updates.adoc
+++ b/modules/ROOT/pages/bootloader-updates.adoc
@@ -1,0 +1,74 @@
+= Updating the bootloader
+
+== bootupd
+
+Updating the bootloader is not currently automatic.  The https://github.com/coreos/bootupd/[bootupd]
+project is included in Fedora CoreOS and may be used for manual updates.
+
+This is usually only relevant on bare metal scenarios, or virtualized
+hypervisors that support Secure Boot.  An example reason to update the
+bootloader is for https://eclypsium.com/2020/07/29/theres-a-hole-in-the-boot/[the BootHole vulnerability].
+
+At the moment, only the EFI system partition (i.e. not the BIOS MBR) can be updated by bootupd.
+
+Inspect the system status:
+
+[source,bash]
+----
+# bootupctl status
+Component EFI
+  Installed: grub2-efi-x64-1:2.04-31.fc33.x86_64,shim-x64-15-8.x86_64
+  Update: At latest version
+#
+----
+
+If an update is available, use `bootupctl update` to apply it; the
+change will take effect for the next reboot.
+
+[source,bash]
+----
+# bootupctl update
+...
+Updated: grub2-efi-x64-1:2.04-31.fc33.x86_64,shim-x64-15-8.x86_64
+#
+----
+
+.Example systemd unit to automate bootupd updates
+[source,yaml]
+----
+variant: fcos
+version: 1.1.0
+systemd:
+  units:
+    - name: custom-bootupd-auto.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Bootupd automatic update
+
+        [Service]
+        ExecStart=/usr/bin/bootupctl update
+        RemainAfterExit=yes
+
+        [Install]
+        WantedBy=multi-user.target
+----
+
+=== Using images that predate bootupd
+
+Older CoreOS images that predate the existence of bootupd need
+an explicit "adoption" phase.  If `bootupctl status` says the component
+is `Adoptable`, perform the adoption with `bootupctl adopt-and-update`.
+
+[source,bash]
+----
+# bootupctl adopt-and-update
+...
+Updated: grub2-efi-x64-1:2.04-31.fc33.x86_64,shim-x64-15-8.x86_64
+#
+----
+
+=== Future versions may default to automatic updates
+
+It is possible that future Fedora CoreOS versions may default
+to automating bootloader updates similar to the above.
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -34,6 +34,7 @@
 ** OS updates
 *** xref:update-streams.adoc[Update Streams]
 *** xref:auto-updates.adoc[Auto-Updates]
+*** xref:bootloader-updates.adoc[Bootloader Updates]
 ** Troubleshooting
 *** xref:manual-rollbacks.adoc[Manual Rollbacks]
 *** xref:access-recovery.adoc[Access Recovery]


modules/ROOT/pages/bootloader-updates.adoc
```diff
@@ -0,0 +1,42 @@
...
+----
+# bootupctl update
+...
+Updated: grub2-efi-x64-1:2.04-31.fc33.x86_64,shim-x64-15-8.x86_64
+----
+
+=== Using images that predate bootupd
+
+Older CoreOS images that predate the existence of bootupd need
+an explicit "adoption" phase:
```
bgilbert, line 34: How will users know if they need to do this?

modules/ROOT/pages/bootloader-updates.adoc
```diff
@@ -0,0 +1,42 @@
+ .
```
bgilbert, line 1: Stray dot?

modules/ROOT/pages/bootloader-updates.adoc
```diff
@@ -0,0 +1,42 @@
...
+= Updating the bootloader
+
+== bootupd
+
+Updating the bootloader is not currently automatic.  The https://github.com/coreos/bootupd/[bootupd]
+project is included in Fedora CoreOS and may be used for manual updates.
+This is usually only relevant on bare metal scenarios, or virtualized
+hypervisors that support Secure Boot.
```
bgilbert, line 9: Can we expand a bit on why users might want to update, and how they'll know that that situation applies to them?

bgilbert (commented):
Thanks for the doc!

modules/ROOT/pages/bootloader-updates.adoc
cgwalters: It's visible in `bootupctl status`.  I know it could be improved, but this PR is more trying to document the current state.

cgwalters:
OK I also need to get https://github.com/coreos/bootupd/pull/102 through the sausage factory before we merge this.

modules/ROOT/pages/bootloader-updates.adoc
```diff
@@ -0,0 +1,71 @@
+= Updating the bootloader
+
+== bootupd
+
+Updating the bootloader is not currently automatic.  The https://github.com/coreos/bootupd/[bootupd]
+project is included in Fedora CoreOS and may be used for manual updates.
+
+This is usually only relevant on bare metal scenarios, or virtualized
+hypervisors that support Secure Boot.  An example reason to update the
+bootloader is for https://eclypsium.com/2020/07/29/theres-a-hole-in-the-boot/[the BootHole vulnerablity]/
```
bgilbert, line 10: `vulnerability`

Trailing slash should be trailing period.

cgwalters:
https://github.com/coreos/bootupd/pull/104

cgwalters:
> Can we rephrase this section in terms of "if you run bootupctl status/update and see this output, you'll need to run bootupctl adopt-and-update"?

Done though: I think at a high level right now it's only going to be a small subset of users that realize they might need this, and those users can probably wade through some of the upstream bootupd docs/code and or simply play with the CLI interactively.
And for hopefully most of those users they're already doing periodic reprovisioning and don't need bootupd at all.

But it did motivate me to do the the above PR which fixes that UX problem anyways.

modules/ROOT/pages/bootloader-updates.adoc
bgilbert: `vulnerability` is still misspelled.

bgilbert:
> Done though: I think at a high level right now it's only going to be a small subset of users that realize they might need this, and those users can probably wade through some of the upstream bootupd docs/code and or simply play with the CLI interactively.

Thanks for the fix (and the upstream fix).  In general, if we think a use case is mainstream enough to be worth documenting, I think it's worth giving actionable advice.

modules/ROOT/pages/bootloader-updates.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+        [Install]
+        WantedBy=multi-user.target
+----
+
+=== Using images that predate bootupd
+
+Older CoreOS images that predate the existence of bootupd need
+an explicit "adoption" phase.  You can see this by looking at
+`bootupctl status` and if it says the component is "Adoptable".
```
bgilbert, line 59:
```suggestion
an explicit "adoption" phase.  If `bootupctl status` says the component
is `Adoptable`, perform the adoption with `bootupctl adopt-and-update`.
```

modules/ROOT/pages/bootloader-updates.adoc
```diff
@@ -0,0 +1,79 @@
...
+=== Future versions may default to automatic updates
+
+It is possible that future Fedora CoreOS versions may default
+to automating bootloader updates similar to the above.
+If you choose to add a systemd unit per above (or manually
+`ssh` to a node for updates), note that concurrent
+`bootupctl update` invocations will not explicitly conflict.
+The only side effect is that without explicit ordering,
+one invocation may see and perform an update, and
+the second may detect no update available.
```
bgilbert, line 78: Is this advice actionable?  If we recommend users add a custom systemd unit, we'll need to continue supporting that unit in the future.  If it's harmless to run the custom unit alongside a hypothetical standard one, I think we can just drop everything but the first sentence of this section.

bgilbert (approved):
:+1:

modules/ROOT/pages/bootloader-updates.adoc
```diff
@@ -0,0 +1,73 @@
+= Updating the bootloader
+
+== bootupd
```
jlebon, line 3: Should we mention that this is just for EFI for now?

modules/ROOT/pages/bootloader-updates.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+[source,bash]
+----
+# bootupctl update
+...
+Updated: grub2-efi-x64-1:2.04-31.fc33.x86_64,shim-x64-15-8.x86_64
+#
+----
+
+.Example systemd unit to automate bootupd updates
```
jlebon, line 34: Should we have a `Note:` somewhere around here to say that one may not actually want to update the bootloader automatically on every host update?

cgwalters:
> In general, if we think a use case is mainstream enough to be worth documenting, I think it's worth giving actionable advice.

In the end I think this doc is on the edge for whether it should live here or in just in bootupd git as is today.  I leaned towards here just to make it more official but if we're hesitating we can just do the latter and come back here when we have a more "final" decision on bootupd's defaults.

jlebon:
Personally, I think having instructions for updating the bootloader manually would be good in the official docs. :+1:  The part I'm not sure about is the automatic updating systemd service. Feels like we should flesh out how we feel about that ourselves before providing a copy-paste example for enabling it?

modules/ROOT/pages/bootloader-updates.adoc
```diff
@@ -0,0 +1,73 @@
+= Updating the bootloader
+
+== bootupd
+
+Updating the bootloader is not currently automatic.  The https://github.com/coreos/bootupd/[bootupd]
+project is included in Fedora CoreOS and may be used for manual updates.
+
+This is usually only relevant on bare metal scenarios, or virtualized
+hypervisors that support Secure Boot.  An example reason to update the
+bootloader is for https://eclypsium.com/2020/07/29/theres-a-hole-in-the-boot/[the BootHole vulnerablity].
```
travier, line 10:
```suggestion
bootloader is for https://eclypsium.com/2020/07/29/theres-a-hole-in-the-boot/[the BootHole vulnerability].
```

travier (approved):
LGTM. I don't really see the bootupd docs growing bigger than a single page so I think keeping it in the Fedora CoreOS docs should be OK, at least until other projects adopt it.

modules/ROOT/pages/bootloader-updates.adoc
cgwalters: Done.

modules/ROOT/pages/bootloader-updates.adoc
cgwalters: Not sure what to say; either you update it automatically or you don't, right?

cgwalters:
OK since FCOS33 is rolling out (which has the updated bootupd) this one should be finally good to go!

travier (approved):

PR #211, opened 2020-12-04
Some installation updates

keszybz:


modules/ROOT/pages/getting-started-libvirt.adoc
```diff
@@ -24,14 +24,12 @@ VCPUS="2"
 RAM_MB="2048"
 DISK_GB="10"
 
-virt-install --connect="qemu:///system" --name="${VM_NAME}" --vcpus="${VCPUS}" --memory="${RAM_MB}" \
-        --os-variant="fedora32" --import --graphics="none" \
+virt-install --connect=qemu:///system --name="${VM_NAME}" --vcpus=${VCPUS} --memory=${RAM_MB} \
+        --os-variant=fedora-coreos-$STREAM --import --graphics=none \
```
dustymabe, line 28: Let's not remove the quotes at lest for entries that have variables since the others have them:

```suggestion
virt-install --connect=qemu:///system --name="${VM_NAME}" --vcpus="${VCPUS}" --memory="${RAM_MB}" \
        --os-variant="fedora-coreos-${STREAM}" --import --graphics=none \
```

modules/ROOT/pages/getting-started-libvirt.adoc
```diff
@@ -24,14 +24,12 @@ VCPUS="2"
...
         --disk="size=${DISK_GB},backing_store=${IMAGE}" \
         --qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=${IGNITION_CONFIG}"
 ----
 
 NOTE: `virt-install` requires both the OS image and Ignition file to be specified as absolute paths.
 
-NOTE: Depending on your version of `virt-install`, you may not be able to use `--os-variant=fedora32` and will get an error. In this case, you should pick an older Fedora variant (`--os-variant=fedora31` or `--os-variant=fedora30` for example). You can find the variants that are supported by you current version of `virt-install` with `osinfo-query os | grep "fedora[2-3][0-9]"`.
-
-TIP: Make sure that your user has access to `/dev/kvm`. On most modern distribution, this means adding yourself to the `kvm` group.
```
dustymabe, line 37: I think we had this tip about KVM in there because some people hit trouble (maybe on older distributions that don't have systemd 236 or newer). Instead of deleting it maybe let's update the comment to mention on newer distributions you should already have access to it?

dustymabe (commented):
Thanks @keszybz! A few comments

modules/ROOT/pages/getting-started-libvirt.adoc
keszybz: It's more complicated than that. Debian sticks to the old setting. I removed the tip because it said "modern distributions". I reworded the sentence.

dustymabe (approved):
LGTM

PR #215, opened 2020-12-07
Document recommended workflow for burning USB stick from Windows

pjvleeuwen:


modules/ROOT/pages/bare-metal.adoc
```diff
@@ -23,7 +23,8 @@ podman run --privileged --pull=always --rm -v .:/data -w /data \
 
 NOTE: You can boot the live ISO in either legacy BIOS or UEFI mode, regardless of what mode the OS will use once installed.
 
-- Burn the ISO to disk and boot it on the target system. The ISO is capable of bringing up a fully functioning FCOS system purely from memory (i.e. without using any disk storage). Once booted, you will have access to a bash command prompt.
+- Burn the ISO to disk (on Windows using rufus-3.13p.exe in DD Image mode is known to give correct results).
```
jlebon, line 26: WDYT of something like

```suggestion
- Burn the ISO to disk. On Linux and macOS, you can use `dd`. On Windows, you can use https://rufus.ie/[Rufus] in "DD Image" mode).
```

?

jlebon:
Related: https://github.com/coreos/fedora-coreos-tracker/issues/566

modules/ROOT/pages/bare-metal.adoc
pjvleeuwen: @jlebon I believe it to be an improvement, thanks!

dustymabe (approved):
LGTM

pjvleeuwen:
@jlebon @dustymabe Thanks for the reviews. Just to be sure: let me know if any further action is needed from me side.

Issue #214, opened 2020-12-06
different ways to get Live ISO give different result


pjvleeuwen:
Since I am a new FCOS user I might have a nice uninformed perspective. Via some tickets I will share some problems I ran into, in the hope that others might find them and be helped by it.

When I follow the instructions on 2020-12-05 on this page :
https://web.archive.org/web/20201206115935/https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/
then the resulting ISO is `fedora-coreos-32.20201104.3.0-live.x86_64.iso`.

Later I found that instead of the podman approach (which was quite a hassle for me) I was also able to just download the Live ISO from here :
https://web.archive.org/web/20201205115430if_/https://getfedora.org/coreos/download?tab=metal_virtualized&stream=stable
but the resulting ISO is `fedora-coreos-32.20201018.3.0-live.x86_64.iso`.

Point of this ticket is to suggest 3 small fixes:

1. make sure that the ISO from the getfedora.org page is up to date (seems to be older)
2. add a link on the docs.fedoraproject.org to the getfedora.org page
3. clarify on the docs.fedoraproject.org page whether we get stable, testing or next

bgilbert:
1. coreos-installer and the web page both get the ISO URL from the exact same metadata source.  I wonder if your browser was caching an older version of the metadata?
2. We could do that.  The idea behind recommending coreos-installer instead of the download site was to encourage use of the latest OS release, rather than "download once and reuse".  (And also to verify GPG signatures, but the container image is unverified so that doesn't work.)
3. Yes, we should do that.  coreos-installer downloads `stable` unless a different stream is specified.

pjvleeuwen:
@bgilbert Thanks for you quick response.

wrt 2 : In both cases the iso file is downloaded to file system. Since the coreos-installer approach is a lot more hassle I am more likely to keep that copy around as I do not want to repeat this process. If this is your aim, maybe it is better to provide an URL that always points to the latest stable / testing / next .

So in addition to https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/32.20201104.3.0/x86_64/fedora-coreos-32.20201104.3.0-live.x86_64.iso
also expose https://builds.coreos.fedoraproject.org/prod/streams/stable/x86_64/iso/latest/
(or similar)

[...]

PR #216, opened 2020-12-07
issue #214 - link to ISO download page and coreos-installer docs

pjvleeuwen:


modules/ROOT/pages/bare-metal.adoc
```diff
@@ -14,7 +14,7 @@ Before installing FCOS, you must have an Ignition configuration file and host it
 
 To install FCOS onto bare metal using the live ISO, follow these steps:
 
-- Download the ISO image:
+- Take the latest ISO image from the https://getfedora.org/coreos/download?tab=metal_virtualized&stream=stable[download page] or using podman (download stable by default, see https://github.com/coreos/coreos-installer/blob/master/docs/cmd/download.md[documentation] for options):
```
travier, line 17: Can you use https://coreos.github.io/coreos-installer/cmd/download/ instead? Thanks!

pjvleeuwen: done, thanks for the feedback!

travier (approved):

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -14,7 +14,7 @@ Before installing FCOS, you must have an Ignition configuration file and host it
 
 To install FCOS onto bare metal using the live ISO, follow these steps:
 
-- Download the ISO image:
+- Take the latest ISO image from the https://getfedora.org/coreos/download?tab=metal_virtualized&stream=stable[download page] or using podman (download stable by default, see https://coreos.github.io/coreos-installer/cmd/download/[documentation] for options):
```
bgilbert, line 17: How about this?

```suggestion
- Download the latest ISO image from the https://getfedora.org/coreos/download?tab=metal_virtualized&stream=stable[download page] or with podman (see https://coreos.github.io/coreos-installer/cmd/download/[documentation] for options):
```

And then let's add `-s stable` to the `podman` options.  Hopefully that'll provide a hint on how to select alternate streams, without emphasizing it too much.

pjvleeuwen: better, thanks!

bgilbert: Hmm, it looks like you opted not to add `-s stable`?

pjvleeuwen: Ah, I considered the update to the CMD documentation as separate. I am not sure how I do that.

bgilbert: Oh, I just meant in the example immediately following this line.

pjvleeuwen: Ah, now I understand what you meant.

More relevant in this context would be the (also default) `-p metal`.
So I have added them both for consistency with the URL to the download page that I used.

bgilbert (approved):
Thanks!

PR #217, opened 2020-12-08
kdump: Update link to kdump.conf(5) and mention /etc/sysconfig/kdump

travier:
Link to mankier instead of die.net which is usually outdated.


kelvinfan001 (approved):
LGTM. Yeah I guess linking to `/etc/sysconfig/kdump` makes more sense. I originally linked to `mkdumprd` because that's what they did in the Fedora documentation.

dustymabe (approved):
LGTM

PR #220, opened 2020-12-16
faq: add entry that explains SSH crypto change in F33

dustymabe:


modules/ROOT/pages/faq.adoc
```diff
@@ -336,3 +336,44 @@ systemd:
 
 For more information see
 https://github.com/coreos/fedora-coreos-tracker/issues/519[the tracker issue discussion].
+
+== After the rebase to Fedora 33 why does my tooling that uses SSH cease to work?
```
jlebon, line 340: How about something simpler like

```suggestion
== Why does SSH stop working after upgrading to Fedora 33?
```

?

modules/ROOT/pages/faq.adoc
```diff
@@ -336,3 +336,44 @@ systemd:
...
+and disabling the use of the `ssh-rsa` public key signature algorithm.
+
+You may hit issues if you use RSA keys and:
+
+* use an old version of the `SSH` client
+* use tooling/software libraries that don't support using RSA SHA2 host key signatures
+
+For example, GoLang has an https://github.com/golang/go/issues/37278[open issue]
+to solve this problem in the GoLang stack, but has yet to resolve the
+issue. This has been hit and worked around by the FCOS community in
```
jlebon, line 355:
```suggestion
For example, Go has an https://github.com/golang/go/issues/37278[open issue]
to solve this problem in its SSH implementation, but has yet to resolve
it. This has been hit and worked around by the FCOS community in
```

?

modules/ROOT/pages/faq.adoc
```diff
@@ -336,3 +336,44 @@ systemd:
...
+our build tooling and also our higher level projects:
+
+- https://github.com/coreos/fedora-coreos-tracker/issues/699[coreos/fedora-coreos-tracker#699]
+- https://github.com/coreos/coreos-assembler/issues/1772[coreos/coreos-assembler#1772]
+
+If you run into this problem and need to workaround the issue you
```
jlebon, line 361:
```suggestion
If you run into this problem and need to work around the issue, you
```

modules/ROOT/pages/faq.adoc
```diff
@@ -336,3 +336,44 @@ systemd:
...
+have a few options:
+
+- Switch to a newer non-RSA key type.
```
jlebon, line 364:
```suggestion
- Switch to a newer non-RSA SHA-1 key type.
```

(RSA SHA-2 works, right?)

jlebon (approved):
Some minor suggestions, but LGTM makes sense overall!

modules/ROOT/pages/faq.adoc
dustymabe: The RSA SHA2 keyex algo works, but using a non RSA key itself will workaround the problem entirely, which is what I'm trying to say here.

dustymabe:
Implemented most of your suggested changes! Thanks for the review.

modules/ROOT/pages/faq.adoc
```diff
@@ -336,3 +336,44 @@ systemd:
...
+In Fedora 33 there was a change to
+https://www.fedoraproject.org/wiki/Changes/StrongCryptoSettings2[implement stronger crypto defaults].
+Part of this included taking the
+https://www.openssh.com/txt/release-8.3[advice of OpenSSH] upstream
+and disabling the use of the `ssh-rsa` public key signature algorithm.
+
+You may hit issues if you use RSA keys and:
+
+* use an old version of the `SSH` client
+* use tooling/software libraries that don't support using RSA SHA2 host key signatures
```
bgilbert, line 351: "public key signatures" would be clearer, since host keys aren't involved here.

bgilbert (commented):
LGTM generally.

modules/ROOT/pages/faq.adoc
dustymabe: thanks for the suggestion. Fix in #222

PR #221, opened 2020-12-17
faq: add entry for local SELinux policy modifications

dustymabe:


modules/ROOT/pages/faq.adoc
```diff
@@ -377,3 +377,53 @@ storage:
...
+https://github.com/coreos/fedora-coreos-tracker/issues/700[coreos/fedora-coreos-tracker#700]:
+
+.Example SELinux denial
+[source, text]
+----
+systemd-resolved[755]: Failed to symlink /run/systemd/resolve/stub-resolv.conf: Permission denied
+audit[755]: AVC avc:  denied  { create } for  pid=755 comm="systemd-resolve" name=".#stub-resolv.confc418434d59d7d93a" scontext=system_u:system_r:systemd_resolved_t:s0 tcontext=system_u:object_r:systemd_resolved_var_run_t:s0 tclass=lnk_file permissive=0
+----
+
+To workaround this incompatibility, please attempt to apply policy
```
jlebon, line 402:
```suggestion
To work around this incompatibility, please attempt to apply policy
```

modules/ROOT/pages/faq.adoc
```diff
@@ -377,3 +377,53 @@ storage:
...
+modifications dynamically. i.e., for an SEBoolean you can use the
```
jlebon, line 403:
```suggestion
modifications dynamically. For example, for an SELinux boolean you can use the
```

modules/ROOT/pages/faq.adoc
```diff
@@ -377,3 +377,53 @@ storage:
...
+        [Install]
+        WantedBy=multi-user.target
+----
+
+If your system's basic functionality has stopped working because of
+SELinux denials there are two options:
+
+* Re-deploy starting with the latest image artifacts.
+** This means you start with the latest policy.
+* See see https://github.com/coreos/fedora-coreos-tracker/issues/701[coreos/fedora-coreos-tracker#701] for a workaround on how to recover the system.
```
jlebon, line 429:
```suggestion
* Follow the workaround in https://github.com/coreos/fedora-coreos-tracker/issues/701[coreos/fedora-coreos-tracker#701] to restore the base policy.
```

modules/ROOT/pages/faq.adoc
```diff
@@ -377,3 +377,53 @@ storage:
...
           PubkeyAcceptedKeyTypes=+ssh-rsa
 ----
+
+== Why do I get SELinux denials after updates if I have local policy modifications?
+
+Currently the OSTree and SELinux tooling conflict a bit. If you have
+permanently applied local policy modifications then policy updates
+delivered by the OS will no longer apply; your policy stays frozen.
+This means any policy "fixes" needed to enable new functionality will
+not get applied. See
```
jlebon, line 387: Should we show the `ostree admin config-diff | grep selinux` check here so they can easily see if they're affected? It's in the tracker issue, but it's kinda nested in there.

jlebon (approved):
Suggestions and comments, LGTM otherwise!

modules/ROOT/pages/faq.adoc
dustymabe: yeah that sounds like a good idea

dustymabe:
Addressed comments in latest upload. Thanks for the review!

PR #230, opened 2020-12-28
provisioning/vmware: update example and guestinfo details

JaimeMagiera:
Fixed the interface name ('ens192') in the afterburn section. Tested with FCOS 33.20201201.3.0 on vSphere 6.7.

### DIFF
--- a/modules/ROOT/pages/provisioning-vmware.adoc
+++ b/modules/ROOT/pages/provisioning-vmware.adoc
@@ -83,7 +83,9 @@ For this reason, on first-boot FCOS instances try to perform network autoconfigu
 If your VMware setup employs static network configuration instead, you can override this automatic DHCP setup with your own custom configuration.
 Custom networking command-line `ip=` parameter can be configured via guestinfo properties as shown below, before booting a VM for the first time.
 
-The provisioning flow follows the usual steps, plus an additional `guestinfo.afterburn.initrd.network-kargs` entry. Note that the network-kargs attribute needs to be added to the the VM's Advanced Configuration parameters ($vm.ExtensionData.Config.ExtraConfig). While the govc tool takes care of this detail when using the "-e" flag with the change command, other management tools may require more explicit reference to the Advanced Parameters attributes.  
+The provisioning flow follows the usual steps, plus an additional `guestinfo.afterburn.initrd.network-kargs` entry. 
+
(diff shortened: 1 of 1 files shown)


dustymabe:
Hey @JaimeMagiera. Thanks for the PR. I'm trying to understand why `ens9` is wrong. The interface name is dependent on the platform I think, which means that `ens9` could be the name of the interface for me, but it could be `ens192` for you. Maybe I'm missing something.

JaimeMagiera:
This is specifically for the "Booting a new VM on vSphere" section. IIRC, on vSphere the VMXNET3 defaults to "ens192" for new VMs which have not had additional network hardware set on them. (VMXNET3 is the default virtual networking interface for ESXI) So, a vSphere VM is mostly likely to have ens192 for its primary interface. Going with the most common in documentation seems a safe bet. Alternately, I could whip up some language about checking for the interface naming. My goal is to make this as easy to understand and implement as possible "OOTB"

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -83,12 +83,12 @@ For this reason, on first-boot FCOS instances try to perform network autoconfigu
 If your VMware setup employs static network configuration instead, you can override this automatic DHCP setup with your own custom configuration.
 Custom networking command-line `ip=` parameter can be configured via guestinfo properties as shown below, before booting a VM for the first time.
 
-The provisioning flow follows the usual steps, plus an additional `guestinfo.afterburn.initrd.network-kargs` entry.
+The provisioning flow follows the usual steps, plus an additional `guestinfo.afterburn.initrd.network-kargs` entry. Note that the network-kargs attribute needs to be added to the the VM's Advanced Configuration parameters ($vm.ExtensionData.Config.ExtraConfig). While the govc tool takes care of this detail when using the "-e" flag with the change command, other management tools may require more explicit reference to the Advanced Parameters attributes.
```
lucab, line 86: This would be better as a standalone `NOTE: ` paragraph, like this:
```
NOTE: if you are using a provisioning method other than `govc`, make sure that the guestinfo attribute is provisioned in the VM's Advanced Configuration Parameters (also known as `ExtraConfig`). Some management tools may default to a vApp Property instead, which does not work in this scenario.
```

lucab:
I think that the netname depends on VM specs, the current one was working in my cases but we can freely tweak it. At the end of the day it's just an example and the variable is split-out, assuming that folks will adjust it as needed.

modules/ROOT/pages/provisioning-vmware.adoc
JaimeMagiera: That reads well.

PR #231, opened 2021-01-04
storage: document root partition sizing considerations

jlebon:
Document that auto-growing is turned off if doing root
reprovisioning. And add an example of how to grow it in the case of
in-place reprovisioning.

Came out of
discussion.fedoraproject.org/t/root-luks-in-ignition/25671.

### DIFF
--- a/modules/ROOT/pages/storage.adoc
+++ b/modules/ROOT/pages/storage.adoc
@@ -175,6 +175,37 @@ storage:
 
 NOTE: You don't need the `path` or `with_mount_unit` keys; FCOS knows that the root partition is special and will figure out how to find it and mount it.
 
+== Sizing the root partition
+
+If you use Ignition to reconfigure or move the root partition, that partition is not automatically grown on first boot (see related discussions in https://github.com/coreos/fedora-coreos-tracker/issues/570[this issue]). In the case of moving the root partition to a new disk (or multiple disks), you should set the desired partition size using the `size_mib` field. If reconfiguring the root filesystem in place, as in the LUKS example above, you can resize the existing partition using the `resize` field:
+
+.Resizing the root partition to its maximum size
+[source.yaml]
+----
+variant: fcos
+version: 1.2.0
+storage:
+  disks:
+    - device: /dev/vda
+      partitions:
+        - label: root
+          number: 4
+          # 0 means to use all available space
+          size_mib: 0
+          resize: true
+  luks:
+    - name: root
+      device: /dev/disk/by-partlabel/root
+      clevis:
+        tpm2: true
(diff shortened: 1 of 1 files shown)


lucab (approved):

modules/ROOT/pages/storage.adoc
```diff
@@ -175,6 +175,36 @@ storage:
...
+[source.yaml]
+----
+variant: fcos
+version: 1.2.0
+storage:
+  disks:
+    - device: /dev/vda
+      partitions:
+        - label: root
+          number: 4
```
bgilbert, line 192: Should be redundant.

modules/ROOT/pages/storage.adoc
```diff
@@ -175,6 +175,36 @@ storage:
...
+          size_mib: 0
```
bgilbert, line 193: Add comment: `# Use all available space`

modules/ROOT/pages/storage.adoc
```diff
@@ -175,6 +175,36 @@ storage:
 
 NOTE: You don't need the `path` or `with_mount_unit` keys; FCOS knows that the root partition is special and will figure out how to find it and mount it.
 
+== Sizing the root partition
+
+Currently, as soon as you reconfigure or move the root partition, it is no longer automatically grown on first boot (see related discussions in https://github.com/coreos/fedora-coreos-tracker/issues/570[this issue]). In the case of moving the root partition to a new disk (or multiple disks), you should be setting the desired partition size(s) as part of your config (via `size_mib`). If reconfiguring the root filesystem in place (e.g. as in the LUKS example above), you can tell Ignition to grow the existing partition to take up the whole disk like this:
```
bgilbert, line 180: Do we plan to change this behavior?  We could tighten up the wording:

```suggestion
If you use Ignition to reconfigure or move the root partition, that partition is not automatically grown on first boot (see related discussions in https://github.com/coreos/fedora-coreos-tracker/issues/570[this issue]). In the case of moving the root partition to a new disk (or multiple disks), you should set the desired partition size using the `size_mib` field. If reconfiguring the root filesystem in place, as in the LUKS example above, you can resize the existing partition using the `resize` field:
```

modules/ROOT/pages/storage.adoc
jlebon: Without it, Ignition creates another partition with label `root`. (Technically, `label: root` is not necessary, but I chose to include it for clarity.)

bgilbert: Whoops, you are correct.  I misread the spec.

bgilbert (approved):

PR #236, opened 2021-01-15
storage: document boot_device sugar

### LINKED ISSUE #232: Improved RAID documentation
**Describe the enhancement**
I'm trying to set up nodes with a pair of drives so that / and /var are on separate raid partitions on the same pair of drives, but there is a serious lack of documentation around RAID setups. 

**System details**
 - Bare Metal
 - Fedora CoreOS version

**Additional information**
I've tried to either create a full-disk raid array and then partition that (which doesn't seem to work - I'm guessing partitioning has to run first?) and partitioning the disks with 2 partitions (first starts at 5000 with a size_mib of 100k, the second starts at 0 and size of 0) and then turn those into arrays but it seems to crash the system during ignition either way.

Some working examples or explicitly saying if it's not supported would be welcome.

bgilbert:
Fixes #232.

### DIFF
--- a/modules/ROOT/pages/storage.adoc
+++ b/modules/ROOT/pages/storage.adoc
@@ -222,12 +222,14 @@ Using partition numbers to refer to specific partitions is discouraged and label
 
 The x86_64 disk image is GPT formatted with a protective MBR. It supports booting via both BIOS and UEFI (including Secure Boot).
 
+The partition table layout has changed over time. The current layout is:
+
 .Partition Table for x86_64
 |============================================================================================
 | Number | Label      | Description                                          | Partition Type
-| 1      | boot       | Contains GRUB configuration, kernel/initramfs images | ext4
+| 1      | BIOS-BOOT  | Contains BIOS GRUB image                             | raw data
 | 2      | EFI-SYSTEM | Contains EFI GRUB image and Secure Boot shim         | FAT32
-| 3      | BIOS-BOOT  | Contains BIOS GRUB image                             | raw data
+| 3      | boot       | Contains GRUB configuration, kernel/initramfs images | ext4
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/storage.adoc
```diff
@@ -249,15 +251,16 @@ TARGET        SOURCE                                                   FSTYPE  O
...
 ----
 
+The EFI System Partition was formerly mounted on `/boot/efi`, but this is no longer the case.
+
 === Immutable `/`, read only `/usr`
 
 As OSTree is used to manage all files belonging to the operating system, the `/` and `/usr` mountpoints are not writable. Any changes to the operating system should be applied via https://coreos.github.io/rpm-ostree/administrator-handbook/[`rpm-ostree`].
 
-Similarly, the `/boot` and `/boot/efi` mountpoints are managed by `rpm-ostree` and changes must not be directly performed by an administrator in those directories. They are not yet mounted as read only but this is expected to change in the future.
+Similarly, the `/boot` mountpoint and EFI System Partition are managed by `rpm-ostree` and changes must not be directly performed by an administrator in those filesystems. `/boot` is not yet mounted as read only but this is expected to change in the future.
```
jlebon, line 340: Maybe "by `rpm-ostree` and `bootupd`" ? Since rpm-ostree/ostree don't actually touch the EFI partition.

jlebon, line 340: `/boot` is now mounted read-only: https://github.com/coreos/fedora-coreos-config/pull/659

Hmm, though that hasn't reached stable yet. So we can update it when it does.

jlebon (approved):
LGTM overall, just minor comments!

bgilbert:
Updated!

PR #239, opened 2021-01-30
Moved examples from old spec page to respective topic pages and updat…

JaimeMagiera:
Various changes to resolve #208

PR #241, opened 2021-02-02
storage: add swap example

bgilbert:
Requires FCCT 0.10.0.

From https://coreos.github.io/fcct/examples/.

### DIFF
--- a/modules/ROOT/pages/storage.adoc
+++ b/modules/ROOT/pages/storage.adoc
@@ -233,6 +233,31 @@ storage:
       label: root
 ----
 
+== Adding swap
+
+This example creates a swap partition spanning all of the `sdb` device, creates a swap area on it, and creates a systemd swap unit so the swap area is enabled on boot.
+
+NOTE: Using `with_mount_unit: true` with `format: swap` requires FCC spec version `1.4.0-experimental`.  After spec 1.4.0 is stabilized, version `1.4.0-experimental` will no longer be accepted by FCCT and configs that use it will need to be updated.
+
+.Configuring a swap partition on a second disk
+[source,yaml]
+----
+variant: fcos
+version: 1.4.0-experimental
+storage:
+  disks:
+    - device: /dev/sdb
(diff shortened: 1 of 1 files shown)


JaimeMagiera:
Note that #239, which is awaiting review, deletes ign-storage.adoc.

bgilbert:
Cool, I think that's consistent with what this PR is doing.

modules/ROOT/pages/storage.adoc
```diff
@@ -233,6 +233,29 @@ storage:
...
 
+== Adding swap
+
+This example creates a swap partition spanning all of the `sdb` device, creates a swap area on it, and creates a systemd swap unit so the swap area is enabled on boot.
+
+.Configuring a swap partition on a second disk
+[source,yaml]
+----
+variant: fcos
+version: 1.4.0-experimental
```
cgwalters, line 244: Hmm but should we really be documenting using the experimental fcc version?

cgwalters (commented):
Hmm

modules/ROOT/pages/storage.adoc
bgilbert: Hmm, it's a good point.  We have done it once before, in #172.  I'm okay with adding a `NOTE` or with deferring the PR outright until whenever 1.4.0 stabilizes.  Thoughts?

arithx: IMO it's situational, for something like LUKS I think getting the documentation in early (with maybe an issue being created to track the dropping of `-experimental` once the spec lands). I'd also be fine if we just want to defer until 1.4.0 stabilizes.

jlebon: No strong opinion either way myself.

bgilbert: I've added a NOTE, on the theory that it's better to document experimental FCOS functionality with caveats than not document it at all.

jlebon (approved):

PR #244, opened 2021-02-04
Fix link to rpm-ostree filesystem layout

0uep:


modules/ROOT/pages/faq.adoc
```diff
@@ -123,9 +123,10 @@ write and modify files.
 
 The directory `/etc` may be changed by deployments, but will not override user
 made changes. The content under `/var` is left untouched by rpm-ostree when
-applying upgrades or rollbacks. For more information, refer to the
-https://rpm-ostree.readthedocs.io/en/latest/manual/administrator-handbook/#filesystem-layout[filesystem
-layout].
+applying upgrades or rollbacks. For more information, refer to the 
+https://docs.fedoraproject.org/en-US/fedora-coreos/storage/#_mounted_filesystems[Filesystem configuration]
```
jlebon, line 127: WDYT about

```suggestion
https://docs.fedoraproject.org/en-US/fedora-coreos/storage/#_mounted_filesystems[Mounted Filesystems]
```

instead to match the title of the linked section?

0uep: I think *« Filesystem configuration »* is more understandable than *« Mounted Filesystems »*. In my opinion, the linked chapter is more about *« Filesystem configuration »* than *« Mounted Filesystems »*.

However you can change my edit as you wish, I do not care. :wink:

jlebon: I think the link name should match the name of the section to which it's linking. You're right that section could probably be renamed "Filesystem Configuration" instead.

jlebon: Want to push another commit to rename it in this PR? If not, I can pick that up.

0uep: Yes please pick up this PR. :+1:

jlebon: OK, tweaked this to just drop the upstream rpm-ostree link, and renamed that section. (I think that text dates back from before we had that section written up).

modules/ROOT/pages/faq.adoc
```diff
@@ -124,8 +124,8 @@ write and modify files.
 The directory `/etc` may be changed by deployments, but will not override user
 made changes. The content under `/var` is left untouched by rpm-ostree when
 applying upgrades or rollbacks. For more information, refer to the
-https://rpm-ostree.readthedocs.io/en/latest/manual/administrator-handbook/#filesystem-layout[filesystem
-layout].
+https://docs.fedoraproject.org/en-US/fedora-coreos/storage/#_mounted_filesystems[Filesystem Configuration]
```
bgilbert, line 127: This anchor won't work, since the heading is being changed below.

modules/ROOT/pages/storage.adoc
```diff
@@ -312,7 +312,7 @@ The partition table layout has changed over time. The current layout is:
 
 The EFI-SYSTEM partition can be deleted or reformatted when BIOS booting. Similarly, the BIOS-BOOT partition can be deleted or reformatted when EFI booting.
 
-== Mounted Filesystems
+== Filesystem Configuration
```
bgilbert, line 315: Hmm.  This will break any existing links to the section, which we may or may not care about.  But also, "Configuration" makes it sound like the section provides guidance on configuring filesystems, when in fact it describes the design and default layout of the filesystem hierarchy.

jlebon: How about just "Filesystems" ?

bgilbert: I can live with it, but it seems pretty generic.  I don't have a better suggestion at the moment, though.  It also seems okay to stick with the original heading for now.

jlebon: OK, in the interest of avoiding churn and breaking links, I just left the name the same.

bgilbert (approved):

PR #246, opened 2021-02-09
console: add more details on console configuration

lucab:
This expands the troubleshooting page, adding more details related
to console settings and showing how to adjust its configuration
permanently.

### DIFF
--- a/modules/ROOT/pages/emergency-shell.adoc
+++ b/modules/ROOT/pages/emergency-shell.adoc
@@ -9,12 +9,16 @@ All Fedora CoreOS (FCOS) images come with a default configuration for the consol
 
 However, it may not always match your specific hardware configuration. In that case, your can tweak console setup by adjusting kernel parameters.
 
+You can specify multiple `console=` options on the kernel command line. Kernel messages will appear on all of them, however only the last-specified device will be used as the foreground interactive console (i.e. `/dev/console`) for the machine.
+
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/emergency-shell.adoc
```diff
@@ -1,14 +1,52 @@
-= Access the emergency shell
+= Emergency console access
 
+Sometimes you may want to access the node console to perform troubleshooting steps or emergency maintenance.
+For instance, you may want to access the emergency shell on the console, in order to debug first boot provisioning issues.
```
jlebon, line 4: I feel like we should mention somewhere around here the semantics that the last console arg wins which is what motivates wanting to modify the kargs when trying to access the emergency console.

jlebon (approved):
Thanks for adding these details, very helpful!

modules/ROOT/pages/emergency-shell.adoc
lucab: Ack. I double-checked how this works with https://www.kernel.org/doc/html/v4.14/admin-guide/serial-console.html, then added a second commit with some more details.

lucab:
@jlebon tweaked with a second commit, feel free to merge if it looks better now.

jlebon (approved):
LGTM!

PR #248, opened 2021-02-11
storage: tweak note re. rootfs sizing

jlebon:
Let's move it up so that it's right at the beginning of the document.
Mention that we will now warn and eventually fail provisioning if the
size is too small.

Also switch the FCCT to use the new `resize` key instead of the more
awkward `start_mib` on the following partition.

The UX for this is awkward, and we'll work to improve this in the
future, but for now let's just make the situation more clear.

### DIFF
--- a/modules/ROOT/pages/storage.adoc
+++ b/modules/ROOT/pages/storage.adoc
@@ -4,6 +4,13 @@ Fedora CoreOS ships with a simple default storage layout: the root partition is 
 
 Below, we provide examples of various ways you can customize this.
 
+[NOTE]
+====
+*Fedora CoreOS requires the root filesystem to be at least 8 GiB.* For practical reasons, disk images for some platforms ship with a smaller root filesystem, which by default automatically expands to fill all available disk space. If you add additional partitions after the root filesystem, you must make sure to explicitly resize the root partition as shown below so that it is at least 8 GiB.
+
+Currently, if the root filesystem is smaller than 8 GiB, a warning is emitted on login. Starting from June 2021, if the root filesystem is smaller than 8 GiB and is followed by another partition, Fedora CoreOS will refuse to boot. For more details, see https://github.com/coreos/fedora-coreos-tracker/issues/586[this bug].
+====
+
 == Setting up separate /var mounts
 
 Here's an example FCC file to set up `/var` on a separate partition on the same primary disk:
@@ -22,10 +29,12 @@ storage:
     # device.
     wipe_table: false
     partitions:
+    - number: 4
+      label: root
+      # Allocate at least 8 GiB to the rootfs. See NOTE above about this.
+      size_mib: 8192
+      resize: true
     - size_mib: 0
-      # Start at 5G so that we leave enough space for the root partition.
-      # See the important NOTE below about this.
-      start_mib: 5000
       # We assign a descriptive label to the partition. This is important
       # for referring to it in a device-agnostic way in other parts of the
       # configuration.
@@ -40,8 +49,6 @@ storage:
       with_mount_unit: true
 ----
 
-NOTE: The `start_mib` field is very important. In the future, we will make more clear how much space should be reserved for the root filesystem (see https://github.com/coreos/fedora-coreos-tracker/issues/586). For now, make sure to leave at least 5G.
-
 You can of course mount only a subset of `/var` into a separate partition. For example, to mount `/var/lib/containers`:
 
 .Adding a /var/lib/containers partition to the primary disk
@@ -54,10 +61,12 @@ storage:
   - device: /dev/vda
     wipe_table: false
     partitions:
+    - number: 4
+      label: root
+      # Allocate at least 8 GiB to the rootfs. See NOTE above about this.
+      size_mib: 8192
+      resize: true
     - size_mib: 0
-      # Start at 5G so that we leave enough space for the root partition.
-      # See the important NOTE above about this.
-      start_mib: 5000
       label: containers
   filesystems:
     - path: /var/lib/containers


jlebon:
Requires: https://github.com/coreos/fedora-coreos-config/pull/850

travier (approved):

modules/ROOT/pages/storage.adoc
```diff
@@ -54,10 +55,11 @@ storage:
   - device: /dev/vda
     wipe_table: false
     partitions:
+    - number: 4
```
bgilbert, line 58: I'm a little uneasy about the tradeoff here, that we're using better syntax in exchange for a hardcoded reference to partition 4.  Should we also reference `label: root`?

modules/ROOT/pages/storage.adoc
```diff
@@ -4,6 +4,8 @@ Fedora CoreOS ships with a simple default storage layout: the root partition is
 
 Below, we provide examples of various ways you can customize this.
 
+NOTE: Fedora CoreOS requires the root filesystem to be at least 5G. If adding additional partitions after the root filesystem, you must make sure to explicitly resize the root partition as shown below. If the root filesystem is less than 5G, a warning is emitted on login. In the future, this will trigger a hard provisioning failure by default. For more details, see https://github.com/coreos/fedora-coreos-tracker/issues/586.
```
bgilbert, line 7: Hmm.  This says what the user should do but doesn't really explain why.  I tried adding more words, but perhaps I've just turned this into a wall-o-text instead.  WDYT?

```suggestion
[NOTE]
====
Fedora CoreOS requires the root filesystem to be at least 5 GiB. The default root filesystem is smaller than 5 GiB on some platforms, but is automatically resized to fill all available disk space. If you add additional partitions after the root filesystem, you must make sure to explicitly resize the root partition as shown below.

Currently, if the root filesystem is smaller than 5 GiB, a warning is emitted on login. In the future, if the root filesystem is smaller than 5 GiB, Fedora CoreOS will refuse to boot. For more details, see https://github.com/coreos/fedora-coreos-tracker/issues/586[this bug].
====
```

modules/ROOT/pages/storage.adoc
```diff
@@ -22,10 +24,11 @@ storage:
     # device.
     wipe_table: false
     partitions:
+    - number: 4
+      # Allocate at least 5G to the rootfs. See NOTE above about this.
```
bgilbert, line 28: This and the f-c-c PR are implicitly committing to a 5 GiB minimum.  Should we socialize that a bit more?  Are we confident it'll be large enough in the long term?

modules/ROOT/pages/storage.adoc
jlebon: > I'm a little uneasy about the tradeoff here, that we're using better syntax in exchange for a hardcoded reference to partition 4.

Yeah, it's not great. Would've been nice if `label: root` alone actually identified the specific partition instead of creating a separate one. (Thoughts on tweaking the next spec 3 subversion to do this? I'm guessing it'd be hard because it affects merge semantics.)

I think I do still prefer doing it this way because it's IMO much easier for users to understand what's going on. Otherwise, it all feels very indirect and we have to talk about offsets and are exposed to the fact that the sizes of the first three partitions affects the actual size of the rootfs. Could probably be convinced otherwise though. :)  I can add the `label: root` as a sanity-check.

modules/ROOT/pages/storage.adoc
jlebon: Let's keep discussing this in https://github.com/coreos/fedora-coreos-tracker/issues/586.

modules/ROOT/pages/storage.adoc
bgilbert: Offhand I'm not sure why `label` isn't defined that way.  It may be relevant that the UEFI spec apparently doesn't require the label to be unique.  (Of course, we make that assumption elsewhere, so perhaps we could do so in Ignition too.)  In any event, I think it'd require a major spec version bump to change this.

modules/ROOT/pages/storage.adoc
jlebon: Thanks, I took your text and just massaged it a bit more.

modules/ROOT/pages/storage.adoc
jlebon: Updated to use 8 GiB now as per https://github.com/coreos/fedora-coreos-tracker/issues/586#issuecomment-785254764.

jlebon:
OK cool, https://github.com/coreos/fedora-coreos-config/pull/850 is merged now, so we should get this in before the next `testing` release.

@bgilbert Any final comments?

modules/ROOT/pages/storage.adoc
```diff
@@ -4,6 +4,13 @@ Fedora CoreOS ships with a simple default storage layout: the root partition is
 
 Below, we provide examples of various ways you can customize this.
 
+[NOTE]
+====
+*Fedora CoreOS requires the root filesystem to be at least 8 GiB.* For practical purposes, disk images ship with a smaller root filesystem on some platforms, which by default automatically expands to fill all available disk space. If you add additional partitions after the root filesystem, you must make sure to explicitly resize the root partition as shown below so that it is at least 8 GiB.
```
bgilbert, line 9: That's not quite what `for practical purposes` means.  Also, `which` could refer to `platforms` on an initial parse.  How about:

```suggestion
*Fedora CoreOS requires the root filesystem to be at least 8 GiB.* For practical reasons, disk images for some platforms ship with a smaller root filesystem, which by default automatically expands to fill all available disk space. If you add additional partitions after the root filesystem, you must make sure to explicitly resize the root partition as shown below so that it is at least 8 GiB.
```

bgilbert (approved):
LGTM; one suggestion.

modules/ROOT/pages/storage.adoc
jlebon: Thanks, tweaked!

bgilbert (approved):

PR #251, opened 2021-02-20
tutorial: Update systemd user example instructions

travier:


modules/ROOT/pages/tutorial-user-systemd-unit-on-boot.adoc
```diff
@@ -47,3 +113,63 @@ storage:
...
+chcon --verbose --type svirt_home_t user.ign
+
+# Start a Fedora CoreOS virtual machine
+virt-install --name=fcos --vcpus=2 --ram=2048 --os-variant=fedora-coreos-stable \
+    --import --network=bridge=virbr0 --graphics=none \
+    --qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=${PWD}/user.ign" \
+    --disk=size=20,backing_store=${PWD}/fedora-coreos.qcow2
+----
+
+We can then verify that the unit as been started under the sleeper systemd user instannce:
```
jlebon, line 140:
```suggestion
We can then verify that the unit has been started under the sleeper systemd user instannce:
```

jlebon (approved):

modules/ROOT/pages/tutorial-user-systemd-unit-on-boot.adoc
```diff
@@ -1,18 +1,84 @@
...
 ----
 variant: fcos
-version: 1.2.0
+version: 1.3.0
 passwd:
   users:
     - name: sleeper
       ssh_authorized_keys:
-        - ssh-rsa ... 
+        - ssh-rsa ...
```
lucab, line 78: I think this can be dropped. Only the interactive admin/debugging user (`core`) would need an SSH key.

modules/ROOT/pages/tutorial-user-systemd-unit-on-boot.adoc
```diff
@@ -1,18 +1,84 @@
...
+NOTE: Make sure that you have completed the steps described in the xref:tutorial-setup.adoc[initial setup page] before starting this tutorial.
+
+In this tutorial, we will setup a user level systemd unit for an unprivileged user. There are times when it's helpful to launch a user-level https://www.freedesktop.org/software/systemd/man/systemd.unit.html[systemd unit] without having to log in. For example, you may wish to launch a container that provides a network service or run an HPC job. For this setup, we will add the following to a Fedora CoreOS config:
+
+* A user level systemd unit: `/home/sleeper/.config/systemd/user/linger-example.service`.
+* Enable it as a user level systemd service.
+
+== Setting up the systemd unit
+
+In this example, we will launch a systemd service for the user "sleeper". First, we add the systemd unit:
```
lucab, line 12: This artificial split is missing a pre-requirement. The `sleeper` user does not exist in the system and has not been introduced up to this point.

I think you'd need to introduce a small `passwd` section for that before this step. It would be good to mention that `/home/sleeper` creation happens there too.

modules/ROOT/pages/tutorial-user-systemd-unit-on-boot.adoc
```diff
@@ -47,3 +113,63 @@ storage:
...
+    --import --network=bridge=virbr0 --graphics=none \
+    --qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=${PWD}/user.ign" \
+    --disk=size=20,backing_store=${PWD}/fedora-coreos.qcow2
+----
+
+We can then verify that the unit has been started under the sleeper systemd user instannce:
+
+[source,bash]
+----
+$ ssh sleeper@192.168.x.y
```
lucab, line 144: I'd be happier if we can show the same through `ssh core@...`.

modules/ROOT/pages/tutorial-user-systemd-unit-on-boot.adoc
travier: In this example you need it to access the user and test that the services have been properly started. But I agree this is not ideal.

modules/ROOT/pages/tutorial-user-systemd-unit-on-boot.adoc
travier: OK, I'll try that

travier:
Updated with feedback.

lucab (approved):

PR #253, opened 2021-03-03
tutorials: add colon punctuation mark

rffontenelle:
Other lines in this page have colon to separate link from description


lucab (approved):
Good catch!

Issue #208, opened 2020-12-01
Delete FCC spec documentation


bgilbert:
The [official spec docs](https://coreos.github.io/fcct/specs/) are on the FCCT docs site.  Meanwhile, the FCOS docs site has a divergent, outdated, buggy, unmaintained spec writeup which also claims to be the official spec.  We've concluded internally that maintaining two separate copies of the spec is unworkable, and that the path forward is to

1. replace that portion of the FCOS docs with a link to the FCCT spec docs, but
2. continue to maintain example FCCs on the FCOS site.

JaimeMagiera:
Are you thinking cutting from "The contents of xxxx is as follows..." to the example text? That would leave something like

- node name
- sentence or two about the node
- example

Let me know I'll whip this one out.

JaimeMagiera:
Are you thinking cutting from "The contents of xxxx is as follows..." to the example. That would leave something like

- Title
- Sentence or two about the parameter
- Example

Or are you thinking delete the FCCT Specification page altogether?

travier:
> Or maybe delete the FCCT Specification page altogether, change the menu link to point to the spec page on the FCCT docs website, and just fold the examples currently on that page into the respective topic pages which are linked below in the main menu? (e.g. take the disk example from the specification page and add it to the "Configuring Storage" page)

I think the goal is to replace the entire page by a link to the upstream FCCT docs while making sure that all the existing examples are moved into the other sub pages.

PR #254, opened 2021-03-03
Adds reference to `local` storage option.

gdonval:
The token `local` has been available since https://github.com/coreos/fcct/pull/99 but the documentation did not reflect that.


bgilbert:
Thanks for the PR.  The [Provisioning and Configuration](https://docs.fedoraproject.org/en-US/fedora-coreos/fcct-config/) page is unmaintained and has various problems; see #208.  #239 will remove that page.  You can find the official FCC specification docs [here](https://coreos.github.io/fcct/specs/).

gdonval:
Ah no worries! Thanks

PR #256, opened 2021-03-08
sysconfig-network-configuration: remove line wrap

rffontenelle:
Wrapping lines of "NOTE:" blocks results in translation strings with
leading whitespace chars and trailing newline chars. It makes it harder
to translate the documentation because the translator has to deal with
adding or not the newlines, and where (if add); Also, Weblate notifies
if number of lines don't match, so it is quite annoying.

Other FCOS docs doesn't wrap any lines, so this commit unwraps all
lines in this document.

For what is worth, here is a screenshot of how the translator sees it as it is now:
wrapped-source-string

### DIFF
--- a/modules/ROOT/pages/sysconfig-network-configuration.adoc
+++ b/modules/ROOT/pages/sysconfig-network-configuration.adoc
@@ -4,31 +4,15 @@
 
 === Background
 
-By default, Fedora CoreOS (FCOS) will attempt DHCP on every interface
-with a cable plugged in. However, if you need to use static addressing or more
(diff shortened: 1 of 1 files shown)


jlebon (approved):
I'm a fan of tight line-wrapping to make things more readable when hacking, but this is 100% a valid reason to not do it.

Can you confirm that the rendering is exactly the same?

rffontenelle:
@jlebon  Looks good to me. Here is a screenshot of my local build & preview (click to enlarge): 

Screenshot of local build of Fedora CoreOS documentation

PR #260, opened 2021-03-16
More stream docs

cgwalters:
docs: Elaborate on stream metadata format a bit more in the intro

---

docs: Move update-streams.adoc to available-updates-streams.adoc

In preparation for more cleanly splitting off "what is stream metadata"
from "what are the FCOS streams".

---

Add a stream-metadata doc

Explicitly e.g. link to our Go library, etc.

---

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -3,6 +3,7 @@
 ** xref:getting-started.adoc[Getting Started]
 ** xref:bare-metal.adoc[Installing on Bare Metal]
 ** Provisioning Machines
+*** xref:stream-metadata.adoc[Stream metadata]
 *** xref:provisioning-aliyun.adoc[Booting on Alibaba Cloud]
 *** xref:provisioning-aws.adoc[Booting on AWS]
 *** xref:provisioning-azure.adoc[Booting on Azure]
--- a/modules/ROOT/pages/getting-started.adoc
+++ b/modules/ROOT/pages/getting-started.adoc
@@ -2,16 +2,16 @@
 
 == Introduction
 
-=== Update Streams
+=== Streams
 
-Metadata about Fedora CoreOS is available in a custom JSON format, called "stream metadata".  For maintaining automation, it is expected that you will interact with this stream metadata.  There are three Fedora CoreOS (FCOS) update streams available: `stable`, `testing`, and `next`. In general, you will want to use `stable`, but it is recommended to run some machines on `testing` and `next` as well and provide feedback.
+There are three Fedora CoreOS (FCOS) update streams available: `stable`, `testing`, and `next`. In general, you will want to use `stable`, but it is recommended to run some machines on `testing` and `next` as well and provide feedback.
 
-Each stream has a canonical URL representing its current state in JSON format.  For example, the URL for the stable stream is: https://builds.coreos.fedoraproject.org/streams/stable.json
+Each stream has a canonical URL representing its current state in JSON format, known as "stream metadata".  For example, the stream metadata URL for `stable` is: https://builds.coreos.fedoraproject.org/streams/stable.json
 
-While Fedora CoreOS does automatic in-place updates, it is generally a good practice to start provisioning new machines from the latest images.
-The format of this JSON file is stable, and you can parse it to e.g. find the latest ISO/AMI/etc.
+For automating Fedora CoreOS installations, it is expected that you will interact with stream metadata.  While Fedora CoreOS does automatic in-place updates, it is generally a good practice to start provisioning new machines from the latest images.
 
-For more information on streams and switching between them, see xref:update-streams.adoc[Update Streams].
+For more information on using stream metadata, see xref:stream-metadata.adoc[Stream Metadata].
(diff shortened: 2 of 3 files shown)


modules/ROOT/pages/stream-metadata.adoc
```diff
@@ -0,0 +1,39 @@
+
+= Stream metadata
```
jlebon, line 2: Let's add a link to this page from the TOC?

modules/ROOT/nav.adoc
```diff
@@ -33,7 +33,7 @@
 *** xref:kernel-args.adoc[Modifying Kernel Arguments]
 *** xref:alternatives.adoc[Setting alternatives]
 ** OS updates
-*** xref:update-streams.adoc[Update Streams]
+*** xref:available-update-streams.adoc[Update Streams]
```
jlebon, line 36: In general, we try to avoid renaming pages because it breaks links (obviously it's fine if the page is substantially changed or deleted). Probably not worth it in this case?

modules/ROOT/pages/getting-started.adoc
```diff
@@ -2,16 +2,17 @@
 
 == Introduction
 
-=== Update Streams
+=== Streams
 
-There are three Fedora CoreOS (FCOS) update streams available: `stable`, `testing`, and `next`. In general, you will want to use `stable`, but it is recommended to run some machines on `testing` and `next` as well.
+Metadata about Fedora CoreOS is available in a custom JSON format, called "stream metadata".  For maintaining automation, it is expected that you will interact with this stream metadata.  There are three Fedora CoreOS (FCOS) update streams available: `stable`, `testing`, and `next`. In general, you will want to use `stable`, but it is recommended to run some machines on `testing` and `next` as well and provide feedback.
```
jlebon, line 7: Feels funny to talk about JSON metadata right away on the Getting Started page before talking about the streams more generally.

Maybe let's invert the order of these sentences? And then at that point, we might as well make it work with the following sentence about the canonical URL. So e.g. something like

```suggestion
There are three Fedora CoreOS (FCOS) update streams available: `stable`, `testing`, and `next`. In general, you will want to use `stable`, but it is recommended to run some machines on `testing` and `next` as well and provide feedback.

Each stream has a canonical URL representing its current state in JSON format. This is called the "stream metadata".  For example, the stream metadata URL for `stable` is: https://builds.coreos.fedoraproject.org/streams/stable.json. For maintaining automation, it is expected that you will interact with stream metadata.
```

?

modules/ROOT/pages/getting-started.adoc
```diff
@@ -2,16 +2,17 @@
...
 
 Each stream has a canonical URL representing its current state in JSON format.  For example, the URL for the stable stream is: https://builds.coreos.fedoraproject.org/streams/stable.json
 
 While Fedora CoreOS does automatic in-place updates, it is generally a good practice to start provisioning new machines from the latest images.
 The format of this JSON file is stable, and you can parse it to e.g. find the latest ISO/AMI/etc.
```
jlebon, line 12: Not from your PR, though we might as well drop this sentence too now. It looked out of order even before, but now we have that info in the stream metadata page anyway.

modules/ROOT/pages/getting-started.adoc
```diff
@@ -2,16 +2,17 @@
...
 
-For more information on streams and switching between them, see xref:update-streams.adoc[Update Streams].
+For more information on using stream metadata, see xref:update-streams.adoc[Stream Metadata].
```
jlebon, line 14:
```suggestion
For more information on using stream metadata, see xref:stream-metadata.adoc[Stream Metadata].
```

jlebon (commented):
Nice, very useful to document this stuff!

cgwalters:
OK addressed comments.

modules/ROOT/pages/getting-started.adoc
```diff
@@ -2,16 +2,16 @@
...
 == Introduction
 
-=== Update Streams
+=== Streams
 
-There are three Fedora CoreOS (FCOS) update streams available: `stable`, `testing`, and `next`. In general, you will want to use `stable`, but it is recommended to run some machines on `testing` and `next` as well.
+There are three Fedora CoreOS (FCOS) update streams available: `stable`, `testing`, and `next`. In general, you will want to use `stable`, but it is recommended to run some machines on `testing` and `next` as well and provide feedback.
 
-Each stream has a canonical URL representing its current state in JSON format.  For example, the URL for the stable stream is: https://builds.coreos.fedoraproject.org/streams/stable.json
+Each stream has a canonical URL representing its current state in JSON format, known as "stream metadata".  For example, the stream metadata URL for `stable` is: https://builds.coreos.fedoraproject.org/streams/stable.json.
```
jlebon, line 9: Does the added final period here render as part of the URL? (At least [Refined GitHub](https://github.com/sindresorhus/refined-github/) gets thrown off by it here.)

jlebon (approved):
LGTM!

modules/ROOT/pages/getting-started.adoc
cgwalters: Fixed.

PR #264, opened 2021-03-27
docs/bare-metal: Discuss iso ignition embed more

cgwalters:
Came up on IRC - this is such a powerful and cool feature, we
should discuss it in docs.

### DIFF
--- a/modules/ROOT/pages/bare-metal.adoc
+++ b/modules/ROOT/pages/bare-metal.adoc
@@ -14,7 +14,7 @@ NOTE: If you have servers with different types and/or number of hard drives, you
 
 == Installing from live ISO
 
-To install FCOS onto bare metal using the live ISO, follow these steps:
+To install FCOS onto bare metal using the live ISO interactively, follow these steps:
 
 - Download the latest ISO image from the https://getfedora.org/coreos/download?tab=metal_virtualized&stream=stable[download page] or with podman (see https://coreos.github.io/coreos-installer/cmd/download/[documentation] for options):
 [source, bash]
@@ -22,6 +22,8 @@ To install FCOS onto bare metal using the live ISO, follow these steps:
 podman run --privileged --pull=always --rm -v .:/data -w /data \
     quay.io/coreos/coreos-installer:release download -s stable -p metal -f iso
 ----
+
+Note this is just using `coreos-installer` as a tool to download the ISO.
 
 NOTE: You can boot the live ISO in either legacy BIOS or UEFI mode, regardless of what mode the OS will use once installed.
 
@@ -35,6 +37,8 @@ sudo coreos-installer install /dev/sda \
 ----
 
 Once the installation is complete, you can simply `sudo reboot`. After rebooting, the first boot process begins. It is at this time that Ignition ingests the configuration file and provisions the system as specified.
+
+For more advanced ISO installs, including automation, see below.
 
 TIP: Check out `coreos-installer install --help` for more options on how to install Fedora CoreOS.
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/bare-metal.adoc
```diff
@@ -113,3 +117,71 @@ IPAPPEND 2
 == PXE rootfs image
 
 include::pxe-artifacts.adoc[]
+
+Proposed addition to https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/
+
+== Automated ISO installs with Ignition embedding
```
jlebon, line 123: This section equally applies to live PXE with `coreos-installer iso ignition embed` swapped for `coreos-installer pxe ignition wrap`.

Hmm, maybe we should restructure this page a bit. E.g. we keep this page about installing, and have a new page for live environments being useful on their own. And then here we have a section called e.g. "Installing from live ISO or PXE using Ignition" which provides the Ignition config, but otherwise links to the live page for more information about the live environments in general and the `embed`/`wrap` commands. WDYT? (We can do that bit in a follow-up!)

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -12,9 +12,9 @@ Before installing FCOS, you must have an Ignition configuration file and host it
 
 NOTE: If you have servers with different types and/or number of hard drives, you must create a separate customized Ignition config for each machine (or machine class).  A good model is to have the common parts of the configuration factored out into a separate Ignition configuration which can be merged into (via HTTP or inline) the per-machine custom config.
 
-== Installing from live ISO
+== Basic installation from live ISO
```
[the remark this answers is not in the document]
cgwalters: I kept it but added PXE as "Automated ISO/PXE installs with Ignition embedding".

modules/ROOT/pages/bare-metal.adoc
cgwalters: OK, for now I just reworked the title and also noted the existence of `coreos-instalelr pxe ignition wrap` (thanks for pointing that out!)

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -113,3 +117,73 @@ IPAPPEND 2
 == PXE rootfs image
 
 include::pxe-artifacts.adoc[]
+
+== Automated ISO/PXE installs with Ignition embedding
+
+The Fedora CoreOS Live environment is also CoreOS in the sense that it can boot via Ignition, execute containers, etc.  It includes all of the same content.
```
bgilbert, line 123: nit: We typically don't capitalize Live.

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -113,3 +117,73 @@ IPAPPEND 2
...
+
+As noted above for Live PXE, the live environment (including ISO) does not have to actually perform 
+a persistent installation.  You can boot it from a read-only medium (physical ISO, USB stick with a physical read-only switch),
```
bgilbert, line 126: Hmm, is the "read-only" and "physical read-only switch" part more confusing than it's worth?  The live environment works just as well from r/w media.

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -113,3 +117,73 @@ IPAPPEND 2
...
+and do everything you do on any other Fedora CoreOS environment.
+Each boot will re-run the Ignition config.
+
+For the ISO, the mechanism to do this is `coreos-installer ignition iso embed`, which will create a new `.iso` file that combines your configuration with the ISO.
+Similarly, there is `coreos-installer pxe ignition wrap` for the PXE case.
+
+However, many system administrators will want to perform fully unattended persistent installations instead of running stateless.
+
+Some documentation on this is on the upstream installer site: https://coreos.github.io/coreos-installer/customizing-install/
```
bgilbert, line 135: nit: Let's properly linkify this.

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -113,3 +117,73 @@ IPAPPEND 2
...
+
+To emphasize, there are *two* Ignition configurations here; the first config ("ISO Ignition") will commonly embed a second rendered configuration that runs on the "target" installation.
```
bgilbert, line 137: I usually call these the "live Ignition config" and the "installed Ignition config" ("target" also wfm).

bgilbert (commented):
I also have concerns about the overall structure of the document, but this WFM for now!  Just some nits.

travier (approved):
LGTM overall. Minor whitespace nit: We generally keep each paragraph on a single line.

cgwalters:
OK, un-wrapped the paras.

PR #265, opened 2021-03-31
Navigation: Reduce side bar nesting for commonly used sections

travier:
nav: Move 'User guides' sub sections at the root

---

nav: Merge bare metal & lives pages into Provisioning section


travier:
Before:
Screenshot_20210331_181423

After:
Screenshot_20210331_181544

travier:
I frequently navigate across those sections and find it unnecessarily redundant to have to open the User Guides one every time I look at the docs.

jlebon (approved):
This makes a lot of sense to me!

modules/ROOT/nav.adoc
```diff
@@ -1,8 +1,9 @@
 * xref:faq.adoc[FAQ]
 * xref:getting-started.adoc[Getting Started]
-* xref:bare-metal.adoc[Installing on Bare Metal]
 * Provisioning Machines
 ** xref:stream-metadata.adoc[Stream metadata]
+** xref:bare-metal.adoc[Installing on Bare Metal]
+** xref:live-booting-ipxe.adoc[Live-booting via iPXE (transient system)]
```
bgilbert, line 6: Maybe we could drop `(transient system)` to save pixels, since it's redundant with `Live-booting`.

bgilbert (approved):
:+1:

travier:
Screenshot_20210402_121601

dustymabe:
Only concern I have is: Will this change break links if people have referenced content from other places (forums, issues, etc)?

bgilbert:
This shouldn't change any page URLs AFAICT.

lucab (approved):

PR #266, opened 2021-03-31
Tutorials: Merge setup & tutorials pages for consistency

travier:
- Keep all content in sub section pages to keep the layout consistent in
  the documentation.
- Also add a note for running the tutorials on other platforms.


jlebon (approved):
Makes sense to me.

PR #268, opened 2021-04-01
Update for FCCT rename to Butane

bgilbert:
The Butane download instructions contain dangling links to upstream release artifacts, which should become valid mid next week, and to the Fedora `butane` package, which might take a week and a half.  I'm half-inclined not to hold this until then, since the PR is pretty intrusive; thoughts?

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -8,8 +8,8 @@ on:
     branches: [master]
 
 jobs:
-  fcc:
-    name: Validate FCCs
+  butane:
+    name: Validate Butane configs
     runs-on: ubuntu-latest
     steps:
       - name: Check out repository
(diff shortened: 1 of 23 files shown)


bgilbert:
This will require an update to the required checks for the `master` branch.

bgilbert:
See also an open discussion in https://github.com/coreos/fedora-coreos-tracker/issues/780 about preferred filename extensions.

travier:
Took a quick look and it looks good.

bgilbert:
The conclusion of https://github.com/coreos/fedora-coreos-tracker/issues/780 was to keep the file extensions used in this PR.

travier:
We should go ahead with this one now that the project has been renamed.

bgilbert:
Still waiting on Butane packages in F32 and F33 `updates`.  The former is arguably non-blocking but the latter probably shouldn't be.

Issue #267, opened 2021-03-31
Update cgroupsv2 docs sections


travier:
Docs changes for https://github.com/coreos/fedora-coreos-tracker/issues/292

Update the docs to mention the cgroupsv2 change and:
- how to stay on v1
- how to manually migrate a v1 system to a v2

PR #269, opened 2021-04-02
kernel-args: Update for cgroups v2 by default

travier:
Fixes https://github.com/coreos/fedora-coreos-docs/issues/267

### DIFF
--- a/modules/ROOT/pages/kernel-args.adoc
+++ b/modules/ROOT/pages/kernel-args.adoc
@@ -1,10 +1,88 @@
 = Modifying Kernel Arguments
 
-Currently to change kernel arguments, you must script a systemd service which runs `rpm-ostree kargs --reboot`. The command supports appending kernel arguments (via `--append KEY[=VAL]`), deleting them (via `--delete KEY[=VAL]`), and replacing them (via `--replace KEY[=VALUE]`).
+Kernel arguments changes are managed by `rpm-ostree` via the https://www.mankier.com/1/rpm-ostree[`rpm-ostree kargs`] sub command. Changes are applied to a new deployment and a reboot is necessary for those to take effect.
 
-NOTE: In the future, we will have a more Ignition-friendly method of doing this with stronger guarantees. See https://github.com/coreos/ignition/issues/1051[this upstream issue] for more information.
+== Adding kernel arguments
 
-Here's an example which drops the default `systemd.unified_cgroup_hierarchy=0` kernel argument so that the machine uses cgroups v2:
+You can append kernel arguments. This is useful with e.g. `console=` that can be used multiple times. An empty value for an argument is allowed:
+
+[source,bash]
+----
+$ sudo rpm-ostree kargs --append=KEY=VALUE
+----
+
+.Example: Add reserved memory for Kdump support
+
+[source,bash]
+----
+$ sudo rpm-ostree kargs --append='crashkernel=256M'
+----
+
+See also xref:debugging-kernel-crashes.adoc[Debugging kernel crashes using kdump].
+
+== Removing existing kernel arguments
+
+You can delete a specific kernel argument key/value pair or an entire argument with a single key/value pair:
+
+[source,bash]
+----
+$ sudo rpm-ostree kargs --delete=KEY=VALUE
+----
+
+.Example: Remove console parameters to enable kernel auto-detection
+
+[source,bash]
+----
+$ sudo rpm-ostree kargs --delete 'console=ttyS0,115200n8'
+----
+
+See also xref:emergency-shell.adoc[Emergency console access].
+
+.Example: Update an existing system from cgroupsv1 to cgroupsv2 and immediately reboot
+
+[source,bash]
+----
+$ sudo rpm-ostree kargs --delete=systemd.unified_cgroup_hierarchy --reboot
+----
+
+== Replacing existing kernel arguments
+
+You can replace an existing kernel argument with a new value. You can directly use `KEY=VALUE` if only one value exists for that argument. Otherwise, you can specify the new value using the following format:
+
+[source,bash]
+----
+$ sudo rpm-ostree kargs --replace=KEY=VALUE=NEWVALUE
+----
+
+.Example: Disable all CPU vulnerability mitigations
+
+[source,bash]
+----
+$ sudo rpm-ostree kargs --replace=mitigations=auto,nosmt=off
+----
+
+This switches `mitigations=auto,nosmt` to `mitigations=off` to disable all CPU vulnerability mitigations.
+
+== Interactive editing
+
+To use an editor to modify the kernel arguments:
+
+[source,bash]
+----
+$ sudo rpm-ostree kargs --editor
+----
+
+== Modifying Kernel Arguments via Ignition
+
+Currently to change kernel arguments via Ignition, you must script a systemd service which runs `rpm-ostree kargs` and then trigger a reboot.
+
+NOTE: In the future, we will have a more Ignition-friendly method of doing this with stronger guarantees. See upstream issues https://github.com/coreos/ignition/issues/1051[ignition#1051], https://github.com/coreos/fedora-coreos-tracker/issues/318[fedora-coreos-tracker#318] and https://github.com/coreos/fedora-coreos-tracker/issues/752[fedora-coreos-tracker#752] for more information.
+
+NOTE:  The `After=systemd-machine-id-commit.service` directive is important in the following examples to avoid some subtle issues. Similarly, any `ConditionFirstBoot=true` services should use `Before=first-boot-complete.target systemd-machine-id-commit.service`. See https://github.com/systemd/systemd/blob/3045c416e1cbbd8ab40577790522217fd1b9cb3b/man/systemd.unit.xml#L1315[the systemd documentation] for more details.
+
+=== Example: Moving to cgroups v2
+
+Cgroups v1 is currently the default on Fedora CoreOS on the `stable` and `testing` stream. Here's an example which removes the `systemd.unified_cgroup_hierarchy=0` kernel argument so that the machine switches cgroups v2:
 
 [source,yaml]
 ----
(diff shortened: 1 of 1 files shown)


travier:
Rebased on top of https://github.com/coreos/fedora-coreos-docs/pull/270

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -1,10 +1,49 @@
...
+$ sudo rpm-ostree kargs --editor
+----
+
+.Exemple: Update an existing system from cgroupsv1 to cgroupsv2 and immediately reboot
 
-Here's an example which drops the default `systemd.unified_cgroup_hierarchy=0` kernel argument so that the machine uses cgroups v2:
+[source,bash]
+----
+$ sudo rpm-ostree --delete=systemd.unified_cgroup_hierarchy --reboot
+----
```
dustymabe, line 34: This seems like a bad example now because we're defaulting to cgroups v2 so `systemd.unified_cgroup_hierarchy=0` won't be on the kernel command line. Is there another example we can use?

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -17,26 +56,61 @@ passwd:
         - $pubkey
 systemd:
   units:
-    - name: cgroups-v2-karg.service
+    - name: cgroups-v1-karg.service
       enabled: true
       contents: |
         [Unit]
-        Description=Switch To cgroups v2
+        Description=Switch back to cgroups v1
```
dustymabe, line 63:
```suggestion
        Description=Switch to cgroups v1
```

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -17,26 +56,61 @@ passwd:
...
         ExecStart=/bin/systemctl --no-block reboot
 
         [Install]
         WantedBy=multi-user.target
 ----
 
-NOTE:  The `After=systemd-machine-id-commit.service` directive is important to avoid some subtle issues. Similarly, any `ConditionFirstBoot=true` services should use `Before=first-boot-complete.target systemd-machine-id-commit.service`. See https://github.com/systemd/systemd/blob/3045c416e1cbbd8ab40577790522217fd1b9cb3b/man/systemd.unit.xml#L1315[the systemd documentation] for more details.
+=== Example: Disabling all CPU vulnerability mitigations
+
+Here's an example which disables all CPU vulnerability mitigations:
```
dustymabe, line 83:
```suggestion
Here's an example which switches `mitigations=auto,nosmt` to `mitigations=off` to disable all CPU vulnerability mitigations:
```

The reason for spelling it out more is because the formatting for `rpm-ostree kargs replace` is a bit awkward and not 100% intuitive what is getting replaced with what.

modules/ROOT/pages/kernel-args.adoc
travier: We can add another example but we still need a place to put this one for all our existing users that will have to manually update their systems.

travier: Added more examples!

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -1,10 +1,87 @@
 = Modifying Kernel Arguments
 
-Currently to change kernel arguments, you must script a systemd service which runs `rpm-ostree kargs --reboot`. The command supports appending kernel arguments (via `--append KEY[=VAL]`), deleting them (via `--delete KEY[=VAL]`), and replacing them (via `--replace KEY[=VALUE]`).
+Kernel arguments changes are managed by `rpm-ostree` via the https://www.mankier.com/1/rpm-ostree[`rpm-ostree kargs`] sub command. Changes are applied to a new deployment and a reboot is necessary for those to take effect.
```
jlebon, line 3: Aside: feels funny to link to a third-party rendering of our own man pages. Maybe we should figure out a way to render them in https://coreos.github.io/rpm-ostree/.

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -1,10 +1,87 @@
...
 
-NOTE: In the future, we will have a more Ignition-friendly method of doing this with stronger guarantees. See https://github.com/coreos/ignition/issues/1051[this upstream issue] for more information.
+== Adding kernel arguments
 
-Here's an example which drops the default `systemd.unified_cgroup_hierarchy=0` kernel argument so that the machine uses cgroups v2:
+You can append kernel argument. This is useful with e.g. `console=` that can be used multiple times. An empty value for an argument is allowed:
```
jlebon, line 7:
```suggestion
You can append kernel arguments. This is useful with e.g. `console=` that can be used multiple times. An empty value for an argument is allowed:
```

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -1,10 +1,87 @@
...
+.Exemple: Update an existing system from cgroupsv1 to cgroupsv2 and immediately reboot
+
+[source,bash]
+----
+$ sudo rpm-ostree kargs --delete=systemd.unified_cgroup_hierarchy --reboot
+----
+
+== Replacing existing kernel arguments
+
+You can replace an existing kernel argument with a new value. You can direclty use `KEY=VALUE` if only one value exist for that argument:
```
jlebon, line 50:
```suggestion
You can replace an existing kernel argument with a new value. You can directly use `KEY=VALUE` if only one value exists for that argument. Otherwise, you can specify the new value using the following format:
```

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -1,10 +1,87 @@
...
 
-Here's an example which drops the default `systemd.unified_cgroup_hierarchy=0` kernel argument so that the machine uses cgroups v2:
+You can append kernel argument. This is useful with e.g. `console=` that can be used multiple times. An empty value for an argument is allowed:
+
+[source,bash]
+----
+$ sudo rpm-ostree kargs --append=KEY=VALUE
+----
+
+.Exemple: Add reserved memory for Kdump support
```
jlebon, line 14:
```suggestion
.Example: Add reserved memory for Kdump support
```

(and below :) )

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -1,10 +1,87 @@
...
+
+Currently to change kernel arguments via Ignition, you must script a systemd service which runs `rpm-ostree kargs` and then trigger a reboot.
+
+NOTE: In the future, we will have a more Ignition-friendly method of doing this with stronger guarantees. See upstream issues https://github.com/coreos/ignition/issues/1051[ignition#1051], https://github.com/coreos/fedora-coreos-tracker/issues/318[fedora-coreos-tracker#318] and https://github.com/coreos/fedora-coreos-tracker/issues/752[fedora-coreos-tracker#752] for more information.
+
+NOTE:  The `After=systemd-machine-id-commit.service` directive is important in the following examples to avoid some subtle issues. Similarly, any `ConditionFirstBoot=true` services should use `Before=first-boot-complete.target systemd-machine-id-commit.service`. See https://github.com/systemd/systemd/blob/3045c416e1cbbd8ab40577790522217fd1b9cb3b/man/systemd.unit.xml#L1315[the systemd documentation] for more details.
+
+=== Example: Staying on cgroups v1
+
+Cgroups v2 is now the default on Fedora CoreOS. Here's an example which adds the `systemd.unified_cgroup_hierarchy=0` kernel argument so that the machine keeps using cgroups v1:
```
jlebon, line 84:
```suggestion
Cgroups v2 is now the default on Fedora CoreOS on the `next` stream. Here's an example which adds the `systemd.unified_cgroup_hierarchy=0` kernel argument so that the machine keeps using cgroups v1:
```

jlebon (commented):
Thanks for writing this up! I'll sanity-check that https://github.com/coreos/fedora-coreos-config/pull/910 works fine on `next-devel` and then we can merge it and then this (and then finally send out the notice!).

jlebon:
> Rebased on top of #270. This one should be merged only when we do the switch to cgroups v2 on stable.

I think we can merge it earlier so that `next` users can use the instructions there to stay on cgroups v1. I'd like us to be able to link to the docs in the email we'll send out. We just need to be clear here that that this applies to the `next` stream only for now.

modules/ROOT/pages/kernel-args.adoc
travier: Yep!

travier:
Merged everything into the current page with all examples and we can push another update once it reaches stable.

travier:
This page should also shortly be much shorter once the kargs support in Butane gets merged.

dustymabe (approved):
LGTM

modules/ROOT/pages/kernel-args.adoc
dustymabe: > We can add another example but we still need a place to put this one for all our existing users that will have to manually update their systems.

Good point. I didn't think of that.

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -39,4 +117,76 @@ systemd:
         WantedBy=multi-user.target
 ----
 
-NOTE:  The `After=systemd-machine-id-commit.service` directive is important to avoid some subtle issues. Similarly, any `ConditionFirstBoot=true` services should use `Before=first-boot-complete.target systemd-machine-id-commit.service`. See https://github.com/systemd/systemd/blob/3045c416e1cbbd8ab40577790522217fd1b9cb3b/man/systemd.unit.xml#L1315[the systemd documentation] for more details.
+=== Example: Staying on cgroups v1
+
+Cgroups v2 is now the default on Fedora CoreOS on the `next` stream. Here's an example which adds the `systemd.unified_cgroup_hierarchy=0` kernel argument so that the machine keeps using cgroups v1:
```
jlebon, line 122: Sorry, suggesting one last tweak for this:

```suggestion
Starting from April 13 2021, Cgroups v2 is the default on Fedora CoreOS on the `next` stream. Here's an example which adds the `systemd.unified_cgroup_hierarchy=0` kernel argument so that the machine keeps using cgroups v1:
```

That way, the docs don't lie until next week. :)

jlebon (approved):
Thanks a lot for writing this!

PR #274, opened 2021-04-14
update-streams: fix rebase command

dustymabe:
Now requires `--bypass-driver` to work.

### DIFF
--- a/modules/ROOT/pages/update-streams.adoc
+++ b/modules/ROOT/pages/update-streams.adoc
@@ -56,7 +56,12 @@ sudo systemctl stop zincati.service
 # Perform the rebase to a different stream
 # Available streams: "stable", "testing", and "next"
 STREAM="testing"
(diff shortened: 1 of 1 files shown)


jlebon (approved):

lucab:
Uhm, but the step right before that is stopping `zincati.service`, so there shouldn't be an update-driver active anymore at that point.
Did you skip a step, or do we have a bug around here maybe?

/cc @kelvinfan001

kelvinfan001:
ah, I think we found a bug here. `rpm-ostree` doesn't have an API to de-register as driver right now. I think this means we need to implement that and then get Zincati to de-register before it becomes inactive.

dustymabe:
> Uhm, but the step right before that is stopping `zincati.service`, so there shouldn't be an update-driver active anymore at that point.
> Did you skip a step, or do we have a bug around here maybe?

Nope. I did stop Zincati 🙂.

dustymabe:
Thanks for opening the bug @kelvinfan001

PR #276, opened 2021-04-20
troubleshooting: Update kdump instructions to reserve more memory

kelvinfan001:
On machines where `dracut-squash` is not installed, `kdump`
requires slightly more memory. Local testing shows that
`crashkernel=auto` is not enough for `kdump` on FCOS. Recommend
at least 300M of memory be reserved for the `kdump` crash kernel.

xref: https://github.com/coreos/fedora-coreos-tracker/issues/798

### DIFF
--- a/modules/ROOT/pages/debugging-kernel-crashes.adoc
+++ b/modules/ROOT/pages/debugging-kernel-crashes.adoc
@@ -1,10 +1,10 @@
 = Debugging kernel crashes using kdump
 
-. Memory must be reserved for the crash kernel during booting of the first kernel. Kernel arguments can be provided like this:
+. Memory must be reserved for the crash kernel during booting of the first kernel. `crashkernel=auto` is likely insufficient memory on Fedora CoreOS. It is recommended to start testing with 300M. Kernel arguments can be provided like this:
 +
 [source, bash]
 ----
-sudo rpm-ostree kargs --append='crashkernel=256M'
+sudo rpm-ostree kargs --append='crashkernel=300M'
 ----
 xref:kernel-args.adoc[More information] on how to modify kargs via `rpm-ostree`.
 
@@ -24,4 +24,4 @@ sudo systemctl enable kdump.service
 sudo systemctl reboot
 ----
 
-TIP: For additional information on how to test that kdump is properly armed and how to analyze the dump, refer to the https://fedoraproject.org/wiki/How_to_use_kdump_to_debug_kernel_crashes[kdump documentation for Fedora] and https://www.kernel.org/doc/html/latest/admin-guide/kdump/kdump.html[the Linux kernel documentation on kdump].
+NOTE: It is highly recommended to test the configuration after setting up the `kdump` service, with extra attention to the amount of memory reserved for the crash kernel. For information on how to test that kdump is properly armed and how to analyze the dump, refer to the https://fedoraproject.org/wiki/How_to_use_kdump_to_debug_kernel_crashes[kdump documentation for Fedora] and https://www.kernel.org/doc/html/latest/admin-guide/kdump/kdump.html[the Linux kernel documentation on kdump].


kelvinfan001:
In my local testing with 2G of RAM, the minimum amount of memory reserved for the crash kernel is around 300M.

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -1,10 +1,10 @@
 = Debugging kernel crashes using kdump
 
-. Memory must be reserved for the crash kernel during booting of the first kernel. Kernel arguments can be provided like this:
+. Memory must be reserved for the crash kernel during booting of the first kernel. `crashkernel=auto` is likely insufficient memory on Fedora CoreOS. It is recommended to start testing with 300M. Kernel arguments can be provided like this:
```
jlebon, line 3: (Aside: I wonder what heuristics `auto` uses and whether we need to tweak it so it also Just Works on FCOS.)

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -1,10 +1,10 @@
...
 +
 [source, bash]
 ----
-sudo rpm-ostree kargs --append='crashkernel=256M'
+sudo rpm-ostree kargs --append='crashkernel=512M'
```
jlebon, line 7: Should we just use 300M here to make it easy to copy/paste if that's what we recommend?

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -24,4 +24,6 @@ sudo systemctl enable kdump.service
 sudo systemctl reboot
 ----
 
-TIP: For additional information on how to test that kdump is properly armed and how to analyze the dump, refer to the https://fedoraproject.org/wiki/How_to_use_kdump_to_debug_kernel_crashes[kdump documentation for Fedora] and https://www.kernel.org/doc/html/latest/admin-guide/kdump/kdump.html[the Linux kernel documentation on kdump].
+NOTE: It is highly recommended to test the configuration after setting up the `kdump` service, with extra attention to the amount of memory reserved for the crash kernel.
+
+NOTE: For additional information on how to test that kdump is properly armed and how to analyze the dump, refer to the https://fedoraproject.org/wiki/How_to_use_kdump_to_debug_kernel_crashes[kdump documentation for Fedora] and https://www.kernel.org/doc/html/latest/admin-guide/kdump/kdump.html[the Linux kernel documentation on kdump].
```
jlebon, line 29: Should these two NOTEs be merged into one?

jlebon (approved):
Minor comments, but LGTM as is too!

modules/ROOT/pages/debugging-kernel-crashes.adoc
kelvinfan001: I think the heuristics can be found here. https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_monitoring_and_updating_the_kernel/installing-and-configuring-kdump_managing-monitoring-and-updating-the-kernel#memory-requirements-for-kdump_supported-kdump-configurations-and-targets
Not sure if kdump just assumes you're using `dracut-squash`, because FCOS seems to require way more reserved memory than it should.

jlebon: Ahh gotcha. Yeah that's likely the case.

We can consider adding dracut-squash if there's strong requests for it, though it seems like the memory requirement is just slightly higher, which seems like a good compromise for keeping those packages out of the host for a feature that most people likely won't turn on.

jlebon (approved):

jlebon:
Merging over broken CI.

PR #277, opened 2021-04-21
Update tutorial-setup doc

SaravanaStorageNetwork:
Update the version number as a variable one

### DIFF
--- a/modules/ROOT/pages/tutorial-setup.adoc
+++ b/modules/ROOT/pages/tutorial-setup.adoc
@@ -83,7 +83,7 @@ To make the tutorial simpler, you should rename the image that we have just down
 
 [source,bash]
 ----
-mv fedora-coreos-32.20200715.3.0-qemu.x86_64.qcow2 fedora-coreos.qcow2
+mv fedora-coreos-<version_number>-qemu.x86_64.qcow2 fedora-coreos.qcow2
 ----
 
 You are now ready to proceed with the xref:tutorial-autologin.adoc[first tutorial].


modules/ROOT/pages/tutorial-setup.adoc
```diff
@@ -83,7 +83,7 @@ To make the tutorial simpler, you should rename the image that we have just down
 
 [source,bash]
 ----
-mv fedora-coreos-32.20200715.3.0-qemu.x86_64.qcow2 fedora-coreos.qcow2
+mv fedora-coreos-<version_number>-qemu.x86_64.qcow2 fedora-coreos.qcow2
```
lucab, line 86: There are a few more cases like this in the rest of the page. I think it's easier to set a `$RELEASE` variable and use it later on. Something like:
```
coreos-installer download -p qemu -f qcow2.xz --decompress
RELEASE="32.20200715.3.0" # Change this according to the current version that has been downloaded.
```

lucab (commented):
Thanks for the PR. I've left a different suggestion in review.

modules/ROOT/pages/tutorial-setup.adoc
bgilbert: coreos-installer prints the full path to the downloaded file to stdout, so we could just capture that into a variable.

travier:
I think we can close this one now that we have https://github.com/coreos/fedora-coreos-docs/pull/356. Thanks for the PR!

PR #281, opened 2021-04-21
Revert "workflows: run check.py as root to avoid Podman bug"

bgilbert:
Switch back to running `check.py` as non-root.  We can land this after https://github.com/actions/virtual-environments/issues/3229 is fixed upstream and deployed (i.e., when CI for this PR is green).

This reverts commit 2bdf638d4d0db9f82f0849390347cce622d87a50.


bgilbert:
CI is now green.  We're now on OS release 20210425.1 with Podman 3.1.2.

travier (approved):

PR #284, opened 2021-04-22
Drop unlinked static-ip-config.adoc page

jlebon:
This information now lives in
https://docs.fedoraproject.org/en-US/fedora-coreos/sysconfig-network-configuration/#_configuring_a_static_ip.


dustymabe (approved):
LGTM

PR #287, opened 2021-05-05
Rename master branch to main; fix links to upstream docs

bgilbert:
We can't change the version `master` to `main` in `antora.yml` because `master` is special-cased to produce non-versioned URLs.

Hold until cutover.

### DIFF
--- a/modules/ROOT/pages/debugging-with-toolbox.adoc
+++ b/modules/ROOT/pages/debugging-with-toolbox.adoc
@@ -65,4 +65,4 @@ toolbox rm --force fedora-toolbox-32
 
 NOTE: The toolbox utility allows you to create toolboxes with your custom
 images. You can find more details in the
-https://github.com/containers/toolbox/tree/master/doc[toolbox manpages].
+https://github.com/containers/toolbox/tree/main/doc[toolbox manpages].
--- a/modules/ROOT/pages/running-containers.adoc
+++ b/modules/ROOT/pages/running-containers.adoc
@@ -69,4 +69,4 @@ systemd:
 ----
 
 === For more information
-See the https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/container.md#docker[etcd documentation] for more information on running etcd in containers and how to set up multi-node etcd.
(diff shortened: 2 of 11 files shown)


travier:
We will probably have to update the branch in https://pagure.io/fedora-docs/docs-fp-o/blob/prod/f/site.yml#_100 too

dustymabe:
> We will probably have to update the branch in https://pagure.io/fedora-docs/docs-fp-o/blob/prod/f/site.yml#_100 too

yep :) - proposed that change here: https://pagure.io/fedora-docs/docs-fp-o/pull-request/159

modules/ROOT/pages/debugging-with-toolbox.adoc
```diff
@@ -65,4 +65,4 @@ toolbox rm --force fedora-toolbox-32
 
 NOTE: The toolbox utility allows you to create toolboxes with your custom
 images. You can find more details in the
-https://github.com/containers/toolbox/tree/master/doc[toolbox manpages].
+https://github.com/containers/toolbox/tree/main/doc[toolbox manpages].
```
dustymabe, line 68: Nice. Already renamed upstream!

modules/ROOT/pages/running-containers.adoc
```diff
@@ -69,4 +69,4 @@ systemd:
 ----
 
 === For more information
-See the https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/container.md#docker[etcd documentation] for more information on running etcd in containers and how to set up multi-node etcd.
+See the https://etcd.io/docs/v3.4/op-guide/container/#docker[etcd documentation] for more information on running etcd in containers and how to set up multi-node etcd.
```
dustymabe, line 72: Is there an uversionsed version of these docs (i.e. latest). If there is, should we use that instead?

dustymabe (approved):
LGTM

modules/ROOT/pages/running-containers.adoc
bgilbert: In fact, there is.  Updated.

PR #294, opened 2021-05-18
Document package layering via systemd

rugk:
As discussed in [the forum thread here](https://discussion.fedoraproject.org/t/layering-packages-via-rpm-ostree-as-part-of-provisioning/29862?u=rugk).

Heavily inspired [by the cargs v2 example](https://docs.fedoraproject.org/en-US/fedora-coreos/kernel-args/#_example_moving_to_cgroups_v2).

Thanks a lot to the community for the help!

### DIFF
--- a/modules/ROOT/pages/layering-packages.adoc
+++ b/modules/ROOT/pages/layering-packages.adoc
@@ -1,6 +1,6 @@
 = Layering packages on the host system
 
-To make other software accessible on the host system installation you can use https://coreos.github.io/rpm-ostree/[`rpm-ostree install`] to layer rpm packages. By default, packages are downloaded from the https://src.fedoraproject.org/[Fedora repository].
+To make other software accessible on the host system installation you can use https://coreos.github.io/rpm-ostree/[`rpm-ostree install`] to layer rpm packages. By default, packages are downloaded from the https://docs.fedoraproject.org/en-US/quick-docs/repositories/[Fedora repositories].
 
 [NOTE]
 ====
--- a/modules/ROOT/pages/layering-packages.adoc
+++ b/modules/ROOT/pages/layering-packages.adoc
@@ -31,7 +31,7 @@ systemd:
       enabled: true
       contents: |
         [Unit]
-        Description=Layer nano as with rpm-ostree
+        Description=Layer nano with rpm-ostree
         # We run after `systemd-machine-id-commit.service` to ensure that
         # `ConditionFirstBoot=true` services won't rerun on the next boot.
         After=systemd-machine-id-commit.service
--- a/modules/ROOT/pages/layering-packages.adoc
+++ b/modules/ROOT/pages/layering-packages.adoc
@@ -1,0 +1,58 @@
+= Layering packages on the host system
+
+To make other software accessible on the host system installation you can use https://coreos.github.io/rpm-ostree/[`rpm-ostree install`] to layer rpm packages. By default, packages are downloaded from the https://src.fedoraproject.org/[Fedora repository].
+
+[NOTE]
+====
+It is recommend to use layering only if there is no other possible containerisation option.
+The intention of Fedora CoreOS is to keep the base image as simple and small as possible for security and maintainability reasons. That is why you should in general prefer the usage of `podman` containers over layering software.
+====
+
+To start the layering of a package, you need to write an systemd unit that executes the `rpm-ostree` command to install the wanted package(s). 
+Changes are applied to a new deployment and a reboot is necessary for those to take effect.
+
+== Example: Layering nano and setting it as a default editor
+
+This example shows on how to install the text editor `nano` and add a script to `/etc/profile.d` so that it is set as the default editor for all users.
+The parameter `--allow-inactive` is useful if the package is added to the root image in a future Fedora CoreOS release. In such a case, the parameter prevents that the service would fail.
+
+NOTE: In the future, we will have a more Ignition-friendly method of doing this with stronger guarantees. See upstream issues https://github.com/coreos/butane/issues/81[butane#81] and https://github.com/coreos/fedora-coreos-tracker/issues/681[fedora-coreos-tracker#681] for more information.
+
+NOTE:  The `After=systemd-machine-id-commit.service` directive is important in the following examples to avoid some subtle issues. Similarly, any `ConditionFirstBoot=true` services should use `Before=first-boot-complete.target systemd-machine-id-commit.service`. See https://github.com/systemd/systemd/blob/3045c416e1cbbd8ab40577790522217fd1b9cb3b/man/systemd.unit.xml#L1315[the systemd documentation] for more details.
+
+[source,yaml]
+----
+variant: fcos
+version: 1.3.0
+systemd:
+  units:
+    # installing nano as a layered package with rpm-ostree
+    - name: rpm-ostree-install-nano.service
+      enabled: true
+      contents: |
+        [Unit]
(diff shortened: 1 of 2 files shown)


travier:
Took a quick look and this looks good! Thanks!

modules/ROOT/pages/layering-packages.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+== Example: Layering nano and setting it as the default editor
+
+This example shows on how to install the text editor `nano` and add a script to `/etc/profile.d` so that it is set as the default editor for all users.
+The parameter `--allow-inactive` is useful if the package is added to the root image in a future Fedora CoreOS release. In such a case, the parameter prevents that the service would fail.
+
+NOTE: In the future, we will have a more Ignition-friendly method of doing this with stronger guarantees. See upstream issues https://github.com/coreos/butane/issues/81[butane#81] and https://github.com/coreos/fedora-coreos-tracker/issues/681[fedora-coreos-tracker#681] for more information.
+
+NOTE:  The `After=systemd-machine-id-commit.service` directive is important in the following examples to avoid some subtle issues. Similarly, any `ConditionFirstBoot=true` services should use `Before=first-boot-complete.target systemd-machine-id-commit.service`. See https://github.com/systemd/systemd/blob/3045c416e1cbbd8ab40577790522217fd1b9cb3b/man/systemd.unit.xml#L1315[the systemd documentation] for more details.
```
rugk, line 21:
```suggestion
NOTE: The `After=systemd-machine-id-commit.service` directive is important in the following examples to avoid some subtle issues. Similarly, any `ConditionFirstBoot=true` services should use `Before=first-boot-complete.target systemd-machine-id-commit.service`. See https://github.com/systemd/systemd/blob/3045c416e1cbbd8ab40577790522217fd1b9cb3b/man/systemd.unit.xml#L1315[the systemd documentation] for more details.
```

UHm I guess/hope this won’t make a difference though…

rugk:
Tested practically and it worked except [that a network error occurred](https://discussion.fedoraproject.org/t/layering-packages-via-rpm-ostree-as-part-of-provisioning/29862/10?u=rugk), but I guess this is related to a known CoreOS problem. (linked it)

modules/ROOT/pages/layering-packages.adoc
```diff
@@ -0,0 +1,58 @@
...
+  units:
+    # installing nano as a layered package with rpm-ostree
+    - name: rpm-ostree-install-nano.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Layer nano with rpm-ostree
+        # We run after `systemd-machine-id-commit.service` to ensure that
+        # `ConditionFirstBoot=true` services won't rerun on the next boot.
+        After=systemd-machine-id-commit.service
```
cgwalters, line 37: Huh, looking at this there is a `first-boot-complete.target` - in theory services like this should probably be `WantedBy=first-boot-complete.target`.  That way further things can order against it cleanly.

Also, we want `After=network-online.target` here, much the same way we do for pulling containers.

modules/ROOT/pages/layering-packages.adoc
```diff
@@ -0,0 +1,58 @@
...
+        ConditionPathExists=!/var/lib/rpm-ostree-install-nano.stamp
+
+        [Service]
+        Type=oneshot
+        RemainAfterExit=yes
+        ExecStart=/usr/bin/rpm-ostree install --allow-inactive nano
+        ExecStart=/bin/touch /var/lib/rpm-ostree-install-nano.stamp
+        ExecStart=/bin/systemctl --no-block reboot
```
cgwalters, line 45: We had a long thread on this somewhere; the problem with rebooting like this is it doesn't compose - what if e.g. two units want to reboot?  They'll race, but you only want to reboot once.

What I think would be cleanest is having something like a `coreos-firstboot-reboot.service` that is off by default and is ordered `After=first-boot-complete.target` - that way multiple units can request a reboot via `Wants=coreos-firstboot-reboot.service`.

cgwalters (commented):
Thanks for filing this!

cgwalters:
To be clear, I'm mostly OK with merging this as is if you prefer after fixing the `network-online.target` bits, and let's document the caveats.  

We can pursue better reboot handling with patches to either fedora-coreos-config or upstream systemd, then circle back and improve this.

cgwalters:
Yet another alternative though is to put this logic in https://github.com/coreos/butane

modules/ROOT/pages/layering-packages.adoc
rugk: > We had a long thread on this somewhere

Likely when you’ve discussed the cgroups example, because [I’ve literally taken it from there](https://docs.fedoraproject.org/en-US/fedora-coreos/kernel-args/#_example_moving_to_cgroups_v2) and it’s also still there – so if we’d change that, we would possibly need to adjust it there, don’t we?

> What I think would be cleanest is having something like a coreos-firstboot-reboot.service that is off by default 

OKay, so is this something and admin should configure? 
Or is this a service that should be included in Fedora CoreOS, so that an admin should only do `Wants=coreos-firstboot-reboot.service` as you say?

Again, we can hardly explain that on two totally separate pages… so I’d tend to having that included in FCOS.

modules/ROOT/pages/layering-packages.adoc
rugk: Okay, added it. That said, "if you think of “services like this” – again, the cgroups example also needs this change then.

rugk:
Oh only read your further comments now. Okay, so should I document these potential problems?
And is there anything to change at the cgroups v2 example?

> Yet another alternative though is to put this logic in https://github.com/coreos/butane

Well… concerning layering the issues are already linked (in the doc too), but given cgroups also uses this reboot switch, I guess there are a lot of other use cases to reboot during the provisioning. So having something “built-in” would indeed be nice for this.

dustymabe (approved):
LGTM

modules/ROOT/pages/layering-packages.adoc
```diff
@@ -0,0 +1,60 @@
...
+    - name: rpm-ostree-install-nano.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Layer nano with rpm-ostree
+        # We run after `systemd-machine-id-commit.service` to ensure that
+        # `ConditionFirstBoot=true` services won't rerun on the next boot.
+        After=systemd-machine-id-commit.service
+        After=network-online.target
+        WantedBy=first-boot-complete.target
```
jlebon, line 39: This is not the right place for this. It would be in the `[Install]` section. But anyway, I think since we have our own stamp file, we don't really need to hook into this at all, other than the `After=systemd-machine-id-commit.service` we already have.

modules/ROOT/pages/layering-packages.adoc
jlebon: For package layering in specific, I think the end goal for https://github.com/coreos/fedora-coreos-tracker/issues/681 is that the packages are available from first boot so that you wouldn't have to reboot at all.

jlebon (approved):
LGTM! Do you mind squashing all your commits into one? Or alternatively, we can do so on merge if you prefer.

rugk:
I don’t mind, but you can also do so when merging. I guess it makes no difference, so just do so, if you want. :wink:

jlebon:
Thanks a lot for writing this up!

I suggested a few tweaks to this in https://github.com/coreos/fedora-coreos-docs/pull/296.

PR #295, opened 2021-05-26
tutorial-services: Update for console-login-helper-messages changes

travier:
Update to match behaviour changes in console-login-helper-messages:
https://github.com/coreos/fedora-coreos-config/commit/3fbd175b16ae2d5dbb26519b89792a590ce07565
https://github.com/coreos/console-login-helper-messages/releases/tag/v0.19

### DIFF
--- a/modules/ROOT/pages/tutorial-services.adoc
+++ b/modules/ROOT/pages/tutorial-services.adoc
@@ -9,13 +9,13 @@ In this tutorial, we will run a script on the first boot via a systemd service. 
 
 == Writing the script
 
-Let's write a simple script that uses https://icanhazip.com/[icanhazip.com] to update the issuegen from https://github.com/coreos/console-login-helper-messages[console-login-helper-messages] to output the node's public IPv4 address on the serial console during bootup:
+Let's write a simple script that uses https://icanhazip.com/[icanhazip.com] to create an issue file to display as a prelogin message on the console:
 
 [source,bash]
 ----
 #!/bin/bash
 echo "Detected Public IPv4: is $(curl https://ipv4.icanhazip.com)" > \
-    /run/console-login-helper-messages/issue.d/50_public-ipv4.issue
+    /etc/issue.d/50_public-ipv4.issue
 ----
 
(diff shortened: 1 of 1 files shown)


travier:
I don't think this is fully correct but @kelvinfan001 should be able to point me in the right direction 🙂

modules/ROOT/pages/tutorial-services.adoc
```diff
@@ -15,7 +15,7 @@ Let's write a simple script that uses https://icanhazip.com/[icanhazip.com] to u
 ----
```
kelvinfan001, line 15: > Let's write a simple script that uses https://icanhazip.com/[icanhazip.com] to update the issuegen from https://github.com/coreos/console-login-helper-messages[console-login-helper-messages] to output the node's public IPv4 address on the serial console during bootup:

I think we can omit the mention of CLHM here since it's not really involved and we are no longer modifying the issuegen created by CLHM

kelvinfan001: Perhaps
> Let's write a simple script that uses icanhazip.com to update the `issue` file (prelogin message and identification file)...

kelvinfan001 (approved):
LGTM, just a minor suggestion.

travier:
Hum, looks like this does not work or does not get run at the right time (before the first login banner is displayed). How should we order it to make sure this happens in the right order?

kelvinfan001:
Ah right, so you can either do [what CLHM does](https://github.com/coreos/console-login-helper-messages/blob/3783badc8bdf2a25b05d69d22d9a49d78e2b4bc3/usr/libexec/console-login-helper-messages/gensnippet_ssh_keys#L26), which is simply to do `/usr/sbin/agetty --reload` at the end of the script.
Or, alternatively, you can do `After=getty@.service` and `After=serial-getty@.service` in the systemd unit.

travier:
My bad, I was checking with `cosa --devshell` and not `--devshell-console`. Looks good now.

kelvinfan001 (approved):
LGTM.

Issue #247, opened 2021-02-10
Document firewall configuration


bgilbert:
This is something that people will want to do.

bgilbert:
I assume we should emphasize nftables rather than legacy iptables?

magnusviri:
I just posted [how I got nftables and iptables to work](https://github.com/coreos/fedora-coreos-tracker/issues/467). I'm more than willing to fork this repository, update it, and do a pull request if what I posted is the way correct.

travier:
@magnusviri Your Butane config looks like a good start. Would you like to make a PR for a doc page with that example? Thanks!

travier:
Related to https://github.com/coreos/fedora-coreos-tracker/issues/1747, we decided that we should document:
- how to setup firewalld either via layering or via a container
- how to setup static nftables/iptables rules
- what ports and services are running / listening on the network by default in Fedora CoreOS
- why the default container networking model using network namespaces reduces the need for a firewall on Fedora CoreOS
- what can be done with cloud firewalls in some platforms

vic1707:
Is there a place to track those documentations improvements please ?

travier:
> Is there a place to track those documentations improvements please ?

It's here, this issue.

vic1707:
Was kinda hoping it got an issue/PR on another repo without being referenced 😓
So no update for now, I'll dig into the previously mentioned issues and try to figure things out, thanks 🙏

PR #296, opened 2021-05-27
Mention "OS extensions" in package layering docs

jlebon:
I like the "OS extensions" framing of looking at package layering to
make it more explicit what the intent should be. Let's add that in the
docs to help users discern when it's appropriate to use layering (though
I agree it's not always clear and some people just really want their vim
directly on the host and that's fine).

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -26,7 +26,7 @@
 ** xref:running-containers.adoc[Running Containers]
 ** xref:authentication.adoc[Configuring Users and Groups]
 ** xref:hostname.adoc[Setting a Hostname]
-** xref:layering-packages.adoc[Layering packages on the host system]
+** xref:os-extensions.adoc[Adding OS extensions]
 ** xref:customize-nic.adoc[How to Customize a NIC Name]
 ** xref:sysconfig-configure-swaponzram.adoc[Configuring SwapOnZRAM]
 ** xref:sysconfig-configure-wireguard.adoc[Configuring WireGuard]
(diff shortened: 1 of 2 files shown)


travier (approved):
LGTM

jlebon:
@rugk Would appreciate a review from you too since you wrote the original doc.

rugk (approved):
Sure LGTM.

Though the example is then of course not the best one as it installs nano. (similarly as you said users may want vim…)

Really should we containerize nano and start it with podman? Well… personally; such a simple application is just a thing I want as an OS extension as I need to open files.
Likely you also would need to install a nano container for root and all users, as you'd want to use it system-wide.

That all said, the change is of course fine with me. And thanks for pinging me. :slightly_smiling_face: 

Another use case you correctly identified of course [is a firewall](https://discussion.fedoraproject.org/t/why-is-no-firewall-firewall-cmd-installed-by-default-on-fedora-coreos/30267?u=rugk) - if someone wanted this could also be added/re-written as an example here. Though maybe not as the first one, as it is a little more complex than just layering a small package such as nano? Well does not make a different from the user's POV, but well… at least I thought a simple example with nano would be better suited.
Users can easily adjust the example then…

jlebon:
> Sure LGTM.
> 
> Though the example is then of course not the best one now as it installs nano. (similarly as you said users may want vim…)
> 
> Really should we containerize nano and start it with podman? Well… personally; such a simple application is just a thing I want as an OS extension as I need to open files.
> Likely you also would need to install a nano container for root and all users, as you'd want to use it system-wide.

Yup I think we're in agreement. Hence why I left the `nano` example. This is more just helping users make an informed decision about what they really want.

jlebon:
Re. firewall, maybe let's try to tackle https://github.com/coreos/fedora-coreos-docs/issues/247 instead and see how ergonomic we can make it. I can imagine e.g. Butane sugar for it if it helps.

PR #298, opened 2021-05-28
kernel-args: document `kernelArguments` in experimental Ignition spec

bgilbert:
Document `kernelArguments` side by side with the existing systemd service docs for now, since users may not want to use an experimental Butane/Ignition spec.

Requires a Butane release that supports the `kernel_arguments` section.

cc @arithx @dustymabe

### DIFF
--- a/modules/ROOT/pages/storage.adoc
+++ b/modules/ROOT/pages/storage.adoc
@@ -335,7 +335,7 @@ storage:
 
 This example creates a swap partition spanning all of the `sdb` device, creates a swap area on it, and creates a systemd swap unit so the swap area is enabled on boot.
 
-NOTE: Using `with_mount_unit: true` with `format: swap` requires Butane spec version `1.4.0-experimental`.  After spec 1.4.0 is stabilized, version `1.4.0-experimental` will no longer be accepted by Butane and configs that use it will need to be updated.
+NOTE: Using `with_mount_unit: true` with `format: swap` requires Butane spec version `1.4.0-experimental`.  After spec 1.4.0 is stabilized, version `1.4.0-experimental` will no longer be accepted by Butane, so Butane configs will need to be updated to replace `1.4.0-experimental` with `1.4.0`.  In addition, the corresponding Ignition config version will no longer be accepted by Ignition, so Ignition configs will need to be regenerated with a new version of Butane.
 
 .Configuring a swap partition on a second disk
 [source,yaml]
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/storage.adoc
```diff
@@ -335,7 +335,7 @@ storage:
 
 This example creates a swap partition spanning all of the `sdb` device, creates a swap area on it, and creates a systemd swap unit so the swap area is enabled on boot.
 
-NOTE: Using `with_mount_unit: true` with `format: swap` requires Butane spec version `1.4.0-experimental`.  After spec 1.4.0 is stabilized, version `1.4.0-experimental` will no longer be accepted by Butane and configs that use it will need to be updated.
+NOTE: Using `with_mount_unit: true` with `format: swap` requires Butane spec version `1.4.0-experimental`.  After spec 1.4.0 is stabilized, version `1.4.0-experimental` will no longer be accepted by Butane and the corresponding Ignition config version will no longer be accepted by Ignition.
```
dustymabe, line 338:
```suggestion
NOTE: Using `with_mount_unit: true` with `format: swap` requires Butane spec version `1.4.0-experimental`.  After spec 1.4.0 is stabilized, version `1.4.0-experimental` will no longer be accepted by Butane and the corresponding Ignition config version will no longer be accepted by Ignition and configs will need to be updated to replace `1.4.0-experimental` with `1.4.0`.
```

Optional addition to make the required future change blatantly obvious.

Of course, it is possible that it won't be that simple (i.e. the spec changes for some reason and the experimental version isn't directly promoted), but I think it's fair to gloss over that here.

dustymabe (approved):
LGTM

One optional wording suggestion.

arithx (approved):

modules/ROOT/pages/storage.adoc
bgilbert: If we're documenting some part of the experimental spec in official docs, I'd assume we're fairly confident that we'll stabilize that part as-is.

Updated to add some more words.

travier (approved):
LGTM

dustymabe:
> Requires a Butane release that supports the `kernel_arguments` section.

Is that planned?

bgilbert:
Yes, it's in progress.

dustymabe:
Just tested with latest butane.. seems to be working nicely.

bgilbert:
Just waiting for the Butane package to reach Fedora stable.

bgilbert:
Followup in #304.

PR #301, opened 2021-06-02
tutorial-services: Ensure issue file written before agetty

kelvinfan001:
We need to write the issue file into `/etc/issue.d` before
`systemd-user-sessions.service` to ensure that it is written before
`agetty` is started so that `agetty` will display our issue snippet.
We can either do this or have `agetty --reload` in the script, but
this is cleaner and it matches what console-login-helper-messages.
https://github.com/coreos/console-login-helper-messages/pull/100

### DIFF
--- a/modules/ROOT/pages/tutorial-services.adoc
+++ b/modules/ROOT/pages/tutorial-services.adoc
@@ -29,6 +29,7 @@ We need to call the script from the previous section by using a systemd unit. He
 [source,service]
 ----
 [Unit]
+Before=systemd-user-sessions.service
 After=network-online.target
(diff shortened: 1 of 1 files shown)


kelvinfan001:
@travier Sorry, when I reviewed your PR I thought the `Before=` dependency was optional since the latest CLHM didn't have it either. But then I checked harder and found out that the latest CLHM _incorrectly_ omits it (it had the dependency in previous versions).

cgwalters (approved):

cgwalters:
Thanks for fixing this issue!

travier:
Nice! Thanks!

PR #304, opened 2021-06-19
kernel-args: drop instructions for upgrading to cgroups v2

bgilbert:
All channels now default to cgroups v2.  Delete upgrade instructions and update wording of downgrade instructions.

Followup to #298.


miabbott (approved):
LGTM

travier (approved):

PR #308, opened 2021-07-08
kernel-args: add `Before=boot-complete.service`

kelvinfan001:
For units that require creating rpm-ostree transactions, there is
a risk of failing if there are other rpm-ostree transactions
taking place at the same time. Order these karg services before
`boot-complete.service` to ensure that we run early enough so that
other services (e.g. `zincati.service`) or user commands won't
cause this unit to fail.

This was noticed in https://github.com/coreos/zincati/issues/588.


modules/ROOT/pages/kernel-args.adoc
```diff
@@ -109,6 +109,10 @@ systemd:
         # We run after `systemd-machine-id-commit.service` to ensure that
         # `ConditionFirstBoot=true` services won't rerun on the next boot.
         After=systemd-machine-id-commit.service
+        # Since this unit creates an rpm-ostree transaction, we run before
+        # `boot-complete.service` to ensure that there are no other conflicting
+        # transactions that might cause this unit to fail.
+        Before=boot-complete.service
```
kelvinfan001, line 115: Note that `systemd-machine-id-commit.service` is ordered `After=` the similarly named `first-boot-complete.service`, and so this would create a cycle if we ordered `Before=first-boot-complete.service`.

But I think `boot-complete.service` is independent of `first-boot-complete.service` so we should be OK here.

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -109,6 +109,10 @@ systemd:
         # We run after `systemd-machine-id-commit.service` to ensure that
         # `ConditionFirstBoot=true` services won't rerun on the next boot.
         After=systemd-machine-id-commit.service
+        # Since this unit creates an rpm-ostree transaction, we run before
+        # `boot-complete.target` to ensure that there are no other conflicting
+        # transactions that might cause this unit to fail.
+        Before=boot-complete.target
```
kelvinfan001, line 115: Note that `systemd-machine-id-commit.service` is ordered `After=` the similarly named `first-boot-complete.target`, and so this would create a cycle if we ordered `Before=first-boot-complete.target`.

But I think `boot-complete.target` is independent of `first-boot-complete.target` so we should be OK here (tested locally and no cycles).

lucab:
This LGTM as I also suggested the initial change. @jlebon @dustymabe can you have a quick double-check?

dustymabe:
I have no objections

jlebon (commented):
We can drop the first commit now that those systemd units were removed in favour of Ignition kargs. Second commit LGTM.

kelvinfan001:
Dropped the first commit :)

jlebon (approved):

lucab:
Mmh, I did prefer the original version with `boot-complete.target`, as this isn't strictly Zincati-specific and other services could similarly benefit from not starting if we already know that we are going to reboot mid-flight. Is there a reason why this became a `Before=zincati.service`?

jlebon:
The commit is the same as before. Only the first commit which modified systemd units which no longer exist was dropped.

But still, re. that tutorial unit, I think *some* ordering wrt `boot-complete.target` can make sense, though it's not clear to me whether one is clearly better than the other. An argument for `After=` is that you don't want to do any further mutations until after the system booted successfully. Overall, seems cleaner to just leave it off from the template and let users decide if they need it and in which direction? (I guess we can add some docs about it somewhere, or just link to systemd docs.) Whereas `Before=zincati.service` is pretty clear cut.

Ideally eventually, we have layering on firstboot via a declarative mechanism, which would allow all these concerns to drop out.

kelvinfan001:
> Is there a reason why this became a Before=zincati.service?

It was because this unit has an `After=` dependency on `network-online.target`, which I remember to be ordered after `boot-complete.target`.

PR #310, opened 2021-07-24
Add Windows instructions for getting Butane

Okeanos:
Contains instructions for standalone binary installation on Windows. I presume that the user installs GPG, e.g. via Git for Windows, on their own for signature verification and, hence, don't provide additional instructions for that.

I would've added instructions for e.g. [scoop](https://scoop.sh) as well but my pull request (lukesampson/scoop-extras/pull/6194) is currently in limbo 😞.

### DIFF
--- a/modules/ROOT/pages/producing-ign.adoc
+++ b/modules/ROOT/pages/producing-ign.adoc
@@ -130,6 +130,23 @@ curl https://getfedora.org/static/fedora.gpg | gpg --import
 gpg --verify butane-x86_64-apple-darwin.asc
 ----
 
+==== Windows
+To use the Butane binary on Windows, follow these steps:
+
+. If you have not already done so, download the https://getfedora.org/static/fedora.gpg[Fedora signing keys] and import them:
++
+[source,powershell]
+----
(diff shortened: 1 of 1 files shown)


travier (approved):
LGTM. Thanks!

modules/ROOT/pages/producing-ign.adoc
```diff
@@ -130,6 +130,23 @@ curl https://getfedora.org/static/fedora.gpg | gpg --import
 gpg --verify butane-x86_64-apple-darwin.asc
 ----
 
+==== Windows
```
jlebon, line 133: Should this say e.g. `Windows (using PowerShell)`?

jlebon (approved):
Cool! One optional comment, but LGTM as is too.

modules/ROOT/pages/producing-ign.adoc
Okeanos: The Linux and macOS standalone procedures don't specify that this should be run in a shell and I assume the intended audience for Windows understands that as well. `cmd.exe` might be a contender and source of confusion, though, maybe? Hovering over the code snippet and the actual syntax should put these doubts to rest, though.

bgilbert (approved):
Thanks!

PR #312, opened 2021-07-28
Fixes various minor issues

Okeanos:
- removes unnecessary shell redirect from butane example
- Aligns prerequisite header across all provisioning manuals
- fixes stream metadata page with broken Adoc syntax and references
- Use `%N` (unit name without extension suffix) instead of writing the name multiple times in systemd os-extension unit
- removes some trailing whitespaces

### DIFF
--- a/modules/ROOT/pages/producing-ign.adoc
+++ b/modules/ROOT/pages/producing-ign.adoc
@@ -190,7 +190,7 @@ TIP: YAML files must have consistent indentation. Although Butane checks for syn
 +
 [source,bash]
 ----
-butane --pretty --strict < example.bu > example.ign
+butane --pretty --strict example.bu > example.ign
 ----
 +
 . Use the `example.ign` file to xref:getting-started.adoc[boot Fedora CoreOS].
--- a/modules/ROOT/pages/provisioning-exoscale.adoc
+++ b/modules/ROOT/pages/provisioning-exoscale.adoc
@@ -4,7 +4,9 @@ This guide shows how to provision new Fedora CoreOS (FCOS) instances on https://
 
 == Prerequisites
 
-Before provisioning a FCOS instance, you may want to have at hand an Ignition configuration file containing your customizations. If you need one, see xref:producing-ign.adoc[Producing an Ignition File].
+Before provisioning an FCOS machine, you must have an Ignition configuration file containing your customizations. If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File].
+
+If you do not want to use Ignition to get started, you can make use of the https://coreos.github.io/afterburn/platforms/[Afterburn support] and provide an SSH key via the cloud provider and continue from there.
 
 You also need to have access to an Exoscale account. https://portal.exoscale.com/register[Register] if you don't have one.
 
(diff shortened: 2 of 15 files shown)


modules/ROOT/pages/producing-ign.adoc
```diff
@@ -190,7 +190,7 @@ TIP: YAML files must have consistent indentation. Although Butane checks for syn
 +
 [source,bash]
 ----
-butane --pretty --strict < example.bu > example.ign
+butane --pretty --strict example.bu > example.ign
```
jlebon, line 193: Note this is now reading from `$PWD` being mounted into the container, whereas before this was handled by bash. It's fine in this simple example, but e.g. `butane --pretty --strict /some/other/path/example.bu > example.ign` won't necessarily work.

Overall, probably safer to just leave it in?

modules/ROOT/pages/provisioning-exoscale.adoc
```diff
@@ -4,7 +4,7 @@ This guide shows how to provision new Fedora CoreOS (FCOS) instances on https://
 
 == Prerequisites
 
-Before provisioning a FCOS instance, you may want to have at hand an Ignition configuration file containing your customizations. If you need one, see xref:producing-ign.adoc[Producing an Ignition File].
+Before provisioning an FCOS machine, you must have an Ignition configuration file containing your customizations. If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File].
```
jlebon, line 7: I think the "may" wording there refers to the fact that for Exoscale you can provision without an Ignition config presuming you provide SSH keys via the cloud. This is the case also for e.g. AWS, OpenStack, and any other clouds which Afterburn supports fetching SSH keys from. (And actually the AWS quickstart leverages this: https://docs.fedoraproject.org/en-US/fedora-coreos/getting-started/#_booting_on_a_cloud_vm_aws_example).

Though... we don't currently have that wording either for the dedicated OpenStack and AWS pages, so agreed let's just harmonize them for now.

jlebon (approved):
Nice, thanks a lot for these fixes! One minor comment, but LGTM overall.

modules/ROOT/pages/producing-ign.adoc
Okeanos: Huh … now that's a tricky one. I know that running butane from a container is supported and works – personally I just never do it and instead use a local butane binary 🤷‍♂️. I am not even sure outside of the explicit container example on the same page the docs actually refer to this use case a lot/at all.

modules/ROOT/pages/provisioning-exoscale.adoc
Okeanos: Yeah, there were a couple of these extra things sprinkled in – was slightly confused because of the "may" but opted for the harmonisation approach anyway. The examples for AWS and OpenStack are interesting. I can update this pull request and incorporate your feedback and make this better visible. At the same time … why would one install FCOS without a butane config? As a barebones hello-world example?

Okeanos: @jlebon please check; I just updated the relevant cloud provider documentations with a link to the Afterburn documentation.
I also restructured the IBM prerequisite section slightly to be more in line with the others.

modules/ROOT/pages/producing-ign.adoc
jlebon: OK the example in that section already talks about using stdin vs passing as a file, which is nice. I'm cool with this change then.

modules/ROOT/pages/provisioning-exoscale.adoc
jlebon: Yup, thanks a lot!

> At the same time … why would one install FCOS without a butane config? As a barebones hello-world example?

It's a simple way to kick the tires on the OS itself before getting involved with Butane/Ignition.

But also, although we strongly recommend using Ignition for provisioning, depending on what you're doing there's nothing inherently wrong with just bringing up the machine and using SSH for configuring.

jlebon (approved):
LGTM! Will let someone else have a look as well.

dustymabe:
LGTM

dustymabe (approved):

Issue #305, opened 2021-06-29
List of unofficial tutorials


travier:
This is an issue to track unofficial tutorials, that have not been vetted for correctness or may not be up to date but that may contain useful content with which one could start its journey with Fedora CoreOS if the topic is not yet covered in the docs.

**Please double check that the topic you're looking for is not covered in the official documentation before using one of those tutorials**

- [How-to: Install Fedora CoreOS at Hetzner Cloud](https://community.hetzner.com/tutorials/howto-hcloud-terraform-fedora-coreos)
- [Installing Fedora CoreOS on a VPS in OVHcloud (OVH)]( https://tim.siosm.fr/blog/2021/01/03/fedora-coreos-ovhcloud/)

travier:
- [CoreOS Home Server Setup](https://github.com/deuill/coreos-home-server)
- [Butane config to host a Matrix homeserver on Fedora CoreOS](https://github.com/travier/fedora-coreos-matrix)
- [Butane configs to deploy Nomad on Fedora CoreOS](https://github.com/travier/fedora-coreos-nomad)

travier:
Added your blog post here @fale. Thanks!

PR #318, opened 2021-08-08
Adds VirtualBox provisioning information

Okeanos:
This adds bare metal inspired instructions for installing CoreOS on VirtualBox.

The basic idea came from [here](https://jjasghar.github.io/blog/2020/05/26/fedora-coreos-working-on-virtualbox/) – I opted for a full rewrite instead of adding it to #305 because the original did not cover `vboxmanage` tooling or the serial port debugging capabilities. There were also a few minor nitpicks (such as curling the ign config instead of providing it directly) that I fixed or documented.

### DIFF
--- a/modules/ROOT/pages/provisioning-virtualbox.adoc
+++ b/modules/ROOT/pages/provisioning-virtualbox.adoc
@@ -1,0 +1,195 @@
+= Provisioning Fedora CoreOS on VirtualBox
+
+This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualBox hypervisor.
+
+== Prerequisites
+
+Before provisioning an FCOS machine, you must have an Ignition configuration file containing your customizations and host it somewhere (e.g. using `python3 -m http.server`, we will assume this option is used in the following examples). If you do not already have an Ignition file, see xref:producing-ign.adoc[Producing an Ignition File].
+
+== Setting up a new VM
+
+You can set up a VirtualBox virtual machine through the GUI or via the https://www.virtualbox.org/manual/UserManual.html#vboxmanage[`vboxmanage` CLI]. Below is a shell script that does most of what the graphical wizard will do. Note that it is tailored to the installation from live ISO or raw disk image as outlined in the following sections.
+
+NOTE: Please be aware that for now there is no dedicated Fedora CoreOS image available for VirtualBox and the installation instructions provided are essentially xref:bare-metal.adoc[bare metal instructions]. A dedicated image may be provided at a later time. Please check the GitHub issues https://github.com/coreos/fedora-coreos-tracker/issues/73[#73] and https://github.com/coreos/fedora-coreos-tracker/issues/144[#144] for more details.
+
+We will rely on the VirtualBox-provided NAT gateway to access the hosts' loopback interface on `10.0.2.2`. Check the https://www.virtualbox.org/manual/UserManual.html#network_nat[VirtualBox User Manual] for more details.
+
+[source, bash]
+----
+VM_NAME='fcos-node01'
+VM_DISK=$(realpath ~/VirtualBox\ VMs/${VM_NAME})/${VM_NAME}.vdi
+
+vboxmanage createvm --name "${VM_NAME}" --ostype Fedora_64 --register
+# Set hardware settings for the VM, including recommended graphics controller
+vboxmanage modifyvm "${VM_NAME}" --memory 4096 --cpus 2 --vram 20 --graphicscontroller vmsvga --rtcuseutc on
+# Add a storage controller for hard disks
+vboxmanage storagectl "${VM_NAME}" --name SATA --add sata --controller IntelAhci --portcount 30 --bootable on
+----
+
+WARNING: VirtualBox and by extension `vboxmanage` always expect absolute file paths when specifying locations.
+
+Make sure to create a port forwarding rule that allows access to the guest's SSH Port with the https://www.virtualbox.org/manual/UserManual.html#networkingmodes[default NAT setting] using e.g. `localhost:2222` as target address:
+[source, bash]
+----
+vboxmanage modifyvm "${VM_NAME}" --natpf1 "guestssh,tcp,,2222,,22"
+----
+
+The virtual machine is now ready and FCOS can be installed.
+
+== Installing from live ISO
+
+To install FCOS into a VirtualBox virtual machine using the live ISO, follow these steps:
+
+- Download the latest ISO image from the https://getfedora.org/coreos/download?tab=metal_virtualized&stream=stable[download page] or with podman (see https://coreos.github.io/coreos-installer/cmd/download/[documentation] for options):
+[source, bash]
+----
+podman run --security-opt label=disable --pull=always --rm -v .:/data -w /data \
+    quay.io/coreos/coreos-installer:release download -f iso
+----
+
+- Create a sufficiently large disk and attach it to the previously created VM:
+[source, bash]
+----
+vboxmanage createmedium --filename "${VM_DISK}" --size 10240 --format VDI
+vboxmanage storageattach "${VM_NAME}" --storagectl SATA --type hdd --port 0 --device 0 --medium "${VM_DISK}"
+----
+
+[NOTE]
+====
+*Fedora CoreOS requires the root filesystem to be at least 8 GiB.* For practical reasons, disk images for some platforms ship with a smaller root filesystem, which by default automatically expands to fill all available disk space. If you add additional partitions after the root filesystem, you must make sure to explicitly resize the root partition as shown so that it is at least 8 GiB.
+
+Currently, if the root filesystem is smaller than 8 GiB, a warning is emitted on login. Starting from June 2021, if the root filesystem is smaller than 8 GiB and is followed by another partition, Fedora CoreOS will refuse to boot. For more details, see https://github.com/coreos/fedora-coreos-tracker/issues/586[this bug].
+====
+
+- Attach a DVD drive and the live ISO to the VM:
+[source, bash]
+----
+FCOS_DISK=$(realpath ~/Downloads/fedora-coreos-34.20210711.3.0-live.x86_64.iso)
+
+vboxmanage storagectl "${VM_NAME}" --name IDE --add ide --controller PIIX4 --hostiocache on --portcount 2 --bootable on
+vboxmanage storageattach "${VM_NAME}" --storagectl IDE --type dvddrive --port 1 --device 0 --medium "${FCOS_DISK}"
+----
+
+You can now boot the target system you have configured. The ISO is capable of bringing up a fully functioning FCOS system purely from memory (i.e. without using any disk storage). Once booted, you will have access to a bash command prompt.
+
+You can now run `coreos-installer` inside the VM:
+[source, bash]
+----
+sudo coreos-installer install /dev/sda \
+    --insecure-ignition --ignition-url http://10.0.2.2:8000/example.ign
+----
+
+TIP: Check out `coreos-installer install --help` for more options on how to install Fedora CoreOS.
+
+Once the installation is complete, you have to remove the live ISO after shutting down (`sudo shutdown -h now`) the VM. Once the ISO is removed, you can start the VM anew. The actual first boot process begins now. It is at this time that Ignition ingests the configuration file and provisions the system as specified.
+
+[source, bash]
+----
+vboxmanage storageattach "${VM_NAME}" --storagectl IDE --type dvddrive --port 1 --device 0 --medium emptydrive
+vboxmanage startvm "${VM_NAME}" --type headless
+# Once the VM is available, you can log in:
+ssh core@localhost -p 2222
+----
+
+== Installing from raw bare metal image
+
+WARNING: The steps below are provided for advanced users, may unexpectedly break in various places with newer Fedora CoreOS images, and require a Unix environment to work. If you encounter issues consider using the xref:_installing_from_live_iso[live ISO instructions] instead.
+
+To install FCOS into a VirtualBox virtual machine using the raw disk image, follow these steps:
+
+- Download the latest raw image from the https://getfedora.org/coreos/download?tab=metal_virtualized&stream=stable[download page] or with podman (see https://coreos.github.io/coreos-installer/cmd/download/[documentation] for options):
+[source, bash]
+----
+podman run --security-opt label=disable --pull=always --rm -v .:/data -w /data \
+    quay.io/coreos/coreos-installer:release download -f raw
+----
+
+Before the installation can begin you have to convert the raw disk image to a VirtualBox compatible one:
+[source, bash]
+----
+FCOS_DISK=$(realpath ~/Downloads/fedora-coreos-34.20210808.3.0-metal.x86_64.raw.xz)
+UNPACKED_DISK=$(realpath ~/Downloads/fedora-coreos-34.20210808.3.0-metal.x86_64.raw)
+
+# Unpack the raw disk image
+unxz "${FCOS_DISK}"
+----
+
+[NOTE]
+====
+If you want to supply the Ignition configuration file URL statically, you can modify the raw disk before converting it to a VirtualBox disk image.
+
+To correctly mount the boot partition you have to find out the start point of it and the disk's sector size. Multiply the two values to calculate the mount `offset`.
+[source, bash]
+----
+TMP_FCOS_BOOT_VOLUME=$(realpath /tmp/fcos_boot_volume)
+mkdir "${TMP_FCOS_BOOT_VOLUME}"
+
+fdisk -l "${UNPACKED_DISK}"
+sudo mount -o loop,offset="${OFFSET}" "${UNPACKED_DISK}" "${TMP_FCOS_BOOT_VOLUME}"
+----
+
+On macOS you have to replace `fdisk -l "${UNPACKED_DISK}"` with `hdiutil imageinfo -imagekey diskimage-class=CRawDiskImage "${UNPACKED_DISK}"`. Mounting the partition requires https://osxfuse.github.io[macFUSE] with an https://github.com/osxfuse/osxfuse/wiki/Ext[ext filesystem extension].
+
+The boot config you are looking for is located in `/loader/entries/ostree-1-fedora-coreos.conf` on the boot partition you just mounted. The file contains a line containing the string `ignition.platform.id=metal $ignition_firstboot` – add `ignition.config.url=http://10.0.2.2:8000/example.ign` after the first boot parameter, save your changes and unmount the disk.
+[source, bash]
+----
+sudo vi "${TMP_FCOS_BOOT_VOLUME}/loader/entries/ostree-1-fedora-coreos.conf"
+# ...
+sudo umount "${TMP_FCOS_BOOT_VOLUME}"
+----
+====
+
+Continue with converting the raw disk image into a VirtualBox disk image:
+[source, bash]
+----
+vboxmanage convertfromraw "${UNPACKED_DISK}" "${VM_DISK}" --format VDI --variant Standard
+# Resize the disk to match FCOS requirements
+vboxmanage modifymedium disk "${VM_DISK}" --resize 10240
+# Add the converted disk to the VM
+vboxmanage storageattach "${VM_NAME}" --storagectl SATA --type hdd --port 0 --device 0 --medium "${VM_DISK}"
+----
+
+TIP: You can use https://www.virtualbox.org/manual/UserManual.html#vboxmanage-snapshot[snapshots] and https://www.virtualbox.org/manual/UserManual.html#vboxmanage-clonevm[cloning] in case you want to reuse the VM as an empty baseline before applying an Ignition configuration to it.
+
+[NOTE]
+====
+*Fedora CoreOS requires the root filesystem to be at least 8 GiB.* For practical reasons, disk images for some platforms ship with a smaller root filesystem, which by default automatically expands to fill all available disk space. If you add additional partitions after the root filesystem, you must make sure to explicitly resize the root partition as shown so that it is at least 8 GiB.
+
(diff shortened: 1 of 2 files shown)


travier (commented):
Minor comments. LGTM overall but I can not test it.

Okeanos:
@travier thank you for your feedback; I believe I fixed these text issues and applied the correct `podman` download change as well.

Okeanos:
@travier maybe this was lost in the partial GitHub outage last week – if there's anything I still have to update/change before this can be merged please let me know.

travier:
Sorry for the delay.

I'm re-reading that and wondering if we could skip the ISO installation step by directly converting a raw disk image to a VirtualBox drive and then use remote Ignition fetch for first boot provisioning. This could significantly simplify and speed up the process.

What do you think?

Okeanos:
Converting from raw does sound like a good suggestion. I am not sure when I have time to get to test that, though.

The process would be:

1. Download raw
2. [Convert from raw](https://www.virtualbox.org/manual/UserManual.html#vboxmanage-convertfromraw)
3. set [`ignition.config.url`](https://coreos.github.io/ignition/supported-platforms/) kernel parameter
4. start vm
5. done?

Step 3 sounds a little tricky, not sure that can be easily modified from outside the VM via vboxmanage.

I did try out what happens when using the [vboxmanage ovf import](https://www.virtualbox.org/manual/UserManual.html#ovf) as a basis for fun, however: it breaks. There's a hard coded check in there somewhere which expects a VMWare CPU Info to be returned along the boot process. Something VBox obviously doesn't supply.

travier:
> The process would be:
> 
> 1. Download raw
> 2. [Convert from raw](https://www.virtualbox.org/manual/UserManual.html#vboxmanage-convertfromraw)
> 3. set [`ignition.config.url`](https://coreos.github.io/ignition/supported-platforms/) kernel parameter
> 4. start vm
> 5. done?

Yes, that should be it

> Step 3 sounds a little tricky, not sure that can be easily modified from outside the VM via vboxmanage.

I don't think we can change that in an automated fashion. Users will have to edit the command line in GRUB.

> I did try out what happens when using the [vboxmanage ovf import](https://www.virtualbox.org/manual/UserManual.html#ovf) as a basis for fun, however: it breaks. There's a hard coded check in there somewhere which expects a VMWare CPU Info to be returned along the boot process. Something VBox obviously doesn't supply.

Yes, you can only use matching platforms or bare metal on any platform (but without platform integration obviously).

Okeanos:
So I had a little time to checkout the idea of using the raw image:

```sh
VM_NAME='fcos-node01'
# vboxmanage always expects absolute file paths
VM_DISK=$(realpath ~/VirtualBox\ VMs/${VM_NAME})/${VM_NAME}.vdi
FCOS_DISK=$(realpath ~/Downloads/fedora-coreos-34.20210808.3.0-metal.x86_64.raw.xz)
PADDED_DISK=$(realpath ~/Downloads/fedora-coreos-34.20210808.3.0-metal.x86_64.raw.xz.padded)

vboxmanage createvm --name "${VM_NAME}" --ostype Fedora_64 --register
# Set hardware settings for the VM, including recommended graphics controller
vboxmanage modifyvm "${VM_NAME}" --memory 4096 --cpus 2 --vram 20 \
  --graphicscontroller vmsvga --rtcuseutc on --usb on --usbehci on
# Add a storage controller for hard disks
vboxmanage storagectl "${VM_NAME}" --name SATA --add sata --controller IntelAhci \
  --portcount 30 --bootable on
# Allow logging in via ssh after installation is done
vboxmanage modifyvm "${VM_NAME}" --natpf1 "guestssh,tcp,,2222,,22"

# use dd to make the raw image readable by vboxmanage using 512 bytes as sector boundary
dd if="${FCOS_DISK}" of="${PADDED_DISK}" bs=512 conv=sync
vboxmanage convertfromraw "${PADDED_DISK}" "${VM_DISK}" --format VDI --variant Standard
# Resize the disk to match FCOS requirements
# vboxmanage modifymedium disk "${VM_DISK}" --resize 10240
# Add the converted disk to the VM
vboxmanage storageattach "${VM_NAME}" --storagectl SATA --type hdd --port 0 --device 0 \
  --medium "${VM_DISK}"
```

Regardless of whether the basic or the 4k raw image is used (and regardless of whether the VDI is resized to 10GB or not) the resulting VM cannot boot because it cannot find a bootable system on the disk. The default setup for VBox is to use BIOS (if that makes any difference). Using VBox EFI is a little weird and I haven't looked at it too closely for now.

The `dd` copy is necessary because otherwise the following happens during `convertfromraw`: 

```sh
$ vboxmanage convertfromraw "${FCOS_DISK}" "${VM_DISK}" --format VDI --variant ESX
Converting from raw image file="~/Downloads/fedora-coreos-34.20210808.3.0-metal.x86_64.raw.xz" to file="~/VirtualBox VMs/fcos-node01/fcos-node01.vdi"...
Creating dynamic image with size 632932832 bytes (604MB)...
VBoxManage: error: VD: The given disk size 632932832 is not aligned on a sector boundary (512 bytes)
VBoxManage: error: Error code VERR_VD_INVALID_SIZE at /Users/vbox/tinderbox/6.1-mac-rel/src/VBox/Storage/VD.cpp(6242) in function int VDCreateBase(PVDISK, const char *, const char *, uint64_t, unsigned int, const char *, PCVDGEOMETRY, PCVDGEOMETRY, PCRTUUID, unsigned int, PVDINTERFACE, PVDINTERFACE)
VBoxManage: error: Cannot create the disk image "~/VirtualBox VMs/fcos-node01/fcos-node01.vdi": VERR_VD_INVALID_SIZE
```

Any ideas?

Okeanos:
Okay, got it working after realising the raw disk image is compressed (🤦‍♂️) – I updated the pull request accordingly.

Technically, I believe the following would also work for unpacking, converting and sizing the disk image correctly but I didn't try it yet. Also I am not sure this condensed example helps with readability.

```bash
unxz --stdout "${FCOS_DISK}" | vboxmanage convertfromraw stdin "${VM_DISK}" 10737418240 --format VDI --variant Standard
```

The `10737418240` represents the VM disk's size in bytes. No fancy notation possible.

travier (approved):
LGTM overall. The second approach as the advantage that you can prepare the disk once and reuse it for several VMs (either as copy or as overlays if VirtualBox has something like that?). Will make another pass for typos soon.

Okeanos:
> LGTM overall. The second approach as the advantage that you can prepare the disk once and reuse it for several VMs (either as copy or as overlays if VirtualBox has something like that?). Will make another pass for typos soon.

VirtualBox supports [snapshots](https://www.virtualbox.org/manual/UserManual.html#vboxmanage-snapshot) for returning to previous states of a VM and [cloning](https://www.virtualbox.org/manual/UserManual.html#vboxmanage-clonevm) to make an exact copy of another VM and/or its snapshots, so yes.
That is indeed an advantage and I think I'll add that as a "Tip" before the "now let's run ignition" paragraph.

Okeanos:
So, as I suspected … modifying the raw disk image before converting it totally works!

I added an additional note that shows how one can add a static Ignition configuration URL to the boot parameters inside of the `/boot` partition. It's even possible to override this value during actual first boot.

Makes supplying the Ignition URL way more comfortable as the GRUB pre-boot selection screen is only available for a very short time.

travier:
Awesome! Do we still need two ways to do this then? Maybe we should only focus on the direct image tweak and import. We could then keep the LiveISO case in second or completely remove it if you think the disk case is strong enough.

Thanks a lot for working on that!

Okeanos:
I haven’t done extensive checks yet but the procedure is somewhat involved on macOS requiring at least an alternative to `fdisk` and a FUSE extension for ext3/4 to check partition boundaries and being able to mount it.
No idea what’s required on Windows.

Hence, I personally would like to keep an OS agnostic option in the docs.

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -0,0 +1,186 @@
...
+
+This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualBox hypervisor.
+
+== Prerequisites
+
+Before provisioning an FCOS machine, you must have an Ignition configuration file containing your customizations and host it somewhere (e.g. using `python3 -m http.server`, we will assume this option is used in the following examples). If you do not alrady have an Ignition file, see xref:producing-ign.adoc[Producing an Ignition File].
+
+== Setting up a new VM
+
+You can do set up VirtualBox based Virtual Machine through the GUI or via the https://www.virtualbox.org/manual/UserManual.html#vboxmanage[`vboxmanage` CLI]. Below is a shell script that does most of what the graphical wizard will do, however, is tailored to the installation from live ISO or Raw disk image as outlined in the following sections.
```
travier, line 11:
```suggestion
You can set up a VirtualBox based Virtual Machine through the GUI or via the https://www.virtualbox.org/manual/UserManual.html#vboxmanage[`vboxmanage` CLI]. Below is a shell script that does most of what the graphical wizard will do. Note that it is tailored to the installation from live ISO or Raw disk image as outlined in the following sections.
```

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -0,0 +1,186 @@
...
+fdisk -l "${UNPACKED_DISK}"
+# calculate the $OFFSET for the mount operation based on the sector size of the disk and the start point of the boot partion
+sudo mount -o loop,offset="${OFFSET}" fedora-coreos-34.20210808.3.0-metal.x86_64.raw "${TMP_FCOS_BOOT_VOLUME}"
+# the boot config you are looking for is located in /loader/entries/ostree-1-fedora-coreos.conf on the boot partition
+sudo vi "${TMP_FCOS_BOOT_VOLUME}/loader/entries/ostree-1-fedora-coreos.conf"
+# ...
+sudo umount "${TMP_FCOS_BOOT_VOLUME}"
+----
+
+The file contains a line containing the string `ignition.platform.id=metal $ignition_firstboot` – add `ignition.config.url=http://10.0.2.2:8000/example.ign` after the first boot parameter and save it. You can now unmount the disk.
```
travier, line 130: Maybe split the code block in two and move this line right under the vi command?

travier (approved):
This looks great! Thanks a lot for working on this. Just a couple of minor suggestions and we're good to go.

modules/ROOT/pages/provisioning-virtualbox.adoc
Okeanos: Fixed

modules/ROOT/pages/provisioning-virtualbox.adoc
Okeanos: I split that note up a little, please check whether that improves readability.

travier:
Let's give @dustymabe @lucab @bgilbert a chance to take a look but otherwise I'll merge this one soon.

travier (approved):

Okeanos:
I added a final note for the macOS difference when inspecting and mounting the raw disk image. Windows is out of luck – I don't think there's a reasonably easy way to achieve this.

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -0,0 +1,191 @@
+= Provisioning Fedora CoreOS on VirtualBox
+
+This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualBox hypervisor.
```
jlebon, line 3: It feels like we should right upfront mention that FCOS currently does not ship VirtualBox images but may in the future and that what follows are platform-agnostic ways to set up Fedora CoreOS as a bare metal system.

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -0,0 +1,191 @@
...
+
+[source, bash]
+----
+vboxmanage storageattach "${VM_NAME}" --storagectl IDE --type dvddrive --port 1 --device 0 --medium emptydrive
+vboxmanage startvm "${VM_NAME}" --type headless
+# Once the VM is available, you can log in:
+ssh core@localhost -p 2222
+----
+
+== Installing from Raw
```
jlebon, line 92: I appreciate the work that went behind writing and testing this. But personally I'm hesitant to document it as if it were a supported workflow. I think to start, just having the ISO install path would be great. Or otherwise if folks really want, maybe we should at least have a disclaimer around here that this is not heavily tested and may break.

Okeanos: That somewhat mirrors my own thoughts, in particular concerning the actual locations of the config file, disk name etc. At the same time this process works spectacularly well (imho better than the regular ISO install with the cumbersome "press e at the right time" issue) if the user is already on Linux. Also, technically I should've probably used `parted` instead of `fdisk` but in the end it probably doesn't matter too much.
I am definitely in favor of the disclaimer, though.

modules/ROOT/pages/provisioning-virtualbox.adoc
Okeanos: That's a good point. Would it be alright to add that hint together with links to the two issues ([#144](https://github.com/coreos/fedora-coreos-tracker/issues/144), [#73](https://github.com/coreos/fedora-coreos-tracker/issues/73)) related to providing VirtualBox images?

modules/ROOT/pages/provisioning-virtualbox.adoc
travier: Let's make a big warning and keep it? It's also already mentioned in a second time after the ISO. I understand it's a bit fragile but it's a nice performance improvement for non casual users.

Otherwise we could split this into 2 PRs to merge the first part and make a new PR to discuss improvements for the raw disk part.

modules/ROOT/pages/provisioning-virtualbox.adoc
jlebon: Yes, that sounds great!

modules/ROOT/pages/provisioning-virtualbox.adoc
jlebon: Yup sure, I'm OK with keeping it and just adding a warning.

Okeanos: I updated the pull request with such a warning.

modules/ROOT/pages/provisioning-virtualbox.adoc
Okeanos: I updated the pull request with such a hint.

travier (approved):
LGTM. Thanks for working on that. Looks great!

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -0,0 +1,195 @@
+= Provisioning Fedora CoreOS on VirtualBox
+
+This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualBox hypervisor.
+
+== Prerequisites
+
+Before provisioning an FCOS machine, you must have an Ignition configuration file containing your customizations and host it somewhere (e.g. using `python3 -m http.server`, we will assume this option is used in the following examples). If you do not alrady have an Ignition file, see xref:producing-ign.adoc[Producing an Ignition File].
```
jlebon, line 7:
```suggestion
Before provisioning an FCOS machine, you must have an Ignition configuration file containing your customizations and host it somewhere (e.g. using `python3 -m http.server`, we will assume this option is used in the following examples). If you do not already have an Ignition file, see xref:producing-ign.adoc[Producing an Ignition File].
```

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -0,0 +1,195 @@
...
+
+== Setting up a new VM
+
+You can set up a VirtualBox based Virtual Machine through the GUI or via the https://www.virtualbox.org/manual/UserManual.html#vboxmanage[`vboxmanage` CLI]. Below is a shell script that does most of what the graphical wizard will do. Note that it is tailored to the installation from live ISO or Raw disk image as outlined in the following sections.
```
jlebon, line 11:
```suggestion
You can set up a VirtualBox virtual machine through the GUI or via the https://www.virtualbox.org/manual/UserManual.html#vboxmanage[`vboxmanage` CLI]. Below is a shell script that does most of what the graphical wizard will do. Note that it is tailored to the installation from live ISO or raw disk image as outlined in the following sections.
```

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -0,0 +1,195 @@
...
+
+NOTE: Please be aware that for now there is no dedicated Fedora CoreOS image available for VirtualBox and the installation instructions provided are essentially xref:bare-metal.adoc[bare metal instructions]. A dedicated image may be provided at a later time. Please check the GitHub issues https://github.com/coreos/fedora-coreos-tracker/issues/73[#73] and https://github.com/coreos/fedora-coreos-tracker/issues/144[#144] for more details.
+
+We will rely on the VirtualBox provided NAT gateway to access the hosts's loopback interface on `10.0.2.2`. Check the https://www.virtualbox.org/manual/UserManual.html#network_nat[VirtualBox Documentation] for more details.
```
jlebon, line 15:
```suggestion
We will rely on the VirtualBox-provided NAT gateway to access the hosts' loopback interface on `10.0.2.2`. Check the https://www.virtualbox.org/manual/UserManual.html#network_nat[VirtualBox User Manual] for more details.
```

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -0,0 +1,195 @@
...
+
+[source, bash]
+----
+VM_NAME='fcos-node01'
+VM_DISK=$(realpath ~/VirtualBox\ VMs/${VM_NAME})/${VM_NAME}.vdi
+
+vboxmanage createvm --name "${VM_NAME}" --ostype Fedora_64 --register
+# Set hardware settings for the VM, including recommended graphics controller
+vboxmanage modifyvm "${VM_NAME}" --memory 4096 --cpus 2 --vram 20 --graphicscontroller vmsvga --rtcuseutc on --usb on --usbehci on
```
jlebon, line 24: Is USB actually required?

modules/ROOT/pages/provisioning-virtualbox.adoc
jlebon:
```suggestion
== Installing from raw bare metal image
```

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -0,0 +1,195 @@
...
+vboxmanage startvm "${VM_NAME}" --type headless
+# Once the VM is available, you can log in:
+ssh core@localhost -p 2222
+----
+
+== Installing from Raw
+
+WARNING: The steps below are provided for advanced users, may unexpectedly break in various places with newer Fedora CoreOS images, and require a Unix environment to work. If you encounter issues consider using the xref:_installing_from_live_iso[live ISO instructions] instead.
+
+To install FCOS into a VirtualBox virtual machine using the Raw disk image, follow these steps:
```
jlebon, line 98:
```suggestion
To install FCOS into a VirtualBox virtual machine using the raw disk image, follow these steps:
```

jlebon (approved):
Minor stylistic things, but cool to fix in a follow-up too.
Thanks a lot!

modules/ROOT/pages/provisioning-virtualbox.adoc
Okeanos: Fair question: I doubt it but as described in the leading paragraph, the shell command(s) are designed to give a very similar output the GUI wizard. I can test that again without USB and fix this later accordingly.

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -0,0 +1,195 @@
...
+
+== Setting up a new VM
+
+You can set up a VirtualBox virtual machine through the GUI or via the https://www.virtualbox.org/manual/UserManual.html#vboxmanage[`vboxmanage` CLI]. Below is a shell script that does most of what the graphical wizard will do. Note that it is tailored to the installation from live ISO or raw disk image as outlined in the following sections.
+
+NOTE: Please be aware that for now there is no dedicated Fedora CoreOS image available for VirtualBox and the installation instructions provided are essentially xref:bare-metal.adoc[bare metal instructions]. A dedicated image may be provided at a later time. Please check the GitHub issues https://github.com/coreos/fedora-coreos-tracker/issues/73[#73] and https://github.com/coreos/fedora-coreos-tracker/issues/144[#144] for more details.
+
+We will rely on the VirtualBox.provided NAT gateway to access the hosts' loopback interface on `10.0.2.2`. Check the https://www.virtualbox.org/manual/UserManual.html#network_nat[VirtualBox User Manual] for more details.
```
jlebon, line 15:
```suggestion
We will rely on the VirtualBox-provided NAT gateway to access the hosts' loopback interface on `10.0.2.2`. Check the https://www.virtualbox.org/manual/UserManual.html#network_nat[VirtualBox User Manual] for more details.
```

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -0,0 +1,195 @@
...
+[source, bash]
+----
+vboxmanage storageattach "${VM_NAME}" --storagectl IDE --type dvddrive --port 1 --device 0 --medium emptydrive
+vboxmanage startvm "${VM_NAME}" --type headless
+# Once the VM is available, you can log in:
+ssh core@localhost -p 2222
+----
+
+== Installing from Raw
+
```
jlebon, line 95:
```suggestion
== Installing from raw bare metal image

```

jlebon (approved):

modules/ROOT/pages/provisioning-virtualbox.adoc
Okeanos: Ooops 😅

modules/ROOT/pages/provisioning-virtualbox.adoc
Okeanos: Changed.

jlebon (approved):
Thanks a lot!

PR #319, opened 2021-09-05
Fix misspellings

rffontenelle:


miabbott (approved):
Thanks for your contribution!

PR #321, opened 2021-09-26
Improve VMWare documentation with gzip+base64 information

Okeanos:
Ignition supports specifying configs in gzip+base64 format for VMWare. This alleviates the issue of not being able to use large inline config values with govc (see [govmomi/govc#2488](https://github.com/vmware/govmomi/issues/2488)) in some cases.

This should also address #50 in part.

I switched to using `$()` syntax for the encode snippets from `` syntax as well to make the use of a subshell more apparent.


lucab:
LGTM, thanks!

PR #327, opened 2021-10-15
provisioning-vmware: fix explanation of govc argument limit

bgilbert:
The relevant limit is OS-dependent and isn't always `ARG_MAX`.  On Linux it's `MAX_ARG_STRLEN`, and on Windows, `cmd.exe` and PowerShell impose lower limits than the kernel does.

While we're here, switch `gzip` to use max compression.

cc @Okeanos


bgilbert:
Thanks for testing!  The macOS limit you found is smaller than 256 KiB, which is 262144 bytes.  I wouldn't be surprised if the size of the environment block is also factored into that limit.

PR #338, opened 2021-11-24
Add xrefs and reorder final few tutorials

jldohmann:
Hi, thanks for these tutorials! It's helped me a lot as I learn more about CoreOS

I noticed that `tutorial-updates.adoc` has a conclusion, which means it's probably meant to be the final tutorial, as there is no `xref` linking it to `tutorial-user-systemd-unit-on-boot.adoc`

This PR reorders the tutorials, so that _Testing Fedora CoreOS updates_ is the final one, and the one before is _Launching a user-level systemd unit on boot_. The links are also updated to reflect this.

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -54,8 +54,8 @@
 ** xref:tutorial-autologin.adoc[Enabling autologin and custom hostname]
 ** xref:tutorial-services.adoc[Starting a service on first boot]
 ** xref:tutorial-containers.adoc[SSH access and starting containers]
-** xref:tutorial-updates.adoc[Testing Fedora CoreOS updates]
 ** xref:tutorial-user-systemd-unit-on-boot.adoc[Launching a user-level systemd unit on boot]
+** xref:tutorial-updates.adoc[Testing Fedora CoreOS updates]
 * Reference pages
 ** xref:platforms.adoc[Platforms]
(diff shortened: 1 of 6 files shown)


travier (approved):

travier:
You also need to re-order them in https://github.com/coreos/fedora-coreos-docs/blob/main/modules/ROOT/pages/tutorial-setup.adoc.

Maybe we should split the conclusion in it's own page at the end to avoid this issue the next time we had a tutorial page.

jldohmann:
> Thanks! Can you also re-order them in https://github.com/coreos/fedora-coreos-docs/blob/main/modules/ROOT/pages/tutorial-setup.adoc?
> 
> Maybe we should split the conclusion in it's own page at the end to avoid this issue the next time we had a tutorial page.

Sure! And I think that's a good idea: modularize the conclusion so that it's just an include statement that gets moved around, I'll add it to the PR. I think there's a lot of similar such cases (e.g. the cleanup sections in each tutorial), and I can make a separate PR for more modularization efforts across the docs.

jldohmann:
97e3e5e...b2c40a8

- reorder `tutorial-setup.adoc`
- modularize tutorial conclusion

travier (approved):
LGTM

jlebon (approved):
LGTM too, thanks!

PR #341, opened 2021-12-02
Don't use '--tty-' for podman shell aliases

travier:
Using '--tty' conflicts with stdin redirection:
https://docs.podman.io/en/latest/markdown/podman-run.1.html#tty-t

### DIFF
--- a/modules/ROOT/pages/producing-ign.adoc
+++ b/modules/ROOT/pages/producing-ign.adoc
@@ -57,7 +57,7 @@ To make it simpler to type, you may also add the following alias to your shell c
 
 [source,bash]
 ----
-alias butane='podman run --rm --tty --interactive \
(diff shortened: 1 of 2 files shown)


travier:
We probably still need it for coreos-installer to be able to cancel downloads.

dustymabe (approved):
LGTM

bgilbert (approved):
coreos-installer has a weird tradeoff: we need a TTY to cancel downloads, but need to not have one for the caller to retrieve the path to the downloaded file.

So yeah, might as well convert everything for consistency,

travier:
I'd opt for keeping things simple here in the tutorial and adding the details to the coreos-installer docs: https://github.com/coreos/coreos-installer/pull/703

PR #344, opened 2021-12-11
provisioning/rpi4: add warning around use of toolbx

miabbott:
When I ran through the RPi4 docs myself, I encountered issues trying to
mix and match commands run from inside my Toolbx container and on the
host itself. I've added a warning about running certain commands outside
the Toolbx container, which allowed me to install FCOS successfully on
my RPi4.

### DIFF
--- a/modules/ROOT/pages/provisioning-raspberry-pi4.adoc
+++ b/modules/ROOT/pages/provisioning-raspberry-pi4.adoc
@@ -13,14 +13,14 @@ The Raspberry Pi 4 uses an EEPROM to boot the system. The EEPROM/Firmware in the
 
 NOTE: Ignore the October 8th 2021 release as it was just a repackaging of a previous release.
 
-The https://www.raspberrypi.org/documentation/computers/raspberry-pi.html#updating-the-bootloader[Raspberry Pi Documentation] recommends using the Raspberry Pi Imager for creating a boot disk that can be used to update the EEPROM. If you're on a flavor of Fedora Linux the Raspberry Pi Imager is packaged up and available in the repositories. You can install it with: 
+The https://www.raspberrypi.org/documentation/computers/raspberry-pi.html#updating-the-bootloader[Raspberry Pi Documentation] recommends using the Raspberry Pi Imager for creating a boot disk that can be used to update the EEPROM. If you're on a flavor of Fedora Linux the Raspberry Pi Imager is packaged up and available in the repositories. You can install it with:
 
 [source, bash]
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/provisioning-raspberry-pi4.adoc
```diff
@@ -45,6 +45,8 @@ sudo dnf install -y --downloadonly --release=$RELEASE --forcearch=aarch64 --dest
 
 Now extract the contents of the RPMs and copy the proper `u-boot.bin` for the RPi4 into place:
 
+WARNING: The following commands to extract the contents of the RPMs, the use of `coreos-installer`, and copying of the files to the ESP partition should be done from outside of a Toolbx container.
```
dustymabe, line 48: I wonder if we should add some explanation of why. I think the issue you were hitting was because `root` inside the toolbx container was not the same as `UID=0` on the host.

modules/ROOT/pages/provisioning-raspberry-pi4.adoc
```diff
@@ -76,7 +78,6 @@ Now take the USB/SD card and attach it to the RPi4 and boot.
 
 TIP: It can take some time to boot, especially if the disk is slow. Be patient. You may not see anything on the screen for 20-30 seconds.
 
-
```
dustymabe, line 79: Do we need to remove these extra line breaks? When viewing the file plain text I like having the extra break before the start of a new section. Maybe the issue here is consistency?

modules/ROOT/pages/provisioning-raspberry-pi4.adoc
miabbott: Sure, I can expand on this.

modules/ROOT/pages/provisioning-raspberry-pi4.adoc
miabbott: They didn't seem needed to me; I don't think the rendering changes, I'll confirm.

If you have a preference for the additional breaks, it's fine to leave them in.

miabbott: Render changed slightly; I added them back in

dustymabe (approved):
LGTM

modules/ROOT/pages/provisioning-raspberry-pi4.adoc
dustymabe: reviewed!

PR #348, opened 2022-01-06
De-duplicate platform support and identifier docs

travier:
Avoid duplication (and thus outdated information) and consolidate the
supported platform list in its dedicated page outside of the FAQ.

### DIFF
--- a/modules/ROOT/pages/platforms.adoc
+++ b/modules/ROOT/pages/platforms.adoc
@@ -1,12 +1,40 @@
-= Platforms
+= Supported Platforms
 
 Fedora CoreOS is provisioned via prebuilt disk images, and configured on first-boot via https://github.com/coreos/ignition[Ignition]. Each platform may require specific logic and components, thus dedicated images are provided for each supported environment. Additionally, a unique platform ID is available in the host environment for runtime introspection.
 
-== Runtime introspection
+The currently supported platforms and their identifiers are listed below.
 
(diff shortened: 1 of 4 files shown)


dustymabe:
Hmm.. While what we currently have is problematic we (FCOS) don't support every platform that Ignition supports, so the update is misleading IMO. For example we don't have a Brightbox FCOS image we produce.

Similarly we support and document running on things like the Raspberry Pi4. Not sure if we should list things like that here or not.

travier:
We might need to add arch support info too indeed.

modules/ROOT/pages/platforms.adoc
```diff
@@ -1,12 +1,31 @@
-= Platforms
+= Supported platforms
 
 Fedora CoreOS is provisioned via prebuilt disk images, and configured on first-boot via https://github.com/coreos/ignition[Ignition]. Each platform may require specific logic and components, thus dedicated images are provided for each supported environment. Additionally, a unique platform ID is available in the host environment for runtime introspection.
 
-== Runtime introspection
+The currently supported platforms and their identifier are:
```
dustymabe, line 5: Should this be plural?

```suggestion
The currently supported platforms and their identifiers are:
```

dustymabe (approved):
LGTM

dustymabe (approved):
LGTM

PR #349, opened 2022-01-06
os-extensions: Update page to use vim instead of nano

travier:
We now include nano by default thus update this example to install vim
as an os extension via package layering.

### DIFF
--- a/modules/ROOT/pages/os-extensions.adoc
+++ b/modules/ROOT/pages/os-extensions.adoc
@@ -9,10 +9,11 @@ To do this, you can use https://coreos.github.io/rpm-ostree/[`rpm-ostree install
 To start the layering of a package, you need to write a systemd unit that executes the `rpm-ostree` command to install the wanted package(s).
 By default with `rpm-ostree install`, changes are queued for the next boot.  The `-A/--apply-live` option can be used to apply changes live *and* have them persist.
 
-== Example: Layering nano and setting it as the default editor
+== Example: Layering vim and setting it as the default editor
 
-This example shows on how to install the text editor `nano` and add a script to `/etc/profile.d` so that it is set as the default editor for all users.
-The parameter `--allow-inactive` is useful if the package is added to the root image in a future Fedora CoreOS release. In such a case, the parameter prevents that the service would fail.
+Fedora CoreOS includes both `nano` and `vi` as text editors, with the former set as default (see the corresponding https://fedoraproject.org/wiki/Changes/UseNanoByDefault[Fedora change]).
+
+This example shows how to install the fully fledged `vim` text editor and how to set it up as default for all users by setting up the required configuration in `/etc/profile.d/`.
 
 NOTE: In the future, we will have a more Ignition-friendly method of doing this with stronger guarantees. See upstream issues https://github.com/coreos/butane/issues/81[butane#81] and https://github.com/coreos/fedora-coreos-tracker/issues/681[fedora-coreos-tracker#681] for more information.
 
@@ -22,12 +23,12 @@ variant: fcos
 version: 1.4.0
 systemd:
   units:
-    # installing nano as a layered package with rpm-ostree
-    - name: rpm-ostree-install-nano.service
+    # Installing vim as a layered package with rpm-ostree
+    - name: rpm-ostree-install-vim.service
       enabled: true
       contents: |
         [Unit]
-        Description=Layer nano with rpm-ostree
+        Description=Layer vim with rpm-ostree
         Wants=network-online.target
         After=network-online.target
         # We run before `zincati.service` to avoid conflicting rpm-ostree
@@ -38,18 +39,24 @@ systemd:
         [Service]
         Type=oneshot
         RemainAfterExit=yes
-        ExecStart=/usr/bin/rpm-ostree install --apply-live --allow-inactive nano
+        # `--allow-inactive` ensures that rpm-ostree does not return an error
+        # if the package is already installed. This is useful if the package is
+        # added to the root image in a future Fedora CoreOS release as it will
+        # prevent the service from failing.
+        ExecStart=/usr/bin/rpm-ostree install --apply-live --allow-inactive vim
         ExecStart=/bin/touch /var/lib/%N.stamp
 
         [Install]
         WantedBy=multi-user.target
 storage:
   files:
-    # use nano as default editor
-    - path: /etc/profile.d/nano.sh
+    # Set vim as default editor
+    # We use `zz-` as prefix to make sure this is processed last in order to
+    # override any previously set defaults.
+    - path: /etc/profile.d/zz-default-editor.sh
       overwrite: true
       contents:
         inline: |
           #/bin/sh
-          export EDITOR=nano
+          export EDITOR=vim
 ----


modules/ROOT/pages/os-extensions.adoc
```diff
@@ -38,18 +39,27 @@ systemd:
...
         ExecStart=/bin/touch /var/lib/%N.stamp
 
         [Install]
         WantedBy=multi-user.target
 storage:
   files:
-    # use nano as default editor
-    - path: /etc/profile.d/nano.sh
+    # Remove the existing default editor configuration
+    - path: /etc/profile.d/
```
travier, line 54: Exact path missing here

modules/ROOT/pages/os-extensions.adoc
```diff
@@ -38,18 +39,27 @@ systemd:
...
+      overwrite: true
+      contents:
+        inline: ""
```
travier, line 57: Needs to be checked

modules/ROOT/pages/os-extensions.adoc
```diff
@@ -38,18 +39,27 @@ systemd:
...
+    # Set vim as default editor
+    - path: /etc/profile.d/vim-default-editor.sh
```
dustymabe, line 59: WDYT about making it `/etc/profile.d/z-default-editor.sh` so it would work even if we were intalling a different editor.

modules/ROOT/pages/os-extensions.adoc
```diff
@@ -38,18 +39,27 @@ systemd:
...
+        ExecStart=/usr/bin/rpm-ostree install --apply-live --allow-inactive vim
         ExecStart=/bin/touch /var/lib/%N.stamp
 
         [Install]
         WantedBy=multi-user.target
 storage:
   files:
-    # use nano as default editor
-    - path: /etc/profile.d/nano.sh
+    # Remove the existing default editor configuration
```
dustymabe, line 53: Maybe let's not blank out this file and instead write our new preference to something like `/etc/profile.d/z-default-editor.sh` to override it? see https://github.com/coreos/fedora-coreos-config/pull/1394#issuecomment-1007488297

travier:
Commented in https://github.com/coreos/fedora-coreos-config/pull/1394#issuecomment-1007561996

I'll try the ordering option. Wondering where to place that section in the docs too.

travier:
Replying here from https://github.com/coreos/fedora-coreos-config/pull/1394 to keep the conversation in the right place.

> What's confusing? It's alphabetical sorting and last one wins. Just make our file name start with `z` and add a sentence explaining why.

It might be confusing that we have both configs installed at the same time thus you have to realize that they both get loaded, that the order matters, etc. I agree we can add an explanation but blanking the other config makes it "simpler" as you don't really have to explain anything anymore, you only have the right one.

> > > This page is good for an OS Extensions example, but I feel it would be good to have a separate section somewhere specifically targeted at the editor. For example, what if I just want to set my editor back to `vi` and not `vim`. I shouldn't have to look at the os extensions doc for that.
> > 
> > 
> > Agree. I did not find a good place to add that yet but I can create another page. It's also a sort of _one-liner_:
> > ```
> > $ sed -i 's/EDITOR=nano/EDITOR=vi/g /etc/profile.d/nano-*
> > ```
> 
> ehh. I would make it a butane snippet that defines a file `/etc/profile.d/z-default-editor.sh` and not a dynamically run command (though providing the dynamic command to run could be useful for people who installed first and then came to the docs).

Will add both. Would user configuration be a good place? Creating a page just for that feels a bit too much but we can still do it.

dustymabe:
> Replying here from [coreos/fedora-coreos-config#1394](https://github.com/coreos/fedora-coreos-config/pull/1394) to keep the conversation in the right place.
> 
> > What's confusing? It's alphabetical sorting and last one wins. Just make our file name start with `z` and add a sentence explaining why.
> 
> It might be confusing that we have both configs installed at the same time thus you have to realize that they both get loaded, that the order matters, etc. I agree we can add an explanation but blanking the other config makes it "simpler" as you don't really have to explain anything anymore, you only have the right one.

I'm just not really a big fan of blanking the other config (for example, now ostree tracks it as a change we made to what was delivered by the OS). If we were to go that route I'd rather just tell the users to `rpm-ostree override remove` the package. I think the user isn't really going to dig into this too much if the steps we give them work. If they do dig they'll realize in the docs we explained it to them.

Obviously this "overriding" behavior would seem more natural if there was a /usr/lib/profile.d/ with OS delivered defaults and /etc/profile.d was just user edits, similar to the way systemd works.

 
> > ehh. I would make it a butane snippet that defines a file `/etc/profile.d/z-default-editor.sh` and not a dynamically run command (though providing the dynamic command to run could be useful for people who installed first and then came to the docs).
> 
> Will add both. Would user configuration be a good place? Creating a page just for that feels a bit too much but we can still do it.

I agree there doesn't seem like a logical place right now, but I think this generic area (i.e. user preference customizations) might deserve its own page in the future. Could we think of an appropriate "category" for this? Then we could create a new page for that category and let this be the first entry there.

travier:
> Obviously this "overriding" behavior would seem more natural if there was a /usr/lib/profile.d/ with OS delivered defaults and /etc/profile.d was just user edits, similar to the way systemd works.

Indeed, this is the real fix. Let's do it as you suggested for now until (if ever) this is fixed.

> I agree there doesn't seem like a logical place right now, but I think this generic area (i.e. user preference customizations) might deserve its own page in the future. Could we think of an appropriate "category" for this? Then we could create a new page for that category and let this be the first entry there.

We could also include an example setting some rlimits for users via `/etc/security/limits.d`.

Suggestions:

* Global user configuration
* Global user limits and configuration
* Other global configuration
* Global configuration examples

modules/ROOT/pages/os-extensions.adoc
```diff
@@ -38,18 +39,24 @@ systemd:
         [Service]
         Type=oneshot
         RemainAfterExit=yes
-        ExecStart=/usr/bin/rpm-ostree install --apply-live --allow-inactive nano
+        # `--allow-inactive` ensure that rpm-ostree does not return an error if
```
dustymabe, line 42:
```suggestion
        # `--allow-inactive` ensures that rpm-ostree does not return an error if
```

modules/ROOT/pages/os-extensions.adoc
```diff
@@ -9,10 +9,11 @@ To do this, you can use https://coreos.github.io/rpm-ostree/[`rpm-ostree install
 To start the layering of a package, you need to write a systemd unit that executes the `rpm-ostree` command to install the wanted package(s).
 By default with `rpm-ostree install`, changes are queued for the next boot.  The `-A/--apply-live` option can be used to apply changes live *and* have them persist.
 
-== Example: Layering nano and setting it as the default editor
+== Example: Adding vim and setting it as the default editor
```
dustymabe, line 12: Not sure if I prefer `Layering vim` to `Adding vim` here, but just wanted to bring it up as something that crossed my mind.

dustymabe (approved):
LGTM. Some suggestions.

modules/ROOT/pages/os-extensions.adoc
travier: Reverted to layering to keep things explicit

dustymabe (approved):

jlebon (approved):

PR #351, opened 2022-01-15
Update to reflect coreos-installer instructions

jackmonty:
From https://coreos.github.io/coreos-installer/getting-started/#run-from-a-container, added additional switches to `podman run` command. This prevents `Error: building HTTP client` from occuring.

### DIFF
--- a/modules/ROOT/pages/getting-started-libvirt.adoc
+++ b/modules/ROOT/pages/getting-started-libvirt.adoc
@@ -8,7 +8,8 @@ STREAM="stable"
 # as an installed binary:
 coreos-installer download -s "${STREAM}" -p qemu -f qcow2.xz --decompress -C ~/.local/share/libvirt/images/
 # or as a container:
-podman run --pull=always --rm -v $HOME/.local/share/libvirt/images/:/data -w /data \
+podman run --pull=always --privileged --rm \
+    -v /dev:/dev -v /run/udev:/run/udev -v $HOME/.local/share/libvirt/images/:/data -w /data \
     quay.io/coreos/coreos-installer:release download -s "${STREAM}" -p qemu -f qcow2.xz --decompress
 ----
 +


travier:
In this case this is only to download a QCOW2 image and not to install on the host so you shouldn't need that kind of access. This is unexpected. Will have to give it a try.

bgilbert:
On an SELinux system you'll need `--security-opt label=disable` (and we should add that to the docs), but on my system, the download works correctly with that change.  Could you post the full error message you're receiving?

jackmonty:
Here is the error I am getting when attempting to run the container using the original command:

```bash
$ podman run --pull=always --rm -v /srv/images/:/data -w /data quay.io/coreos/coreos-installer:release download -s "stable" -p qemu -f qcow2.xz --decompress

Trying to pull quay.io/coreos/coreos-installer:release...
Getting image source signatures
Copying blob 76dc685b05ee skipped: already exists  
Copying blob f949d798783d skipped: already exists  
Copying blob de49e9a09291 skipped: already exists  
Copying blob a3ed95caeb02 done  
Writing manifest to image destination
Storing signatures
Downloading Fedora CoreOS stable x86_64 qemu image (qcow2.xz) and signature
Error: building HTTP client

Caused by:
    0: builder error: Operation not permitted (os error 1)
    1: Operation not permitted (os error 1)
```

Here is the error when adding the `--security-opt label=disable` switch to the original command:

```bash
$ podman run --pull=always --rm --security-opt label=disable -v /srv/images/:/data -w /data quay.io/coreos/coreos-installer:release download -s "stable" -p qemu -f qcow2.xz --decompress

Trying to pull quay.io/coreos/coreos-installer:release...
Getting image source signatures
Copying blob 76dc685b05ee skipped: already exists  
Copying blob de49e9a09291 skipped: already exists  
Copying blob f949d798783d skipped: already exists  
Copying blob a3ed95caeb02 done  
Writing manifest to image destination
Storing signatures
Downloading Fedora CoreOS stable x86_64 qemu image (qcow2.xz) and signature
Error: building HTTP client

Caused by:
    0: builder error: Operation not permitted (os error 1)
    1: Operation not permitted (os error 1)
```

The proposed command submitted as a pull request does work, although I needed to change `/srv/images` ownership from `root:users` to `jack:users` (perm. 775) to get it to work and resolve a new error `Permission denied (os error 13)`. I did re-run the original and original plus switch commands with the new ownership but they failed as before with the same errors.

Let me know if you need anything else from me. This one has me scratching my head a little, especially the new error!

bgilbert:
Interesting: something inside the HTTP library is getting an `EPERM` when creating an HTTP client.  At that point apparently the library is mainly initializing TLS and the DNS resolver, and it's unusual that we're getting that error there.  At a guess, you're running podman with an old seccomp filter and it's rejecting a newer system call used by reqwest.  What's your host OS version?  Does running podman with `--security-opt=seccomp=unconfined` avoid the problem?

jackmonty:
I tried the following command:

```bash
podman run --pull=always --security-opt=seccomp=unconfined --rm -v /srv/images/:/data -w /data quay.io/coreos/coreos-installer:release download -s "stable" -p qemu -f qcow2.xz --decompress

Trying to pull quay.io/coreos/coreos-installer:release...
Getting image source signatures
Copying blob f949d798783d skipped: already exists  
Copying blob 76dc685b05ee skipped: already exists  
Copying blob de49e9a09291 skipped: already exists  
Copying blob a3ed95caeb02 done  
Writing manifest to image destination
Storing signatures
Downloading Fedora CoreOS stable x86_64 qemu image (qcow2.xz) and signature
Error: opening ./fedora-coreos-35.20220116.3.0-qemu.x86_64.qcow2

Caused by:
    Permission denied (os error 13)
```

Note the error has changed from `os error 1` to `os error 13`. I confirm that the ownership of the `/srv/images` directory remains `jack:user`, which is what I need to do as discussed [at the bottom of a previous post](https://github.com/coreos/fedora-coreos-docs/pull/351#issuecomment-1023286599).

bgilbert:
Okay, so it seems like seccomp is part of the problem, but not all of it.  Maybe SELinux also?  Let's try adding `--security-opt=label=disable`.

If that works, I'll ask you to run `strace` inside the container to see what the failing syscall is.

And again, what's your host OS version?

jackmonty:
That did it! Tried the following:

```bash
podman run --pull=always --security-opt=seccomp=unconfined --security-opt=label=disable --rm -v /srv/images/:/data -w /data quay.io/coreos/coreos-installer:release download -s "stable" -p qemu -f qcow2.xz --decompress

Trying to pull quay.io/coreos/coreos-installer:release...
Getting image source signatures
Copying blob f949d798783d skipped: already exists  
Copying blob 76dc685b05ee skipped: already exists  
Copying blob de49e9a09291 skipped: already exists  
Copying blob a3ed95caeb02 done  
Writing manifest to image destination
Storing signatures
Downloading Fedora CoreOS stable x86_64 qemu image (qcow2.xz) and signature
Read disk 10.8 MiB/638.2 MiB (1%)
...
Read disk 638.2 MiB/638.2 MiB (100%)
gpg: Signature made Tue Feb  1 17:17:16 2022 UTC
gpg:                using RSA key 787EA6AE1147EEE56C40B30CDB4639719867C58F
gpg: Good signature from "Fedora (35) <fedora-35-primary@fedoraproject.org>" [ultimate]
./fedora-coreos-35.20220116.3.0-qemu.x86_64.qcow2
```

Apologies, my host OS version is CentOS Linux 7 (Core). Thanks!

travier:
> Apologies, my host OS version is CentOS Linux 7 (Core). Thanks!

This is likely due to a new syscall used by the newer glibc in Fedora 35 that are not correctly listed in the seccomp filter on CentOS 7. I don't think this is going to be fixed at this point. The easiest way to fix this is to update to CentOS Stream 8.

bgilbert:
Yeah, agreed.  Perhaps we should add a note to docs saying that `--security-opt=seccomp=unconfined` is required on CentOS 7.

travier:
Yes, I would prefer we do that too.

jlebon:
@jackmonty Are you still interested in tweaking the documentation as suggested above?

PR #355, opened 2022-02-11
Rework live image docs

bgilbert:
Separate out bare-metal installation, live image quickstart instructions, and live image reference material into their own pages.  Document live image customization and various other details.  Do some cleanup.

Requires https://github.com/coreos/coreos-installer/pull/776.

Issue #299, opened 2021-05-28
Document automated reprovisioning for bare metal


bgilbert:
We want to encourage users to reprovision their bare-metal nodes when their config changes, but we don't document how to do it.  In the long run, https://github.com/coreos/fedora-coreos-tracker/issues/399 might be the right approach, but in the short term, other mechanisms are required.  This is arguably part of https://github.com/coreos/fedora-coreos-docs/issues/117, but I'm filing it separately to distinguish the higher-level workflow issue from the nuts-and-bolts topics in #117.

There are a few possible reprovisioning flows:

- UEFI + PXE: use `efibootmgr --bootnext` to switch the machine into netboot mode just long enough to reinstall the node.  (I haven't actually tried this.)
- PXE generally: set the machine to always netboot, but have the netboot server redirect the boot back to the local disk unless it's been told to reprovision this particular machine.  Does [Matchbox](https://github.com/poseidon/matchbox) help here?
- IPMI + virtual CD + ISO: use `coreos-installer iso kargs embed` to set `coreos.inst.install_dev=` and related kargs, then push the ISO to the system with IPMI.
- Just run from live ISO / PXE and never install to disk.  This requires some other mechanism for OS updates.

Whichever flows we decide to recommend, we should provide step-by-step instructions.

bgilbert:
We should also document how to preserve data volumes when reprovisioning.  The traditional model is to do this with Ignition, but there are some quirks and they're not documented in one place:

- `storage.disks`: specify the data partition(s); set `wipe_table: false` and `wipe_partition_entry: false`.  On secondary disks, Ignition will leave existing partitions alone.  On the boot disk, coreos-installer will wipe the partition table entries (but not the data) and then Ignition will recreate them with the same parameters as before.
- `storage.raid`: probably safest to avoid, as Ignition always recreates the RAID.  Recreating a RAID with the same parameters might preserve data, but I don't know if this has been tested.
- `storage.luks`: specify the data volumes; set `wipe_volume: false` and use a static `key_file`.  Ignition will reuse the existing LUKS volumes, or create new ones if the backing storage is blank.  Note that Ignition can't reuse LUKS volumes encrypted with ephemeral keys or Clevis.  The static key will be stored on the root filesystem, which can be encrypted with Clevis if desired.
- `storage.filesystems`: specify the filesystems and set `wipe_filesystem: false`.  Ignition will reuse existing filesystems, or create new ones if the backing storage is blank.

[...]

PR #359, opened 2022-02-14
bare-metal: add small section about reinstallation

jlebon:
And note the restrictions around the `boot` filesystem.

### DIFF
--- a/modules/ROOT/pages/bare-metal.adoc
+++ b/modules/ROOT/pages/bare-metal.adoc
@@ -165,3 +165,15 @@ Commonly bare metal systems will have a diversity of hardware - some systems may
 A useful approach is to script generating a per-machine `.iso`.  If you have a hardware database (whether a text file in git or relational database) then it will work to generate a per-machine `target-dell.ign` and `target-hp.ign` for example, and specify that to `--dest-ignition` alongside the appropriate `--dest-device` to generate `fedora-coreos-install-dell.iso` and `fedora-coreos-install-hp.iso`.
 
 Alternatively, instead of generating per-machine ISOs, you can have a `--pre-install` script run a privileged container which inspects the target system and writes an appropriate https://coreos.github.io/coreos-installer/customizing-install/#config-file-format[installer config] to `/etc/coreos/installer.d`.
+
+== Reinstalling Fedora CoreOS
+
+You can use any of the methods described above to reinstall Fedora CoreOS on the same machine via the live environment.
+
+=== Data persistence
+
+The installer does not completely scrub the target disk. Thanks to Ignition's https://coreos.github.io/ignition/operator-notes#filesystem-reuse-semantics[filesystem-reuse semantics], if you apply a config with the exact same partition and filesystem settings as the first install, all previously stored data in additional partitions will still be available. This can be useful for persisting e.g. `/var/lib/containers` or `/var/home` between reinstalls. There are some restrictions however. See the xref:live-booting.adoc#_using_persistent_state[Using persistent state] section for more information. You will also want to ensure the root filesystem is a fixed number at least 8 GiB as described in xref:storage.adoc[Configuring Storage] so that data partitions are not accidentally overwritten.
+
+=== Destination drive
+
+Fedora CoreOS requires that only a single filesystem labeled `boot` exists. If multiple such filesystems are found on the first boot, provisioning will error out as a fail-safe. If reinstalling Fedora CoreOS on a different disk than the previous installation, make sure to wipe the previous disk using e.g. `wipefs`. For example, if reinstalling to `/dev/sdb`, but `/dev/sda` still contains the previous installation of FCOS, use `wipefs -a /dev/sda` in the installation environment.


jlebon:
Requires: https://github.com/coreos/fedora-coreos-docs/pull/355

bgilbert:
xref #299

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -179,3 +179,15 @@ Commonly bare metal systems will have a diversity of hardware - some systems may
 A useful approach is to script generating a per-machine `.iso`.  If you have a hardware database (whether a text file in git or relational database) then it will work to generate a per-machine `target-dell.ign` and `target-hp.ign` for example, and specify that to `--dest-ignition` alongside the appropriate `--dest-device` to generate `fedora-coreos-install-dell.iso` and `fedora-coreos-install-hp.iso`.
 
 Alternatively, instead of generating per-machine ISOs, you can have a `--pre-install` script run a privileged container which inspects the target system and writes an appropriate https://coreos.github.io/coreos-installer/customizing-install/#config-file-format[installer config] to `/etc/coreos/installer.d`.
+
+== Re-installing Fedora CoreOS
+
+You can use any of the methods described above to re-install Fedora CoreOS on the same machine.
```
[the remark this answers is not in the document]
bgilbert: We should make it clear that you need to boot from install media to reinstall.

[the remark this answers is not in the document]
dustymabe: ahh yes. I was totally confused and thought this was related to https://github.com/coreos/fedora-coreos-tracker/issues/399 which is why I asked for docs over in https://github.com/coreos/fedora-coreos-tracker/issues/976#issuecomment-1039419566

jlebon:
Updated this! Not sure about the data persistence section now that we have https://docs.fedoraproject.org/en-US/fedora-coreos/live-booting/#_using_persistent_state. Though I guess it's worth calling out that the same concept also applies to non-live-booted setups.

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -165,3 +165,15 @@ Commonly bare metal systems will have a diversity of hardware - some systems may
...
 
 Alternatively, instead of generating per-machine ISOs, you can have a `--pre-install` script run a privileged container which inspects the target system and writes an appropriate https://coreos.github.io/coreos-installer/customizing-install/#config-file-format[installer config] to `/etc/coreos/installer.d`.
+
+== Re-installing Fedora CoreOS
+
+You can use any of the methods described above to re-install Fedora CoreOS on the same machine via the live environment.
+
+=== Data persistence
+
+The installer does not completely scrub the target disk. Thanks to Ignition's https://coreos.github.io/ignition/operator-notes/#filesystem-reuse-semantics[filesystem-reuse semantics], if you apply a config with the exact same partition and filesystem settings as the first install, all previously stored data will still be available. This can be useful for persisting e.g. `/var/lib/containers` or `/var/home` between re-installs. See the xref:live-booting.adoc#_using_persistent_state[Using persistent state] section for more information.
```
bgilbert, line 175: Due to the potential for data loss, we probably want to qualify this in a couple ways:

- This only works for data partitions, not for the rootfs.
- The data partitions must not specify `wipe_filesystem`, `wipe_volume`, or RAID volumes.

Those are mentioned in the linked doc, but we don't explicitly call out that there are limitations described at the other end of that link.  Also, the intro of that doc section describes a different goal than the one here.  So I'm wondering if we should do some restructuring: maybe we could pull out the common parts into a separate page/fragment and link/inline it in both places.

Should we also call out here that the root partition should be at least 8 GiB to prevent potentially clobbering data partitions on reinstall?

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -165,3 +165,15 @@ Commonly bare metal systems will have a diversity of hardware - some systems may
...
+
+=== Destination drive
+
+Fedora CoreOS expects only a single filesystem labeled `boot` to exist. If multiple such filesystems are found on the first boot, provisioning will error out as a fail-safe. If re-installing Fedora CoreOS on a different disk than the previous installation, make sure to wipe the previous disk using e.g. `wipefs`. For example, if re-installing to `/dev/sdb`, but `/dev/sda` still contains the previous installation of FCOS, use `wipefs -a /dev/sda` in the installation environment.
```
bgilbert, line 179: Maybe s/expects/requires/?

bgilbert (commented):
Yeah, I agree that we should be discussing data persistence on both pages.

modules/ROOT/pages/bare-metal.adoc
travier: We should also probably mention that it requires the keep partition logic for the rootfs disk?

bgilbert: In general it doesn't.  That logic is mainly for the case where the data partitions were created outside Ignition, and (1) might vary from system to system or (2) can't be expressed in an Ignition config (i.e. aren't aligned to a 1 MiB boundary).

There is one caveat though, which is that the partition-saving logic will physically prevent the new OS image from clobbering the data partitions, even if the rootfs minimum-size rule is violated.  That's a nice property, but I'm still hesitant to recommend partition-saving instead of recreating the partitions in the Ignition config, since the former encourages multiple sources of truth.

In any event, we do need docs for the partition-saving functionality.

jlebon: > Due to the potential for data loss, we probably want to qualify this in a couple ways:
> 
>     * This only works for data partitions, not for the rootfs.
> 
>     * The data partitions must not specify `wipe_filesystem`, `wipe_volume`, or RAID volumes.
> 
> 
> Those are mentioned in the linked doc, but we don't explicitly call out that there are limitations described at the other end of that link.

Updated to mention this!

> Should we also call out here that the root partition should be at least 8 GiB to prevent potentially clobbering data partitions on reinstall?

Good call, done!

travier (approved):

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -165,3 +165,15 @@ Commonly bare metal systems will have a diversity of hardware - some systems may
...
+
+== Re-installing Fedora CoreOS
+
+You can use any of the methods described above to re-install Fedora CoreOS on the same machine via the live environment.
+
+=== Data persistence
+
+The installer does not completely scrub the target disk. Thanks to Ignition's https://coreos.github.io/ignition/operator-notes/#filesystem-reuse-semantics[filesystem-reuse semantics], if you apply a config with the exact same partition and filesystem settings as the first install, all previously stored data in additional partitions will still be available. This can be useful for persisting e.g. `/var/lib/containers` or `/var/home` between re-installs. There are some restrictions however. See the xref:live-booting.adoc#_using_persistent_state[Using persistent state] section for more information. You will also want to ensure the root filesystem is a fixed number at least 8 GiB as described in xref:storage.adoc#[Configuring Storage] so that data partitions are not accidentally overwritten.
```
bgilbert, line 175: Looks like the first link isn't parsing correctly, and the last link has a spurious `#`.

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -165,3 +165,15 @@ Commonly bare metal systems will have a diversity of hardware - some systems may
 A useful approach is to script generating a per-machine `.iso`.  If you have a hardware database (whether a text file in git or relational database) then it will work to generate a per-machine `target-dell.ign` and `target-hp.ign` for example, and specify that to `--dest-ignition` alongside the appropriate `--dest-device` to generate `fedora-coreos-install-dell.iso` and `fedora-coreos-install-hp.iso`.
 
 Alternatively, instead of generating per-machine ISOs, you can have a `--pre-install` script run a privileged container which inspects the target system and writes an appropriate https://coreos.github.io/coreos-installer/customizing-install/#config-file-format[installer config] to `/etc/coreos/installer.d`.
+
+== Re-installing Fedora CoreOS
```
bgilbert, line 169: Nit: should these be `reinstall` instead of `re-install`?

bgilbert (approved):

modules/ROOT/pages/bare-metal.adoc
jlebon: Interestingly, that's the fully resolved URL. But agreed it looks funny.

bgilbert (approved):

Issue #353, opened 2022-02-04
Document new `/dev/disk/by-id/coreos-boot-disk` symlink


jlebon:
This was added in https://github.com/coreos/fedora-coreos-config/pull/1308 as part of https://github.com/coreos/fedora-coreos-tracker/issues/759. Would be good to have docs for this here and go over the configs we already have to use them so they're more platform independent.

dustymabe:
fixed in #362

PR #362, opened 2022-02-16
storage: document new '/dev/disk/by-id/coreos-boot-disk' symlink

nikita-dubrovskii:
Issue: https://github.com/coreos/fedora-coreos-docs/issues/353

### DIFF
--- a/modules/ROOT/pages/storage.adoc
+++ b/modules/ROOT/pages/storage.adoc
@@ -15,9 +15,9 @@ Currently, if the root filesystem is smaller than 8 GiB, a warning is emitted on
 
 Many of the examples below will reference a block device, such as `/dev/vda`.  The name of the available block devices depends on the underlying infrastructure (bare metal vs cloud), and often the specific instance type.  For example in AWS, some instance types have NVMe drives (`/dev/nvme*`), others use `/dev/xvda*`.
 
-You will need to look for relevant documentation for your infrastructure around this.
+If your disk configuration is simple and uses the same disk the OS was booted from then the `/dev/disk/by-id/coreos-boot-disk` link can be used to conveniently refer to that device. This link is only available during provisioning for the purpose of making it easy to refer to the same disk the OS was booted from.
 
-In many cases, the simplest thing is to boot a single machine with an Ignition configuration that just gives you SSH access, and inspect the block devices via e.g. the `lsblk` command.
+In cases where you need to access other disks, the simplest thing to do is to boot a single machine with an Ignition configuration that just gives you SSH access, and inspect the block devices via e.g. the `lsblk` command.
 
 For physical hardware, a good best practice is to reference devices via the `/dev/disk/by-id/` or `/dev/disk/by-path` links.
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/storage.adoc
```diff
@@ -15,9 +15,9 @@ Currently, if the root filesystem is smaller than 8 GiB, a warning is emitted on
 
 Many of the examples below will reference a block device, such as `/dev/vda`.  The name of the available block devices depends on the underlying infrastructure (bare metal vs cloud), and often the specific instance type.  For example in AWS, some instance types have NVMe drives (`/dev/nvme*`), others use `/dev/xvda*`.
 
-You will need to look for relevant documentation for your infrastructure around this.
+In most cases, the simplest way to refer the target block device is using the `/dev/disk/by-id/coreos-boot-disk` link, which is created automatically and available only during machine provisioning.
```
dustymabe, line 18: I think maybe we should add a little more about what this link is. Maybe something like:

```suggestion
If your disk configuration is simple and uses the same disk the OS was booted from then the `/dev/disk/by-id/coreos-boot-disk` link can be used to conveniently refer to that device. This link is only available during provisioning for the purpose of making it easy to refer to the same disk the OS was booted from.
```

modules/ROOT/pages/storage.adoc
```diff
@@ -15,9 +15,9 @@ Currently, if the root filesystem is smaller than 8 GiB, a warning is emitted on
...
 
-In many cases, the simplest thing is to boot a single machine with an Ignition configuration that just gives you SSH access, and inspect the block devices via e.g. the `lsblk` command.
+Otherwise you will need to look for relevant documentation for your infrastructure around this, and/or by booting a single machine with an Ignition configuration that just gives you SSH access, and inspect the block devices via e.g. the `lsblk` command.
```
dustymabe, line 20: If the above suggestion is accepted we'll want to do something like:

```suggestion
In cases where you need to access other disks, the simplest thing to do is to boot a single machine with an Ignition configuration that just gives you SSH access, and inspect the block devices via e.g. the `lsblk` command.
```

modules/ROOT/pages/storage.adoc
```diff
@@ -32,9 +32,9 @@ variant: fcos
 version: 1.4.0
 storage:
   disks:
-  - # The name of the primary block device. In virtio-based setups, this is
-    # likely `/dev/vda`. Elsewhere, it's likely `/dev/sda`.
-    device: /dev/vda
+  - # The link to the primary block device. In virtio-based setups, this points
+    # likely to `/dev/vda`, elsewhere, likely to `/dev/sda`.
```
dustymabe, line 36:
```suggestion
  - # The link to the block device the OS was booted from.
```

dustymabe (commented):
Thanks for writing this up. A few suggestions to further clarify what `/dev/disk/by-id/coreos-boot-disk` is.

dustymabe (approved):
LGTM

PR #364, opened 2022-02-16
Move migration help pages to the bottom

travier:
Those should be less used now thus move them at the bottom.


cgwalters (approved):
I'm good with migrating them there.

bgilbert (approved):

PR #366, opened 2022-02-17
live-reference: support TFTP in `coreos.live.rootfs_url`

bgilbert:
This is supported as of https://github.com/coreos/fedora-coreos-config/pull/1526 and allows us to recommend `rootfs_url` even without an HTTP(S) server.


bgilbert:
https://github.com/coreos/fedora-coreos-config/pull/1526 should reach stable the week of March 14.

jlebon (approved):

PR #368, opened 2022-02-19
update-streams: include arch selection

miabbott:


modules/ROOT/pages/update-streams.adoc
```diff
@@ -54,9 +54,11 @@ Switching between streams may introduce regressions or bugs due to skipping upda
 sudo systemctl stop zincati.service
 
 # Perform the rebase to a different stream
+# Supported architectures: aarch64, ppc64le, x86_64
```
bgilbert, line 57: We don't have ppc64le yet, right?

miabbott: Oops, got confused looking at the different refs on the remote

modules/ROOT/pages/update-streams.adoc
```diff
@@ -54,9 +54,11 @@ Switching between streams may introduce regressions or bugs due to skipping upda
 sudo systemctl stop zincati.service
 
 # Perform the rebase to a different stream
+# Supported architectures: aarch64, x86_64
 # Available streams: "stable", "testing", and "next"
+ARCH="$(arch)"
```
dustymabe, line 59:
```suggestion
ARCH="x86_64"
```

I feel like `$(arch)` is more confusing. We should either switch it to a real value or switch `STREAM="testing"` below to `STREAM=$(stream)`.

miabbott: `$(arch)` is the output of the `arch` command...`$(stream) is going to result in an error

dustymabe: oh wow. TIL about `/usr/bin/arch`. Sorry for the noise.

dustymabe (approved):
LGTM

PR #369, opened 2022-02-22
storage: add `with_mount_unit` to "Defining a filesystem" section

bgilbert:
This one example doesn't set `with_mount_unit`, so the resulting filesystem won't be mounted by default.  Fix this.


lucab (approved):

dustymabe:
Thanks @bgilbert

Issue #322, opened 2021-09-29
Unable to set user password following documentation


sandrobonazzola:
**Describe the bug**
Documentation for setting up a password for a user on coreos doesn't work for me.
I'm following: <https://docs.fedoraproject.org/en-US/fedora-coreos/authentication/#_using_password_authentication>.
It says:
```
To generate a secure password hash, use the mkpasswd command:

$ mkpasswd --method=yescrypt
Password:
$y$j9T$A0Y3wwVOKP69S.1K/zYGN.$S596l11UGH3XjN...
```

I'm on CentOS Stream 8 which I used for the bastion of an OKD deployment. I run:  `dnf provides "*/mkpasswd"` getting:

```
expect-5.45.4-5.el8.x86_64 : A program-script interaction and testing utility
Repo        : baseos
Matched from:
Filename    : /usr/bin/mkpasswd
```

Then I install it: `dnf install expect`. But running the suggested command:
```
mkpasswd --method=yescrypt
passwd: bad argument --method=yescrypt: unknown option
```

**Reproduction steps**
1.  see above description

**Expected behavior**
documentation should provide exact steps to be able to set the password for coreos users

**Actual behavior**
can't setup a password for the user given existing documentation

**Ignition config**
Please attach your FCC or Ignition config used to provision your system. Be sure to sanitize any private data. If not using FCCT to generate your Ignition config, does the Ignition config pass validation using [ignition-validate](https://coreos.github.io/ignition/getting-started/#config-validation)?

**Additional information**
Add any other information about the problem here.

sandrobonazzola:
Looks like the documented commands works on Fedora 34 with the following package:

```
$ rpm -qf `which mkpasswd`
mkpasswd-5.5.9-1.fc34.x86_64
```

travier:
Running that command from a Fedora container might be the easiest workaround.

bgilbert:
Created a `mkpasswd` container and documented it in https://github.com/coreos/fedora-coreos-docs/pull/370.

PR #370, opened 2022-02-22
authentication: use `mkpasswd` container

bgilbert:
RHEL 8 ships a different implementation of `mkpasswd` that doesn't support modern hashes.  For ease of use, just containerize `mkpasswd`.

Fixes https://github.com/coreos/fedora-coreos-docs/issues/322.

Corresponding upstream docs in https://github.com/coreos/butane/pull/327.


sandrobonazzola (approved):
LGTM ; verified working on CentOS Stream 8

lucab (approved):

PR #371, opened 2022-02-24
Add Major Changes page

travier:
Add new page to store documentation for major changes happening in
Fedora CoreOS.

Fixes https://github.com/coreos/fedora-coreos-docs/issues/314.

PR #372, opened 2022-02-25
build.sh: Fix overquoting

travier:
Split from https://github.com/coreos/fedora-coreos-docs/pull/371 to simplify review.

See https://pagure.io/fedora-docs/template/pull-request/21


jlebon (approved):
Nice work fixing the template as well!

PR #373, opened 2022-02-27
os-extensions: update typo in bash profile for $EDITOR override

dustymabe:
I think the intention here was to use `#!/bin/sh` but I don't think
this line is required so let's just delete it.


bgilbert (approved):

travier:
Most likely not needed indeed: https://src.fedoraproject.org/rpms/nano/blob/rawhide/f/nano-default-editor.sh

PR #376, opened 2022-03-24
vmware: provide instructions for modifying OVA

miabbott:
Users may want to modify the OVA to run on older versions of VMware products, so provide a simple example.

### DIFF
--- a/modules/ROOT/pages/provisioning-vmware.adoc
+++ b/modules/ROOT/pages/provisioning-vmware.adoc
@@ -2,7 +2,8 @@
 
 This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VMware hypervisor.
 
-NOTE: Only https://kb.vmware.com/s/article/1003746[hardware version] 13 or later, vSphere ESXi hosts version 6.5 or later and vCenter host version 6.5 or later are supported.
+NOTE: Fedora CoreOS supports VMware ESXi &ge; 6.5, VMware Workstation &ge; 16, and VMware Fusion &ge; 12. It may be possible to
+xref:provisioning-vmware.adoc#_modifying_ovf_metadata[modify the metadata of the OVF] to run in older VMware products, but compatibility and supportability cannot be guaranteed.
 
 == Prerequisites
 
@@ -144,7 +145,7 @@ For further information on first-boot networking, see https://coreos.github.io/a
 
 You may encounter problems with your Ignition configuration that require access to the system log which appears during first-boot. To make a copy of the system log you can attach a serial device to the VM before booting. vSphere as well as Workstation and Fusion allow this and will save the output to a file of your choice.
 
-To attach a serial device simply modify the hardware settings of the powered off VM and add a `Serial Port`. Select the destination and name of the file to be created. Afterwards power on the VM. When encountering an error, check the file you initially specified – it should contain a copy of the system log.
+To attach a serial device simply modify the hardware settings of the powered off VM and add a `Serial Port`. Select the destination and name of the file to be created. Afterwards power on the VM. When encountering an error, check the file you initially specified - it should contain a copy of the system log.
 
 The serial device can also be added to the VM via `govc` as described in the https://github.com/vmware/govmomi/blob/master/govc/USAGE.md#deviceserialconnect[official usage documentation]:
 
@@ -154,4 +155,33 @@ VM_NAME='fcos-node01'
 
 govc device.serial.add -vm "${VM_NAME}"
 govc device.serial.connect -vm "${VM_NAME}" "[datastore] ${VM_NAME}/console.log"
+----
+
+== Modifying OVF metadata
+
+NOTE: While we provide these instructions for modifying the OVF metadata, we cannot
+guarantee that any modifications to the OVF metadata will result in a usable
+guest VM.
+
+Fedora CoreOS is intended to run on
+https://lifecycle.vmware.com/[generally supported] releases of VMware ESXi,
+VMware Workstation, and VMware Fusion. Accordingly, the Fedora CoreOS VMware
+OVA image specifies a
+https://kb.vmware.com/s/article/1003746[virtual hardware version]
+that may not be compatible with older, unsupported VMware products.
+However, you can modify the image's OVF metadata to specify an older
+virtual hardware version.
+
+The VMware OVA is simply a tarball that contains the files `disk.vmdk` and
+`coreos.ovf`. In order to edit the metadata used by FCOS as a guest VM, you
+should untar the OVA artifact, edit the OVF file, then create a new OVA file.
+
+The example commands below change the OVF hardware version from the preconfigured
+value to hardware version `13`. (Note: the defaults in the OVF are subject to change.)
+
+[source,bash,subs="attributes"]
+----
+tar -xvf fedora-coreos-{stable-version}-vmware.x86_64.ova
+sed -iE 's/vmx-[0-9]*/vmx-13/' coreos.ovf
+tar -H posix -cvf fedora-coreos-{stable-version}-vmware-vmx-13.x86_64.ova coreos.ovf disk.vmdk
 ----


modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -2,7 +2,9 @@
 
 This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VMware hypervisor.
 
-NOTE: Only https://kb.vmware.com/s/article/1003746[hardware version] 13 or later, vSphere ESXi hosts version 6.5 or later and vCenter host version 6.5 or later are supported.
+NOTE: Only https://kb.vmware.com/s/article/1003746[hardware version] 13 or later, vSphere ESXi hosts version 6.5 or later and vCenter host version 6.5 or later are supported. It may be possible to
+xref:modifying-ovf-metadta.adoc[modify the metadata of the OVA] to run in older versions of vSphere or vCenter,
```
dustymabe, line 6: This mentions OVA, but the title of the other document we're adding mentions OVF.. As someone dumb to the vmwhare sphere, what's the difference?

dustymabe: Ahh I see a paragraph in the new doc about it "shipped as an OVA, which is simply a tarball that contains the files disk.vmdk and coreos.ovf"

dustymabe: in that case maybe we can say "modify the OVF metadata of the OVA" here.

modules/ROOT/pages/modifying-ovf-metadata.adoc
```diff
@@ -0,0 +1,26 @@
+= Modifying OVF metadata
```
dustymabe, line 1: Should we mention VMWare in this title?
```suggestion
= Modifying VMWare Disk OVF metadata
```

dustymabe (approved):
LGTM - a few suggestions

bgilbert:
I'm not 100% convinced that we should even document this.  It's a pretty low-level operation, and we support what we support.  If we do want to document it, maybe it'd be better to add it as a section in the VMware provisioning guide, rather than giving it its own page.

dustymabe:
> I'm not 100% convinced that we should even document this. It's a pretty low-level operation, and we support what we support. If we do want to document it, maybe it'd be better to add it as a section in the VMware provisioning guide, rather than giving it its own page.

Part of the need for this documentation was predicated on the assumption we were going to switch the defaults before EOL of those platforms. This was the general direction from the first community meeting discussion on the topic.

So, it's no longer as strictly needed specifically for hardware version, but since the docs are already written I'm inclined to let it merge.

Also, can the same flow be used to move back to BIOS if someone needs that for whatever reason? That could be a good reason to keep the docs.

bgilbert:
> So, it's no longer as strictly needed specifically for hardware version, but since the docs are already written I'm inclined to let it merge.

Yeah, I was suggesting moving it to the provisioning guide as a compromise of sorts.  Giving it a top-level page, and a listing in the sidebar, feels like an endorsement of sorts.

dustymabe:
> > So, it's no longer as strictly needed specifically for hardware version, but since the docs are already written I'm inclined to let it merge.
> 
> Yeah, I was suggesting moving it to the provisioning guide as a compromise of sorts. Giving it a top-level page, and a listing in the sidebar, feels like an endorsement of sorts.

Seems like a reasonable compromise.. Almost wish it could be a floating page that didn't live as part of the nav bar that we could just link to from the provisioning guide for VMWare.

modules/ROOT/nav.adoc
```diff
@@ -47,6 +47,7 @@
 ** xref:emergency-shell.adoc[Emergency Console Access]
 ** xref:debugging-with-toolbox.adoc[Debugging with Toolbx]
 ** xref:debugging-kernel-crashes.adoc[Debugging Kernel Crashes]
+** xref:modifying-ovf-metadata.adoc[Modifying OVF Metadata]
```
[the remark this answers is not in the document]
miabbott: Didn't even notice it, but definitely see it now that you point it out;  will fix it

miabbott:
> Part of the need for this documentation was predicated on the assumption we were going to switch the defaults before EOL of those platforms. This was the general direction from the first community meeting discussion on the topic.

Even if we switch defaults after EOL of those vSphere versions, there are going to be users that are not able to upgrade their platform to newer versions because of `$REASONS`.  I still think it is useful to provide some guidance on how to modify the images in those cases.

> Yeah, I was suggesting moving it to the provisioning guide as a compromise of sorts. Giving it a top-level page, and a listing in the sidebar, feels like an endorsement of sorts.

I'll move the instructions into a sub-heading in the VMware provisioning guide.

dustymabe (approved):
still LGTM

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -155,3 +156,31 @@ VM_NAME='fcos-node01'
 govc device.serial.add -vm "${VM_NAME}"
 govc device.serial.connect -vm "${VM_NAME}" "[datastore] ${VM_NAME}/console.log"
 ----
+
+== Modifying OVA metadata
```
bgilbert, line 160: Let's try to distinguish OVA and OVF throughout this section.  The metadata is OVF and the single-file archive is OVA.

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -155,3 +156,31 @@ VM_NAME='fcos-node01'
...
+
+NOTE: While we provide these instructions for modifying the OVA, we cannot
+guarantee that any modifications to the OVA metadata will result in a usable
+guest VM.
+
+Fedora CoreOS ships the VMware OVA with certain metadata defaults in the OVA file.
+The hardware version and default firmware are of particular interest, as they
+can affect how the guest VM operates in the hypervisor. These defaults may
+prevent the use of FCOS as a guest VM in older versions of VMware products and
```
bgilbert, line 169: Should we explicitly state a policy that we will support non-EOL versions of VMware products?

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -155,3 +156,31 @@ VM_NAME='fcos-node01'
...
+it may be necessary to modify the metadata in order to use FCOS as a guest VM.
+
+The VMware OVA is simply a tarball that contains the files `disk.vmdk` and
+`coreos.ovf`. In order to edit the metadata used by FCOS as a guest VM, you
+should untar the OVA artifact, edit the OVF file, then create a new OVA file.
```
bgilbert, line 174: In principle the OVA should also include a `.mf` file, which is a list of SHA digests for the other files.  We're not doing that because when I tried, I got mysterious hash mismatches I couldn't track down.  I don't think it's especially important, but wanted to note that these instructions assume the user doesn't need to update or remove an `.mf` file after changing the `.ovf`.

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -155,3 +156,31 @@ VM_NAME='fcos-node01'
...
+
+In the example below, the hardware version in the OVF is changed from the
+default value of `17` to a newer version of `13`. (Note: the defaults in the
```
bgilbert, line 177: Did 13 and 17 get swapped here?

modules/ROOT/pages/provisioning-vmware.adoc
miabbott: I think that is fair; though we don't have any regular testing on those platforms.

modules/ROOT/pages/provisioning-vmware.adoc
miabbott: Oops, good catch.

modules/ROOT/pages/provisioning-vmware.adoc
miabbott: Is this something we should discuss more broadly before that kind of statement?

modules/ROOT/pages/provisioning-vmware.adoc
miabbott: Actually, I am just going to update the wording to align with the example of changing from a newer default value to an older value.

modules/ROOT/pages/provisioning-vmware.adoc
bgilbert: > I think that is fair; though we don't have any regular testing on those platforms.

I meant in the limited sense that we won't bump the HW version to exclude non-EOL products.  While we're telling users how to change the defaults, it may be helpful to mention when they won't need to.

> Is this something we should discuss more broadly before making that kind of statement?

Probably.

miabbott:
@bgilbert added a blurb about supporting non-EOL versions; let me know if you think it is in the right place

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -155,3 +156,36 @@ VM_NAME='fcos-node01'
...
+Fedora CoreOS ships the VMware OVA with metadata defaults defined in the OVF
+file contained within the OVA file. The hardware version and default firmware
+are of particular interest, as they can affect how the guest VM operates in the
+hypervisor. These defaults may prevent the use of FCOS as a guest VM in older
+versions of VMware products and it may be necessary to modify the metadata in
+order to use FCOS as a guest VM.
+
+NOTE: The Fedora CoreOS project will not make any changes to the hardware
+version in the OVF metadata of the VMware OVA that would exclude non-EOL
+platforms from being supported.
```
bgilbert, line 175: Maybe something like this?

```suggestion
Fedora CoreOS is intended to run on
https://lifecycle.vmware.com/[generally supported] releases of VMware ESXi,
VMware Workstation, and VMware Fusion. Accordingly, the Fedora CoreOS VMware
OVA image specifies a
https://kb.vmware.com/s/article/1003746[virtual hardware version]
that may not be compatible with older, unsupported VMware products.
However, you can modify the image's OVF metadata to specify an older
virtual hardware version.
```

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -155,3 +156,36 @@ VM_NAME='fcos-node01'
...
+
+The VMware OVA is simply a tarball that contains the files `disk.vmdk` and
+`coreos.ovf`. In order to edit the metadata used by FCOS as a guest VM, you
+should untar the OVA artifact, edit the OVF file, then create a new OVA file.
+
+In the example below, the hardware version in the OVF is changed from the
```
bgilbert, line 181:
```suggestion
The example commands below change the OVF hardware version from the
```

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -155,3 +156,36 @@ VM_NAME='fcos-node01'
...
+default value of `17` to an older version of `13`. (Note: the defaults in the
+OVF are subject to change.)
+
+[source, bash]
+----
+tar -xvf fedora-coreos-35.20220213.3.0-vmware.x86_64.ova
+sed -i 's|vmx-17|vmx-13|' coreos.ovf
+tar -H posix -cvf fedora-coreos-35.20220213.3.0-vmware-vmx-13.x86_64.ova coreos.ovf disk.vmdk
+----
```
bgilbert, line 190: Let's avoid breaking user scripts when we change the default, and avoid hardcoding version numbers.

```suggestion
[source,bash,subs="attributes"]
----
tar -xvf fedora-coreos-{stable-version}-vmware.x86_64.ova
sed -iE 's/vmx-[0-9]*/vmx-13/' coreos.ovf
tar -H posix -cvf fedora-coreos-{stable-version}-vmware-vmx-13.x86_64.ova coreos.ovf disk.vmdk
----
```

miabbott:
Thanks for the feedback, Benjamin.  Incorporated your suggestions :arrow_up:

bgilbert (approved):
LGTM!

dustymabe:
without re-reviewing - you had my blessing a while ago so merge away if you're ready!

PR #377, opened 2022-03-24
platforms.adoc: add VirtualBox documentation

prestist:
Added documentation around virtualbox. Going through steps to add virtualbox to a supported type for our offical release as per https://github.com/coreos/fedora-coreos-tracker/issues/1008

### DIFF
--- a/modules/ROOT/pages/provisioning-virtualbox.adoc
+++ b/modules/ROOT/pages/provisioning-virtualbox.adoc
@@ -4,193 +4,144 @@ This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualB
 
 == Prerequisites
 
-Before provisioning an FCOS machine, you must have an Ignition configuration file containing your customizations and host it somewhere (e.g. using `python3 -m http.server`, we will assume this option is used in the following examples). If you do not already have an Ignition file, see xref:producing-ign.adoc[Producing an Ignition File].
+Before importing an FCOS machine, you must have an Ignition configuration file containing your customizations. If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File].
 
-== Setting up a new VM
+=== Downloading the OVA
 
-You can set up a VirtualBox virtual machine through the GUI or via the https://www.virtualbox.org/manual/UserManual.html#vboxmanage[`vboxmanage` CLI]. Below is a shell script that does most of what the graphical wizard will do. Note that it is tailored to the installation from live ISO or raw disk image as outlined in the following sections.
-
-NOTE: Please be aware that for now there is no dedicated Fedora CoreOS image available for VirtualBox and the installation instructions provided are essentially xref:bare-metal.adoc[bare metal instructions]. A dedicated image may be provided at a later time. Please check the GitHub issues https://github.com/coreos/fedora-coreos-tracker/issues/73[#73] and https://github.com/coreos/fedora-coreos-tracker/issues/144[#144] for more details.
-
-We will rely on the VirtualBox-provided NAT gateway to access the hosts' loopback interface on `10.0.2.2`. Check the https://www.virtualbox.org/manual/UserManual.html#network_nat[VirtualBox User Manual] for more details.
+Fedora CoreOS is designed to be updated automatically, with different schedules per stream.
+Once you have picked the relevant stream, you can download the latest OVA:
 
 [source, bash]
 ----
-VM_NAME='fcos-node01'
-VM_DISK=$(realpath ~/VirtualBox\ VMs/${VM_NAME})/${VM_NAME}.vdi
-
-vboxmanage createvm --name "${VM_NAME}" --ostype Fedora_64 --register
-# Set hardware settings for the VM, including recommended graphics controller
-vboxmanage modifyvm "${VM_NAME}" --memory 4096 --cpus 2 --vram 20 --graphicscontroller vmsvga --rtcuseutc on
-# Add a storage controller for hard disks
-vboxmanage storagectl "${VM_NAME}" --name SATA --add sata --controller IntelAhci --portcount 30 --bootable on
+STREAM="stable"
+mkdir ova-templates
+coreos-installer download -s "${STREAM}" -p virtualbox -f ova -C ./ova-templates/
 ----
 
-WARNING: VirtualBox and by extension `vboxmanage` always expect absolute file paths when specifying locations.
+Alternatively, OVA images can be manually downloaded from the https://getfedora.org/coreos/download?tab=metal_virtualized[download page].
 
-Make sure to create a port forwarding rule that allows access to the guest's SSH Port with the https://www.virtualbox.org/manual/UserManual.html#networkingmodes[default NAT setting] using e.g. `localhost:2222` as target address:
-[source, bash]
+== Booting a new VM on VirtualBox
+
+You can set up a VirtualBox virtual machine through the GUI or via the https://www.virtualbox.org/manual/UserManual.html#vboxmanage[`VBoxManage` CLI]. This guide will use the CLI for setting up the VM.
+
+=== Importing the OVA
+
+To import the OVA, use `VBoxManage import`:
+
+[source, bash, subs="attributes"]
 ----
-vboxmanage modifyvm "${VM_NAME}" --natpf1 "guestssh,tcp,,2222,,22"
-----
-
-The virtual machine is now ready and FCOS can be installed.
-
-== Installing from live ISO
-
-To install FCOS into a VirtualBox virtual machine using the live ISO, follow these steps:
-
-- Download the latest ISO image from the https://getfedora.org/coreos/download?tab=metal_virtualized&stream=stable[download page] or with podman (see https://coreos.github.io/coreos-installer/cmd/download/[documentation] for options):
-[source, bash]
-----
-podman run --security-opt label=disable --pull=always --rm -v .:/data -w /data \
-    quay.io/coreos/coreos-installer:release download -f iso
+VM_NAME=my-instance
+VBoxManage import --vsys 0 --vmname "$VM_NAME" fedora-coreos-{stable-version}-virtualbox.x86_64.ova
 ----
 
-- Create a sufficiently large disk and attach it to the previously created VM:
-[source, bash]
-----
-vboxmanage createmedium --filename "${VM_DISK}" --size 10240 --format VDI
-vboxmanage storageattach "${VM_NAME}" --storagectl SATA --type hdd --port 0 --device 0 --medium "${VM_DISK}"
-----
+=== Setting the Ignition config
 
-[NOTE]
-====
-*Fedora CoreOS requires the root filesystem to be at least 8 GiB.* For practical reasons, disk images for some platforms ship with a smaller root filesystem, which by default automatically expands to fill all available disk space. If you add additional partitions after the root filesystem, you must make sure to explicitly resize the root partition as shown so that it is at least 8 GiB.
-
-Currently, if the root filesystem is smaller than 8 GiB, a warning is emitted on login. Starting from June 2021, if the root filesystem is smaller than 8 GiB and is followed by another partition, Fedora CoreOS will refuse to boot. For more details, see https://github.com/coreos/fedora-coreos-tracker/issues/586[this bug].
-====
-
-- Attach a DVD drive and the live ISO to the VM:
-
-[source,bash,subs="attributes"]
-----
-FCOS_DISK=$(realpath ~/Downloads/fedora-coreos-{stable-version}-live.x86_64.iso)
-
-vboxmanage storagectl "$\{VM_NAME}" --name IDE --add ide --controller PIIX4 --hostiocache on --portcount 2 --bootable on
-vboxmanage storageattach "$\{VM_NAME}" --storagectl IDE --type dvddrive --port 1 --device 0 --medium "$\{FCOS_DISK}"
-----
-
-You can now boot the target system you have configured. The ISO is capable of bringing up a fully functioning FCOS system purely from memory (i.e. without using any disk storage). Once booted, you will have access to a bash command prompt.
-
-You can now run `coreos-installer` inside the VM:
-[source, bash]
-----
-sudo coreos-installer install /dev/sda \
-    --insecure-ignition --ignition-url http://10.0.2.2:8000/example.ign
-----
-
-TIP: Check out `coreos-installer install --help` for more options on how to install Fedora CoreOS.
-
-Once the installation is complete, you have to remove the live ISO after shutting down (`sudo shutdown -h now`) the VM. Once the ISO is removed, you can start the VM anew. The actual first boot process begins now. It is at this time that Ignition ingests the configuration file and provisions the system as specified.
+Ignition reads its configuration from the `/Ignition/Config` https://docs.oracle.com/en/virtualization/virtualbox/6.0/user/guestadd-guestprops.html[guest property] of the virtual machine. At present, guest properties can only be set from the host command line, and not via the GUI. To set the Ignition config for a VM:
 
 [source, bash]
 ----
-vboxmanage storageattach "${VM_NAME}" --storagectl IDE --type dvddrive --port 1 --device 0 --medium emptydrive
-vboxmanage startvm "${VM_NAME}" --type headless
-# Once the VM is available, you can log in:
-ssh core@localhost -p 2222
+IGN_PATH="/path/to/config.ign"
+VM_NAME=my-instance
+VBoxManage guestproperty set "$VM_NAME" /Ignition/Config "$(cat $IGN_PATH)"
 ----
 
-== Installing from raw bare metal image
+==== Ignition config size limitations
 
-WARNING: The steps below are provided for advanced users, may unexpectedly break in various places with newer Fedora CoreOS images, and require a Unix environment to work. If you encounter issues consider using the xref:_installing_from_live_iso[live ISO instructions] instead.
+The length of the `/Ignition/Config` guestinfo property is constrained by the maximum length of a command line on your host operating system. The OS-specific limits are approximately:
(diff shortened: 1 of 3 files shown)


modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -4,193 +4,57 @@ This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualB
...
-----
+== Booting a new VM on VirtualBox
 
-TIP: Check out `coreos-installer install --help` for more options on how to install Fedora CoreOS.
+This section shows how to use VirtualBox to configure and run VMs from thee CLI. Alternatively the graphical UI can be used as well.
 
-Once the installation is complete, you have to remove the live ISO after shutting down (`sudo shutdown -h now`) the VM. Once the ISO is removed, you can start the VM anew. The actual first boot process begins now. It is at this time that Ignition ingests the configuration file and provisions the system as specified.
+=== Importing the OVA
 
+You can set up a VirtualBox virtual machine through the GUI or via the https://www.virtualbox.org/manual/UserManual.html#vboxmanage[`vboxmanage` CLI]. Below is a shell script that does most of what the graphical wizard will do. Note that this is tailored to a ova import.
```
bgilbert, line 30: "Most of"?  What does the GUI do that `vboxmanage` doesn't?

At this point it's not really a script; let's just call it a command.

Since we only provide an OVA image, I think the last sentence doesn't really add anything?

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -4,193 +4,57 @@ This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualB
...
 [source, bash]
 ----
-vboxmanage storageattach "${VM_NAME}" --storagectl IDE --type dvddrive --port 1 --device 0 --medium emptydrive
-vboxmanage startvm "${VM_NAME}" --type headless
-# Once the VM is available, you can log in:
-ssh core@localhost -p 2222
-----
-
-== Installing from raw bare metal image
+OVA_PATH="SetThisWithYourPathToOVA"
```
bgilbert, line 33: We're only using this in one place, so let's drop the variable and cut the code block down to one line.

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -4,193 +4,57 @@ This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualB
...
-[source, bash]
-----
-TMP_FCOS_BOOT_VOLUME=$(realpath /tmp/fcos_boot_volume)
-mkdir "${TMP_FCOS_BOOT_VOLUME}"
-
-fdisk -l "${UNPACKED_DISK}"
-sudo mount -o loop,offset="${OFFSET}" "${UNPACKED_DISK}" "${TMP_FCOS_BOOT_VOLUME}"
-----
+Once imported make sure that the network adapter is set to the type you want. More then likely it will be set to NAT.
+This seems to be due to the networking settings falling back to default.
```
bgilbert, line 40: This isn't super-clear.  Let's say 1) the default network attachment is NAT, 2) you might want e.g. bridged networking so the VM is accessible on the local network, and 3) here's how to change it (is there a `vboxmanage` subcommand?).

The last sentence is more relevant to a tracker ticket than user-facing docs.

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -4,193 +4,57 @@ This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualB
...
-vboxmanage createvm --name "${VM_NAME}" --ostype Fedora_64 --register
-# Set hardware settings for the VM, including recommended graphics controller
-vboxmanage modifyvm "${VM_NAME}" --memory 4096 --cpus 2 --vram 20 --graphicscontroller vmsvga --rtcuseutc on
-# Add a storage controller for hard disks
-vboxmanage storagectl "${VM_NAME}" --name SATA --add sata --controller IntelAhci --portcount 30 --bootable on
-----
-
-WARNING: VirtualBox and by extension `vboxmanage` always expect absolute file paths when specifying locations.
-
-Make sure to create a port forwarding rule that allows access to the guest's SSH Port with the https://www.virtualbox.org/manual/UserManual.html#networkingmodes[default NAT setting] using e.g. `localhost:2222` as target address:
```
bgilbert, line 31: Maybe we should keep this around as an alternative to switching to bridging?

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -4,193 +4,57 @@ This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualB
...
- ignition.firstboot ignition.config.url=http://10.0.2.2:8000/example.ign
-initrd ($root)/ostree/fedora-coreos- #...
-----
-
-The actual first boot process begins now. It is at this time that Ignition ingests the configuration file and provisions the system as specified.
-
-Once the VM finished booting, you can log in from your host:
-[source, bash]
-----
-ssh core@localhost -p 2222
```
bgilbert, line 179: If we keep the NAT port forwarding case, we should probably keep this too as a hint.

modules/ROOT/pages/provisioning-virtualbox.adoc
prestist: Admittedly I added 'most of' from my own ignorance. I could find documentation on how to set some guest properties without the CLI. I kept coming across documentation that pointed me back to CLI. https://docs.oracle.com/en/virtualization/virtualbox/6.0/user/vboxmanage-guestproperty.html . While I assumed I was not finding the correct documentation / googlefooing I figured I would cover my bases with 'most of'.

That is vary fair, will do.

bgilbert:
Requires https://github.com/coreos/coreos-assembler/pull/2489 to land in a stable release.

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -4,193 +4,144 @@ This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualB
 
 == Prerequisites
 
-Before provisioning an FCOS machine, you must have an Ignition configuration file containing your customizations and host it somewhere (e.g. using `python3 -m http.server`, we will assume this option is used in the following examples). If you do not already have an Ignition file, see xref:producing-ign.adoc[Producing an Ignition File].
+Before importing an FCOS machine, you must have an Ignition configuration file containing your customizations. If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File].
 
-== Setting up a new VM
+=== Downloading the OVA
 
-You can set up a VirtualBox virtual machine through the GUI or via the https://www.virtualbox.org/manual/UserManual.html#vboxmanage[`vboxmanage` CLI]. Below is a shell script that does most of what the graphical wizard will do. Note that it is tailored to the installation from live ISO or raw disk image as outlined in the following sections.
```
bgilbert, line 11: nit: `VBoxManage` for consistency

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -4,193 +4,144 @@ This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualB
...
 ----
 
-[NOTE]
-====
-*Fedora CoreOS requires the root filesystem to be at least 8 GiB.* For practical reasons, disk images for some platforms ship with a smaller root filesystem, which by default automatically expands to fill all available disk space. If you add additional partitions after the root filesystem, you must make sure to explicitly resize the root partition as shown so that it is at least 8 GiB.
+==== Ignition config size limitations
 
-Currently, if the root filesystem is smaller than 8 GiB, a warning is emitted on login. Starting from June 2021, if the root filesystem is smaller than 8 GiB and is followed by another partition, Fedora CoreOS will refuse to boot. For more details, see https://github.com/coreos/fedora-coreos-tracker/issues/586[this bug].
-====
+The length of the `/Ignition/Config` guestinfo property is constrained by the maximum length of a command line in your host operating system. The OS-specific limits are approximately:
```
bgilbert, line 50:
```suggestion
The length of the `/Ignition/Config` guestinfo property is constrained by the maximum length of a command line on your host operating system. The OS-specific limits are approximately:
```

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -4,193 +4,144 @@ This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualB
...
-    --insecure-ignition --ignition-url http://10.0.2.2:8000/example.ign
+variant: fcos
+version: 1.4.0
+ignition:
+  config:
+    replace:
+      source: https://example.com/config.ign
 ----
+. Use xref:producing-ign.adoc[Butane] to convert the Butane config to an Ignition config.
+. Set the `/Ignition/Config` guest property to the contents of the pointer Ignition config, following the instructions in <_setting_the_ignition_config>.
```
bgilbert, line 80: Ah, [looks](https://docs.asciidoctor.org/asciidoc/latest/macros/xref/#internal-cross-references) like the xref should use double angle brackets.

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -4,193 +4,144 @@ This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualB
...
-Once the installation is complete, you have to remove the live ISO after shutting down (`sudo shutdown -h now`) the VM. Once the ISO is removed, you can start the VM anew. The actual first boot process begins now. It is at this time that Ignition ingests the configuration file and provisions the system as specified.
-
-[source, bash]
-----
-vboxmanage storageattach "${VM_NAME}" --storagectl IDE --type dvddrive --port 1 --device 0 --medium emptydrive
-vboxmanage startvm "${VM_NAME}" --type headless
-# Once the VM is available, you can log in:
-ssh core@localhost -p 2222
-----
+=== Configuring networking
```
bgilbert, line 82: We should update this section for https://github.com/coreos/coreos-assembler/pull/2489#discussion_r845659754, or possibly drop it.

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -23,7 +23,7 @@ mkdir ova-templates
 coreos-installer download -s "${STREAM}" -p vmware -f ova -C ./ova-templates/
 ----
 
-Alternatively, OVA images can be manually downloaded from the https://getfedora.org/coreos/download/[download page].
+Alternatively, OVA images can be manually downloaded from the https://getfedora.org/coreos/download?tab=metal_virtualized[download page]
```
bgilbert, line 26: Trailing period got dropped.

Commit message nit: maybe something like `provisioning-vmware: point to correct tab on download page`?

bgilbert (commented):
LGTM generally.  I paired with @prestist on this, though, so shouldn't approve it.

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -4,193 +4,144 @@ This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualB
...
+To import the OVA, use `VBoxManage import`:
 
-- Download the latest ISO image from the https://getfedora.org/coreos/download?tab=metal_virtualized&stream=stable[download page] or with podman (see https://coreos.github.io/coreos-installer/cmd/download/[documentation] for options):
-[source, bash]
+[source, bash, subs="attributes"]
 ----
-podman run --security-opt label=disable --pull=always --rm -v .:/data -w /data \
-    quay.io/coreos/coreos-installer:release download -f iso
+VM_NAME=my-instance
+VBoxManage import --vsys 0 --vmname "$VM_NAME" fedora-coreos-{stable-version}-virtualbox.x86_64.ova
```
dustymabe, line 34: I'm hesitant to use `stable-version` here.. We have a few other pages that do it something like this:

```suggestion
FCOS_VERSION='...'
VM_NAME=my-instance
VBoxManage import --vsys 0 --vmname "$VM_NAME" "fedora-coreos-${FCOS_VERSION}-virtualbox.x86_64.ova"
```

modules/ROOT/pages/provisioning-virtualbox.adoc
```diff
@@ -4,193 +4,144 @@ This guide shows how to provision new Fedora CoreOS (FCOS) nodes on the VirtualB
...
-VM_DISK=$(realpath ~/VirtualBox\ VMs/${VM_NAME})/${VM_NAME}.vdi
-
-vboxmanage createvm --name "${VM_NAME}" --ostype Fedora_64 --register
-# Set hardware settings for the VM, including recommended graphics controller
-vboxmanage modifyvm "${VM_NAME}" --memory 4096 --cpus 2 --vram 20 --graphicscontroller vmsvga --rtcuseutc on
-# Add a storage controller for hard disks
-vboxmanage storagectl "${VM_NAME}" --name SATA --add sata --controller IntelAhci --portcount 30 --bootable on
+STREAM="stable"
+mkdir ova-templates
+coreos-installer download -s "${STREAM}" -p virtualbox -f ova -C ./ova-templates/
```
dustymabe, line 18: why the `mkdir ova-templates` followed by `-C ./ova-templates`. IIUC we don't use that anywere else in the doc here and we don't do it like that on any other page that I know of.

dustymabe (commented):
A few suggestions. Though the most valuable thing to do would be to have someone grab one of these new artifacts and run a test. IIUC @ravanelli has a mac with VBOX set up so maybe she would be willing.

dustymabe (approved):

modules/ROOT/pages/provisioning-virtualbox.adoc
bgilbert: > I'm hesitant to use `stable-version` here

Hmm, why not?

modules/ROOT/pages/provisioning-virtualbox.adoc
bgilbert: Turns out that bridged doesn't actually work: https://github.com/coreos/coreos-assembler/pull/2489#discussion_r846136796

modules/ROOT/pages/provisioning-virtualbox.adoc
dustymabe: partially because we have 3 streams and could be using `testing` or `next` here instead.

dustymabe: yes it is. Actually I wasn't aware of its use and prevalence in other pages.

oh hmm. I see - `{stable-version}` gets replaced on compile with https://github.com/coreos/fedora-coreos-docs/blob/0790bcddf55a3c3d2a1b109442879cb211ee7183/antora.yml#L11

I had missed that.

No reason to hold up this PR on that change now that I understand it a bit better. I do think that hardcoding in the version can lead to some confusion. Might be nice if a user had a slider or option to select the stream and then it would update the docs page.

modules/ROOT/pages/provisioning-virtualbox.adoc
prestist: We did use it in the vmwear documentation.[ here ](https://github.com/coreos/fedora-coreos-docs/blob/main/modules/ROOT/pages/provisioning-vmware.adoc)

modules/ROOT/pages/provisioning-virtualbox.adoc
prestist: Ah okay resolved I think?

modules/ROOT/pages/provisioning-virtualbox.adoc
prestist: Easy.

modules/ROOT/pages/provisioning-virtualbox.adoc
prestist: Nice catch!

modules/ROOT/pages/provisioning-virtualbox.adoc
prestist: This sounds like a larger ongoing conversation. Going to resolve this and recommend revising once the larger conversation happens.

modules/ROOT/pages/provisioning-virtualbox.adoc
dustymabe: and looking at that page we don't use it there either.. I really don't understand why we need to create the directory. Can we remove it from both pages (maybe in a 3rd commit)?

prestist: Yeah we certainly can. Ill do that here shortly.

prestist: Okay I updated it please take a look and let me know if its sufficient.

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -19,11 +19,10 @@ Once you have picked the relevant stream, you can download the latest OVA:
 [source, bash]
 ----
 STREAM="stable"
-mkdir ova-templates
-coreos-installer download -s "${STREAM}" -p vmware -f ova -C ./ova-templates/
+coreos-installer download -s "${STREAM}" -p vmware -f ova -C ./
```
dustymabe, line 22:
```suggestion
coreos-installer download -s "${STREAM}" -p vmware -f ova
```

`-C ./` is the default.

modules/ROOT/pages/provisioning-virtualbox.adoc
dustymabe: @bgilbert RFE in https://github.com/coreos/fedora-coreos-docs/issues/395

dustymabe (approved):

bgilbert:
LGTM!

PR #378, opened 2022-03-26
Move asciidoc attributes to antora.yml

travier:
This should fix the substitution on the Fedora built & hosted version of the docs.


travier:
This works for the Kinoite docs. Merging to give it a try.

bgilbert:
Thanks!

PR #379, opened 2022-03-26
Use partials to define attributes

travier:
Alternative to https://github.com/coreos/fedora-coreos-docs/pull/378.

Needs an update to the CI script. Might not be worth the effort.


bgilbert:
Yeah, this seems a bit heavyweight.  What's the upside vs. #378?

travier:
This would enable us to keep the configuration and the comments as and let us keep a single file dedicated for versions.

travier:
Let's close this one for now. We will revisit later if needed.

PR #388, opened 2022-04-05
Add page for enabling kdump through ignition

mike-nguyen:
As a part of https://github.com/coreos/fedora-coreos-tracker/issues/1147, add a page for enabling kdump through ignition.

### DIFF
--- a/modules/ROOT/pages/debugging-kernel-crashes.adoc
+++ b/modules/ROOT/pages/debugging-kernel-crashes.adoc
@@ -1,14 +1,35 @@
 = Debugging kernel crashes using kdump
 
-. Memory must be reserved for the crash kernel during booting of the first kernel. `crashkernel=auto` is likely insufficient memory on Fedora CoreOS. It is recommended to start testing with 300M. Kernel arguments can be provided like this:
+== Introduction
+kdump is a service that creates crash dumps when there is a kernel crash. It uses kexec to boot into a secondary kernel (known as a capture kernel), then exports the contents of the kernel's memory (known as a crash dump or vmcore) to the filesystem. The contents of vmcore can then be analyzed to root cause the kernel crash.
+
+Configuring kdump requires setting the `crashkernel` kernel argument and enabling the kdump systemd service. Memory must be reserved for the crash kernel during booting of the first kernel. `crashkernel=auto` generally doesn't reserve enough memory on Fedora CoreOS, so it is recommended to specify `crashkernel=300M`. 
+
+By default, the vmcore will be saved in `/var/crash`. It is also possible to write the dump to some other location on the local system or to send it over the network by editing `/etc/kdump.conf`. For additional information, see https://www.mankier.com/5/kdump.conf[`kdump.conf(5)`] and the comments in `/etc/kdump.conf` and `/etc/sysconfig/kdump`.
+
+== Configuring kdump via Ignition
+.Example kdump configuration
+[source,yaml]
+----
+variant: fcos
+version: 1.4.0
+kernel_arguments:
+  should_exist:
+  - 'crashkernel=300M'
+systemd:
+  units:
+  - name: kdump.service
+    enabled: true
+----
+
+== Configuring kdump after initial provision
+. Set the crashkernel kernel argument
 +
 [source, bash]
 ----
 sudo rpm-ostree kargs --append='crashkernel=300M'
 ----
 xref:kernel-args.adoc[More information] on how to modify kargs via `rpm-ostree`.
-
-. By default, the path in which the vmcore will be saved is `/var/crash`. It is also possible to write the dump over the network or to some other location on the local system by editing `/etc/kdump.conf`. For additional information, see https://www.mankier.com/5/kdump.conf[`kdump.conf(5)`] and the comments in `/etc/kdump.conf` and `/etc/sysconfig/kdump`.
 
 . Enable the kdump systemd service.
 +


dustymabe:
Thanks for writing this up @mike-nguyen.

It seems to me like this page has a lot of overlap with [the other page](). Maybe we should put the meat of the steps in this new page you've created (i.e. in the last section we show the `rpm-ostree kargs` that's currently documented on the other page. Then we could link to it from the other page and also maybe make the other page more about what you do with a vmcore once you have it.

modules/ROOT/pages/sysconfig-configure-kdump.adoc
```diff
@@ -0,0 +1,51 @@
...
+        ExecStart=/bin/touch /var/lib/%N.stamp
+        ExecStart=/bin/systemctl reboot
+
+        [Install]
+        WantedBy=multi-user.target
+----
+
+[NOTE]
+====
+The reboot in the systemd unit is required for kdump to use the correct kernel arguments when it kexecs into the crash kernel.
```
dustymabe, line 46:
```suggestion
The reboot in the systemd unit is required for kdump to use the correct kernel arguments. When kexec runs it re-uses the kernel arguments from the current boot and this has the undesired side effect of re-running Ignition if the machine had never rebooted.
```

I'm thinking we either give a little more context here, or we link to an issue where this is explained a little further.

modules/ROOT/pages/sysconfig-configure-kdump.adoc
```diff
@@ -0,0 +1,51 @@
...
+        RemainAfterExit=yes
+        ExecStart=/bin/systemctl enable kdump.service
+        ExecStart=/bin/touch /var/lib/%N.stamp
+        ExecStart=/bin/systemctl reboot
+
+        [Install]
+        WantedBy=multi-user.target
+----
+
+[NOTE]
```
dustymabe, line 44: might be nice to mention somewhere in here the info from the other page about `/etc/kdump.conf`

dustymabe:
> It seems to me like this page has a lot of overlap with the other page. Maybe we should put the meat of the steps in this new page you've created (i.e. in the last section we show the `rpm-ostree kargs` that's currently documented on the other page. Then we could link to it from the other page and also maybe make the other page more about what you do with a vmcore once you have it.

I guess another option is we don't create a new page and just add the "how to configure this via Ignition" to the existing page.

modules/ROOT/pages/sysconfig-configure-kdump.adoc
```diff
@@ -0,0 +1,51 @@
+:experimental:
```
travier, line 1: Not needed anymore

modules/ROOT/pages/sysconfig-configure-kdump.adoc
travier: I think this note should be at the top before the Ignition config.

travier (commented):
Agree we should probably merge this with https://docs.fedoraproject.org/en-US/fedora-coreos/debugging-kernel-crashes/

bgilbert:
This is incurring two reboots during first boot.  Don't we want something more like this:

```yaml
variant: fcos
version: 1.4.0
kernel_arguments:
  should_exist:
  - 'crashkernel=300M'
systemd:
  units:
    - name: kdump.service
      enabled: true
```

coupled with a `postprocess` script in the fedora-coreos-config `user-experience.yaml` that updates `KDUMP_COMMANDLINE_REMOVE` in `/etc/sysconfig/kdump`?

dustymabe:
> This is incurring two reboots during first boot. Don't we want something more like this:
> 
> ```yaml
> variant: fcos
> version: 1.4.0
> kernel_arguments:
>   should_exist:
>   - 'crashkernel=300M'
> systemd:
>   units:
>     - name: kdump.service
>       enabled: true
> ```
> 
> coupled with a `postprocess` script in the fedora-coreos-config `user-experience.yaml` that updates `KDUMP_COMMANDLINE_REMOVE` in `/etc/sysconfig/kdump`?

I know @mike-nguyen had looked into trying to get something like this to work but I don't know if he ever quite got it. @mike-nguyen do you remember what the complications were?

travier:
It's https://github.com/coreos/fedora-coreos-docs/pull/388#discussion_r845334608. We can not do it first boot and need an additional reboot.

bgilbert:
> It's [#388 (comment)](https://github.com/coreos/fedora-coreos-docs/pull/388#discussion_r845334608). We can not do it first boot and need an additional reboot. The note needs to be front and center to prevent folks from falling into this trap.

What's wrong with the approach in https://github.com/coreos/fedora-coreos-docs/pull/388#issuecomment-1092379772 though?  Ignition will reboot to add the `crashkernel` karg, and an appropriate value for `KDUMP_COMMANDLINE_REMOVE` should ensure the crash kernel doesn't inherit `ignition.firstboot`.

travier:
> What's wrong with the approach in [#388 (comment)](https://github.com/coreos/fedora-coreos-docs/pull/388#issuecomment-1092379772) though? Ignition will reboot to add the `crashkernel` karg, and an appropriate value for `KDUMP_COMMANDLINE_REMOVE` should ensure the crash kernel doesn't inherit `ignition.firstboot`.

Whoops sorry, indeed, I had missed that.

dustymabe:
> I know I tried adding `ignition.firstboot` to `KDUMP_COMMANDLINE_REMOVE` in `/etc/kdump.conf`--this was in the context of after the system already being provisioned through ignition. Its been a while and I forgot why that didn't work. I'll also try making it a part of the post process scripts and will report back.

Seems to work from what I can tell: https://github.com/coreos/fedora-coreos-config/pull/1708

mike-nguyen:
> > I know I tried adding `ignition.firstboot` to `KDUMP_COMMANDLINE_REMOVE` in `/etc/kdump.conf`--this was in the context of after the system already being provisioned through ignition. Its been a while and I forgot why that didn't work. I'll also try making it a part of the post process scripts and will report back.
> 
> Seems to work from what I can tell: [coreos/fedora-coreos-config#1708](https://github.com/coreos/fedora-coreos-config/pull/1708)

Sorry this fell off my radar.  I tested your PR manually and it worked!

mike-nguyen:
Consolidated everything into the original debugging kdump page and added steps to enable via ignition.  This doc is dependent on https://github.com/coreos/fedora-coreos-config/pull/1708 (thanks @dustymabe!) which will remove `ignition.firstboot` from the kernel arguments for the crash kernel.  A reboot is will no longer be required.

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -1,15 +1,36 @@
 = Debugging kernel crashes using kdump
 
-. Memory must be reserved for the crash kernel during booting of the first kernel. `crashkernel=auto` is likely insufficient memory on Fedora CoreOS. It is recommended to start testing with 300M. Kernel arguments can be provided like this:
+== Introduction
+kdump is a service that creates crash dumps when there is a kernel crash. It uses kexec to boot into a secondary kernel (known as a capture kernel), then exports the contents of the kernel's memory (known as a crash dump or vmcore) to the filesystem. The contents of vmcore can then be analyzed to root cause the kernel crash.
+
+Configuring kdump requires setting the `crashkernel` kernel argument enabling the kdump systemd service. Memory must be reserved for the crash kernel during booting of the first kernel. `crashkernel=auto` is likely insufficient memory on Fedora CoreOS. It is recommended to start testing with 300M.
```
bgilbert, line 6:
```suggestion
Configuring kdump requires setting the `crashkernel` kernel argument and enabling the kdump systemd service. Memory must be reserved for the crash kernel during booting of the primary kernel. `crashkernel=auto` generally doesn't reserve enough memory on Fedora CoreOS, so it is recommended to specify `crashkernel=300M`. 
```

modules/ROOT/pages/debugging-kernel-crashes.adoc
```diff
@@ -1,15 +1,36 @@
...
+== Introduction
+kdump is a service that creates crash dumps when there is a kernel crash. It uses kexec to boot into a secondary kernel (known as a capture kernel), then exports the contents of the kernel's memory (known as a crash dump or vmcore) to the filesystem. The contents of vmcore can then be analyzed to root cause the kernel crash.
+
+Configuring kdump requires setting the `crashkernel` kernel argument enabling the kdump systemd service. Memory must be reserved for the crash kernel during booting of the first kernel. `crashkernel=auto` is likely insufficient memory on Fedora CoreOS. It is recommended to start testing with 300M. 
+
+By default, the path in which the vmcore will be saved is `/var/crash`. It is also possible to write the dump over the network or to some other location on the local system by editing `/etc/kdump.conf`. For additional information, see https://www.mankier.com/5/kdump.conf[`kdump.conf(5)`] and the comments in `/etc/kdump.conf` and `/etc/sysconfig/kdump`.
```
bgilbert, line 8:
```suggestion
By default, the vmcore will be saved in `/var/crash`. It is also possible to write the dump to some other location on the local system or to send it over the network by editing `/etc/kdump.conf`. For additional information, see https://www.mankier.com/5/kdump.conf[`kdump.conf(5)`] and the comments in `/etc/kdump.conf` and `/etc/sysconfig/kdump`.
```

PR #262, opened 2021-03-24
Added param for wait for finish

kicaj:

PR #389, opened 2022-04-10
No "--wait" option in doctl compute image create

alcir:
There is no "--wait" option in `doctl compute image create`


bgilbert:
Yup, looks like you're right.  In that case, we should fully revert #262.  Want to update this PR to be a `git revert` of 993137fc1d706a5dd6560e131c1c0ff053e8616a?

cc @kicaj in case we're missing something.

kicaj:
You're right, the `--wait` parameter is for `droplet` not for `image`.

bgilbert:
Thanks for the report.  Closing in favor of the revert in https://github.com/coreos/fedora-coreos-docs/pull/392.

PR #394, opened 2022-04-13
Document how to use KubeVirt

rmohr:
Document how to get starting with FCOS on kubevirt.

xref https://github.com/coreos/fedora-coreos-tracker/issues/1126

https://pagure.io/fedora-web/websites/pull-request/241 will add KubeVirt to the getfedora page.

### DIFF
--- a/modules/ROOT/pages/provisioning-kubevirt.adoc
+++ b/modules/ROOT/pages/provisioning-kubevirt.adoc
@@ -75,3 +75,13 @@ username is `core`.
 ----
 kubectl virt ssh core@my-fcos
 ----
+
+== Mirroring the image for use in private registries
+
+If a private registry in air-gaped installations is used, the image can be mirrored to that registry.
+
+.Mirroring a fcos image from stable stream
+[source, bash]
+----
+skopeo copy docker://quay.io/coreos/fedora-coreos:stable docker://myregistry.io/myorg/fedora-coreos:stable
+----
--- a/modules/ROOT/pages/provisioning-kubevirt.adoc
+++ b/modules/ROOT/pages/provisioning-kubevirt.adoc
@@ -1,0 +1,77 @@
+= Provisioning Fedora CoreOS on KubeVirt
+
+This guide shows how to provision new Fedora CoreOS (FCOS) nodes on any KubeVirt-enabled Kubernetes cluster.
+
+== Prerequisites
+
+Before provisioning an FCOS machine, you must have an Ignition configuration file containing your customizations. If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File].
+
+You also need to have access to a Kubernetes environment with https://kubevirt.io/user-guide/operations/installation/[KubeVirt] installed.
+
+== Referencing the KubeVirt Image
+
+Fedora CoreOS is designed to be updated automatically, with different schedules per stream.
+
+The image for each stream can directly be referenced from the official registry:
+
+----
+quay.io/coreos/fedora-coreos:{stable|next|testing}
+----
+
+== Launching a VM instance
+
+Now that you have an image it can be used in your VMI definitions. The following example will use the image from the stable stream.
+
+You'll also need the Ignition config.
+Here the Ignition config is exposed to the VMI via a Kubernetes Secret.
+Learn about various ways to expose userdata to VMIs in the https://kubevirt.io/user-guide/virtual_machines/startup_scripts/#startup-scripts[KubeVirt user guide].
+
+.Creating the secret
+[source, bash]
+----
+kubectl create secret generic ignition-payload --from-file=userdata=example.ign
+----
+
+.Launching a VM instance referencing the secret
+[source, bash]
+----
+cat << END > vmi.yaml
+---
+apiVersion: kubevirt.io/v1
+kind: VirtualMachineInstance
+metadata:
+  name: my-fcos
+spec:
+  domain:
+    devices:
+      disks:
+      - disk:
+          bus: virtio
+        name: containerdisk
+      - disk:
+          bus: virtio
+        name: cloudinitdisk
+      rng: {}
+    resources:
+      requests:
+        memory: 1024M
+  volumes:
+  - containerDisk:
+      image: quay.io/coreos/fedora-coreos:stable
+    name: containerdisk
+  - name: cloudinitdisk
+    cloudInitConfigDrive:
+      secretRef:
+        name: ignition-payload
+END
+kubectl create -f vmi.yaml
+----
+
+Now you should be able to SSH into the instance. If you didn't change the defaults, the
+username is `core`.
+
+.Accessing the VM instance via ssh
+[source, bash]
+----
+kubectl virt ssh core@my-fcos
+----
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/provisioning-kubevirt.adoc
```diff
@@ -0,0 +1,96 @@
...
+            "users": [
+              {
+                "name": "core",
+                "sshAuthorizedKeys": [
+                  "ssh-rsa AAAAB3NzaC1yc2 [...]"
+                ]
+              }
+            ]
+          }
+        }
```
bgilbert, line 87: Rather than expecting users to figure out how to inline the Ignition config, let's give them an example.  Since we're already using a here-doc with `cat`, this might be something like:

```suggestion
$(cat config.ign | sed 's/^/        /')
```

though that's a bit awkward, since it hardcodes the indentation level.

modules/ROOT/pages/provisioning-kubevirt.adoc
```diff
@@ -0,0 +1,96 @@
...
+STREAM='stable'
+coreos-installer download -s $STREAM -p kubevirt -f ociarchive
+----
+
+Alternatively, you can manually download a KubeVirt image from the
+https://getfedora.org/coreos/download?tab=cloud_operators[download page] and directly download the ociarchive from there.
+
+== Uploading the Image to a Container Registry
+
+.Upload the image to your container registry
```
bgilbert, line 29: Do we not plan to upload the image ourselves as part of the release process?

modules/ROOT/pages/provisioning-kubevirt.adoc
rmohr: Switched to using a k8s secret where the ignition file can directly be uploaded via a `kubectl` call. That may be the better solution. :+1:

modules/ROOT/pages/provisioning-kubevirt.adoc
bgilbert: Added a comment there.  If possible, let's get a push location worked out and avoid the copy awkwardness.  We should point users directly to the stream tags in our repo, though, rather than having them copy the image elsewhere.  We might still want to document the download/upload flow, in case there are users using private registries without direct Internet access, but let's put it at the bottom of the page and document it as an advanced case.

modules/ROOT/pages/provisioning-kubevirt.adoc
```diff
@@ -0,0 +1,92 @@
...
+FILE=fedora-coreos-XX.XXXXXXXX.X.X-kubevirt.x86_64.ociarchive
+skopeo copy oci-archive:${FILE} docker://quay.io/myorg/fcos:dev
+----
+
+== Launching a VM instance
+
+Now that you have an image it can be used on your VMI definitions.
+
+You'll also need the Ignition config.
+Here the ignition config is exposed to the VMI via a Kubernetes Secret.
```
bgilbert, line 41:
```suggestion
Here the Ignition config is exposed to the VMI via a Kubernetes Secret.
```

modules/ROOT/pages/provisioning-kubevirt.adoc
```diff
@@ -0,0 +1,92 @@
...
+Learn about various ways on how to expose user data to VMIs in the https://kubevirt.io/user-guide/virtual_machines/startup_scripts/#startup-scripts[user guide].
```
bgilbert, line 42:
```suggestion
Learn about various ways to expose userdata to VMIs in the https://kubevirt.io/user-guide/virtual_machines/startup_scripts/#startup-scripts[KubeVirt user guide].
```

modules/ROOT/pages/provisioning-kubevirt.adoc
```diff
@@ -0,0 +1,92 @@
...
+.Upload the image to your container registry
+[source, bash]
+----
+FILE=fedora-coreos-XX.XXXXXXXX.X.X-kubevirt.x86_64.ociarchive
+skopeo copy oci-archive:${FILE} docker://quay.io/myorg/fcos:dev
+----
+
+== Launching a VM instance
+
+Now that you have an image it can be used on your VMI definitions.
```
bgilbert, line 38:
```suggestion
Now that you have an image, it can be used in your VMI definitions.
```

modules/ROOT/pages/provisioning-kubevirt.adoc
rmohr: :+1: Updated with a possible draft on how this would look like.

dustymabe:
Regarding push location I opened https://github.com/coreos/fedora-coreos-tracker/issues/1171. We need to come up with some idea of how we want our containers to be organized in quay. We'll discuss in next week's meeting.

modules/ROOT/pages/provisioning-kubevirt.adoc
```diff
@@ -0,0 +1,87 @@
...
+
+== Mirroring the image for use in private registries
+
+If a private registry in air-gaped installations is used, the image can be mirrored to that registry.
+
+.Mirroring a fcos image from stable stream
+[source, bash]
+----
+skopeo copy docker://quay.io/coreos/fedora-coreos:stable docker://myregistry.io/myorg/fedora-coreos:stable
+----
```
dustymabe, line 87: may also be cool to document the "download from s3 and upload" case here too (using `coreos-installer download` I presume) in case people need to physically copy the file around to get it into the environment they want. The command above assumes skopeo has access to the net and to the disconnected env, which may not always be true.

dustymabe (commented):
Thanks for working on this @rmohr

modules/ROOT/pages/provisioning-kubevirt.adoc
```diff
@@ -0,0 +1,87 @@
...
+      - disk:
+          bus: virtio
+        name: containerdisk
+      - disk:
+          bus: virtio
+        name: cloudinitdisk
+      rng: {}
+    resources:
+      requests:
+        memory: 1024M
```
dustymabe, line 57:
```suggestion
        memory: 2048M
```

1GiB seems a bit small.

@bgilbert - WDYT?

modules/ROOT/pages/provisioning-kubevirt.adoc
cgwalters: But one doesn't have to copy between two registries.  `skopeo copy docker://quay.io/coreos/fedora-coreos:stable oci:fcos` gives you a local directory that you can then copy around however you'd transport regular files (e.g. `cp fcos /mnt/usbstick`).  Then on the other end of the sneakernet, `skopeo copy oci:fcos docker://examplecorp.registry.int/fedora-coreos:stable`.

I think using native container copy flows should be preferred over plain HTTP.

bgilbert: Ah, okay.  Let's document that flow explicitly then?

cgwalters: Or to say it more strongly...I think we can just say "any tool that understands container images can be used to mirror", and link to e.g. https://github.com/containers/skopeo/blob/main/docs/skopeo-sync.1.md

(Which already documents exactly this with `--dir`, etc.)

rmohr:
Closing this one since I am not actively working on it. Doing some cleanup. Thanks everyone for your help.

PR #405, opened 2022-05-16
Updated getting-started-libvirt.adoc

copperii:
Added note to fix permission and network not found errors.

### DIFF
--- a/modules/ROOT/pages/getting-started-libvirt.adoc
+++ b/modules/ROOT/pages/getting-started-libvirt.adoc
@@ -34,7 +34,10 @@ NOTE: `virt-install` requires both the OS image and Ignition file to be specifie
 
 NOTE: Depending on your version of `virt-install`, you may not be able to use `--os-variant=fedora-coreos-*` and will get an error. In this case, you should pick an older Fedora variant (`--os-variant=fedora31` for example). You can find the variants that are supported by you current version of `virt-install` with `osinfo-query os | grep '^\s*fedora'`.
 
-NOTE: `virt-install` on Fedora 36 required additional steps:  Give proper selinux permission to ign-file: `chcon -t svirt_home_t /path/to/example.ign` and add `--network bridge=virbr0`
+NOTE: `virt-install` on Fedora 36 required additional steps:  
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/getting-started-libvirt.adoc
```diff
@@ -34,6 +34,8 @@ NOTE: `virt-install` requires both the OS image and Ignition file to be specifie
 
 NOTE: Depending on your version of `virt-install`, you may not be able to use `--os-variant=fedora-coreos-*` and will get an error. In this case, you should pick an older Fedora variant (`--os-variant=fedora31` for example). You can find the variants that are supported by you current version of `virt-install` with `osinfo-query os | grep '^\s*fedora'`.
 
+NOTE: `virt-install` on Fedora 36 required additional steps:  Give proper selinux permission to ign-file: `chcon -t svirt_home_t /path/to/example.ign` and add `--network bridge=virbr0`
```
travier, line 37: Could you add the chcon command like we already have in https://github.com/coreos/fedora-coreos-docs/blob/main/modules/ROOT/pages/tutorial-autologin.adoc#booting-fedora-coreos ?

travier, line 37: Any specific reason for specifying explicitly the network?

travier (changes requested):

modules/ROOT/pages/getting-started-libvirt.adoc
copperii: The reason for explicitly specifying the network is "default network not found" error.  Maybe just adding a link to "Booting Fedora CoreOS" would be enough, since needed parameters and steps are specified there.

travier: Sorry I wasn't clear enough. Could you move the command to be part of the bash snippet above like we did on the other pages?

travier: How indeed. Let's use the same ones here.

travier (approved):

dustymabe (approved):
LGTM

PR #416, opened 2022-06-17

coreosbot-releng:


jlebon:
@bgilbert Should we drop this reference? We don't actually use it anywhere.

bgilbert:
Yeah, it was added in case we needed it later, but we might as well reduce noise.

bgilbert:
Closing in favor of #417.

PR #426, opened 2022-07-04
manual-rollbacks: show full roll-back and roll-forward flows

lucab:
This expands the section describing permanent rollbacks, in order
to make it more explicit how to deal with auto-updates and to show
how to manually perform roll-forward steps if required.
The scenario came up in a chat thread, and it is good to have the
commands written down for reference.

### DIFF
--- a/modules/ROOT/pages/manual-rollbacks.adoc
+++ b/modules/ROOT/pages/manual-rollbacks.adoc
@@ -8,5 +8,33 @@ To temporarily boot the previous OS deployment, hold down `Shift` during the OS 
 
 == Permanent rollback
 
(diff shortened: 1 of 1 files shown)


travier:
Would re-enabling zincati not be enough? Directly rebasing might skip barrier releases?

lucab:
Yes, well, this gets into the weird territory where intermediate release (e.g. barriers) may not be working.
But I agree for most cases just re-enabling Zincati would be enough, let's stick to that for now. I've pushed an additional commit to simplify the manual steps.

travier (approved):

travier:
Let's update those once we have something safe in Zincati to specify the version to update too?

PR #429, opened 2022-07-12
pages: setting a GRUB password

saqibali-2k:
We recently landed GRUB password support. Let's add an entry in our docs
that outlines the process of setting up the new functionality.

PR #433, opened 2022-07-20
grub-password: remove implementation detail

bgilbert:
The point of the sugar is that we don't want users to interact with `user.cfg` directly.  By explicitly stating what's happening under the covers, we're encouraging them to do so.  Anyone who's sufficiently curious can still look at the Butane output.

Followup to #429, where I was too slow on the review.  :eyes: 

cc @saqibali-2k @travier @dustymabe


dustymabe (approved):

saqibali-2k (approved):
Fair point :+1:

PR #436, opened 2022-07-28
docs: Remove all overrides for printk

cgwalters:
We effectively do this by default now after
https://github.com/coreos/fedora-coreos-config/pull/1840

### DIFF
--- a/modules/ROOT/pages/tutorial-containers.adoc
+++ b/modules/ROOT/pages/tutorial-containers.adoc
@@ -4,7 +4,7 @@ NOTE: Make sure that you have completed the steps described in the xref:tutorial
 
 In this tutorial, we will setup SSH access and start a container at boot. Fedora CoreOS is focused on running applications/services in containers thus we recommend trying to run containers and avoid modifying the host directly. Running containers and keeping a pristine host layer makes automatic updates more reliable and allows for separation of concerns with the Fedora CoreOS team responsible for the OS and end-user operators/sysadmins responsible for the applications.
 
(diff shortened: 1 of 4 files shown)


modules/ROOT/pages/tutorial-updates.adoc
```diff
@@ -78,13 +78,6 @@ storage:
         inline: |
           # Tell systemd to not use a pager when printing information
           export SYSTEMD_PAGER=cat
-    - path: /etc/sysctl.d/20-silence-audit.conf
```
bgilbert, line 81: Need to remove line 47 also.

modules/ROOT/pages/tutorial-containers.adoc
```diff
@@ -84,13 +84,6 @@ storage:
         inline: |
           # Tell systemd to not use a pager when printing information
           export SYSTEMD_PAGER=cat
-    - path: /etc/sysctl.d/20-silence-audit.conf
```
bgilbert, line 87: Need to fix line 7 also.

dustymabe:
I think it would also be nice to add a section for setting this (i.e. what we just instructed a user to do in IRC) so we can point people at it.

Can we do that here too? Maybe just add an entry under https://docs.fedoraproject.org/en-US/fedora-coreos/sysctl/ ?

bgilbert (commented):
Good catch!

modules/ROOT/pages/tutorial-updates.adoc
cgwalters: Thanks, well spotted!

dustymabe (approved):
LGTM

PR #444, opened 2022-08-17
Add note about disabling SELinux

jldohmann:
This adds a warning to the user when attempting to re-enable SELinux after temporarily disabling.


jldohmann:
@travier could you PTAL? I saw the issue about adding a note for this; I recently added similar notes for RHCOS. Lmk what you think.

travier:
While the text looks good, I don't think we should have that front and center on the getting started page. Maybe we should have dedicated page for SELinux as we also need examples for setting booleans on boot.

travier:
Can you link the the RHCOS docs?

travier:
I've merged a general SELinux page in https://github.com/coreos/fedora-coreos-docs/pull/789 so closing this one.

Issue #441, opened 2022-08-09
Example for non root ownership for files & parent dirs


travier:
Until we have a better syntax / sugar in Butane, let's add an example to explain how one should set the permissions and ownership for a file and its parent dirs with Butane.

This might need to be also included in the Butane & Ignition docs.

See:
- https://discussion.fedoraproject.org/t/ignition-when-a-file-is-created-parent-dir-is-still-owned-by-root/41222
- https://github.com/coreos/butane/issues/380

travier:
Probably in https://docs.fedoraproject.org/en-US/fedora-coreos/managing-files/

travier:
This one also need an update: https://docs.fedoraproject.org/en-US/fedora-coreos/tutorial-user-systemd-unit-on-boot/

travier:
Managing Files page PR: https://github.com/coreos/fedora-coreos-docs/pull/459
Tutorial PR: https://github.com/coreos/fedora-coreos-docs/pull/458

travier:
With https://github.com/coreos/fedora-coreos-docs/pull/459 & https://github.com/coreos/fedora-coreos-docs/pull/460 merged, this should be fixed.

PR #450, opened 2022-08-31
Add example for non-root ownership for files & parent directories

mohelt:
managing-files: Add example for non root ownership

Add an example for non root ownership for files
and parent directories

fixes: #441

### DIFF
--- a/modules/ROOT/pages/managing-files.adoc
+++ b/modules/ROOT/pages/managing-files.adoc
@@ -50,3 +50,33 @@ storage:
       target: /opt/tools/transmogrifier
       hard: false
 ----
+
+.Example for how to set permissions and ownership for a file and its parent directories
+[source,yaml]
+----
+variant: fcos
+version: 1.4.0
+storage:
+  files:
+      # This creates a file, and sets the ownership
+    - path: /home/directory/you/want.txt
+      contents:
+        inline: Hello, world!
+      user:
+        name: exampleuser
+      group:
+        name: examplegroup
+  # This creates the directory that the file should be inside. Its mode is set to 0755 by default, that
+  # is, readable and executable by all, and writable by the owner.
+  directories:
+    - path: /home/directory
+      user:
+        name: exampleuser
+      group:
+        name: examplegroup
+    - path: /home/directory/you
+      user:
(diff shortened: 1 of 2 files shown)


travier:
Let's re-title this PR to something more descriptive?

travier:
Can you also update https://docs.fedoraproject.org/en-US/fedora-coreos/tutorial-user-systemd-unit-on-boot/ as mentioned in the issue?

modules/ROOT/pages/tutorial-user-systemd-unit-on-boot.adoc
```diff
@@ -58,6 +58,30 @@ storage:
...
+        name: sleeper
+      group:
+        name: sleeper
+    - path: /home/sleeper/.config/
+      mode: 0755
+      user:
+        name: sleeper
+      group:
+        name: sleeper
+    - path: /home/sleeper/
```
travier, line 79: I don't think we need this one as it's created with the user.

modules/ROOT/pages/tutorial-user-systemd-unit-on-boot.adoc
```diff
@@ -97,7 +121,31 @@ passwd:
...
```
travier, line 147: same here

modules/ROOT/pages/managing-files.adoc
```diff
@@ -50,3 +50,33 @@ storage:
...
+
+.Example for how to set permissions and ownership for a file and its parent directories
+[source,yaml]
+----
+variant: fcos
+version: 1.4.0
+storage:
+  files:
+      # This creates a file, and sets the ownership
+    - path: /home/directory/you/want.txt
```
travier, line 62: Let's use a path with a name that matches the user and use a real configuration example like https://discussion.fedoraproject.org/t/ignition-when-a-file-is-created-parent-dir-is-still-owned-by-root/41222

travier:
Also make sure to validate that your example pass Butane config validation. There are errors in CI for this one.

travier:
Replaced and updated in https://github.com/coreos/fedora-coreos-docs/pull/459.

PR #451, opened 2022-08-31
Sync with upstream template

travier:
From https://gitlab.com/fedora/docs/templates/fedora-docs-template/-/merge_requests/1

See: https://gitlab.com/fedora/docs/templates/fedora-docs-template

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -19,7 +19,7 @@ elif uname | grep -iq linux; then
     # which requires root.
 
     if [ -f /usr/local/bin/antora ]; then
-        /usr/local/bin/antora "${cmd}"
+        /usr/local/bin/antora ${cmd}
     elif uname -m | grep -iwq aarch64; then
         echo "no antora/aarch64 container try just \`npm install -g @antora/cli @antora/site-generator-default\`"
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 image="docker.io/antora/antora"
-cmd="--html-url-extension-style=indexify site.yml"
+cmd="site.yml"
```
bgilbert, line 4: Will this [change our URLs](https://docs.antora.org/antora/latest/playbook/urls-html-extension-style/#html-extension-style-key)?  The [upstream commit](https://gitlab.com/fedora/docs/templates/fedora-docs-template/-/commit/bca134e1c05d93db7acecacdfefc4afabd8a1223) doesn't say why it was done.

travier: I've asked upstream. The docs are not built by this script however so this only impacts local testing. I'll update to not do that pending a response.

bgilbert (approved):

travier:
Tracked in https://gitlab.com/fedora/docs/templates/fedora-docs-template/-/issues/4

Issue #454, opened 2022-09-13
Document setting the keyboard layout


bgilbert:
Something like this should work:

```yaml
variant: fcos
version: 1.4.0
storage:
  files:
    - path: /etc/vconsole.conf
      contents:
        inline: |
           KEYMAP=fr-latin1
           FONT=eurlatgr
```

lucab:
We do have https://docs.fedoraproject.org/en-US/fedora-coreos/sysconfig-setting-keymap/ already in place.
Is this about showing a more interesting `KEYMAP`/`FONT` combination?

bgilbert:
Nope, it's just that I grepped for `keyboard` and didn't find anything.  We should probably add the words "keyboard layout" to the page for the benefit of searches, but otherwise I think we're all set here.

PR #455, opened 2022-09-13
sysconfig-setting-keymap: prefer the phrase "keyboard layout"

bgilbert:
"Keymap" is a bit jargony.  Introduce the subject with the phrase "keyboard layout" instead, but keep the last "keymap" reference because that's what `localectl` calls it.

Fixes https://github.com/coreos/fedora-coreos-docs/issues/454.


travier (approved):

dustymabe:
Thanks for working on this!

PR #458, opened 2022-09-17
tutorial: Update systemd user example docs

eriksjolund:
Add missing directories to Butane configuration

Adjust file permissions in Butane configuration

Fixes https://github.com/coreos/fedora-coreos-docs/issues/367

A side-note, I created a discussion topic  about what file permissions  the directory _~/.config_ should have:
* https://github.com/containers/podman/discussions/15843

Maybe the group and the world permissions should be removed all together for the files and directories under _~/.config_
?

Signed-off-by: Erik Sjölund <erik.sjolund@gmail.com>

PR #459, opened 2022-09-19
managing-files: Add example for non root ownership & improve layout

travier:
managing-files: Add example for non root ownership

Add an example for non root ownership for files and parent directories

See: https://github.com/coreos/fedora-coreos-docs/issues/441

---

managing-files: Improve layout for examples


travier:
The user tutorial section will be done with https://github.com/coreos/fedora-coreos-docs/pull/458 so I'm focusing this one on the example on the Files page.

modules/ROOT/pages/managing-files.adoc
```diff
@@ -1,52 +1,121 @@
...
+----
+
+This example creates a symbolic link in `/usr/local/bin` to a path in `/opt`.
+This is useful to let local processes invoke a program without altering their
+PATH environment variable.
+
+.Example to create a symbolic link
+[source,yaml]
+----
   links:
```
c4rt0, line 72: I think we're missing here the butane header info.

c4rt0 (changes requested):
As per comment.

modules/ROOT/pages/managing-files.adoc
travier: Fixed, thanks

c4rt0 (approved):

dustymabe (approved):
LGTM

PR #468, opened 2022-09-29
alternatives: use iptables-legacy as example rather than nft

jlebon:
The nft backend of iptables is now the default in FCOS, so make the Butane config instead document how to swap from nft to legacy.


dustymabe (approved):
LGTM

Issue #469, opened 2022-09-29
Add Internet Proxy Configuration Documentation


fifofonix:
Add documentation covering this topic:

https://github.com/coreos/fedora-coreos-tracker/issues/379

PR #470, opened 2022-09-29
Add internet proxy config

fifofonix:
Concerns: https://github.com/coreos/fedora-coreos-docs/issues/469

Also, adds a default nginx.conf file which is necessary for preview.sh to function out of the box for MacOS users.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -1,0 +1,16 @@
+server {
+    listen       80;
+    server_name  localhost;
+
+    location / {
+        root   /antora/public;
+        index  index.html index.htm;
+    }
+
+    error_page   500 502 503 504  /50x.html;
+    location = /50x.html {
+        root   /usr/share/nginx/html;
+    }
+}
+
+
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -29,6 +29,7 @@
 ** xref:running-containers.adoc[Running Containers]
 ** xref:authentication.adoc[Configuring Users and Groups]
 ** xref:hostname.adoc[Setting a Hostname]
+** xref:proxy.adoc[Proxied Internet Access]
 ** xref:sysconfig-setting-keymap.adoc[Setting Keyboard Layout]
 ** xref:os-extensions.adoc[Adding OS extensions]
 ** xref:customize-nic.adoc[How to Customize a NIC Name]
--- a/modules/ROOT/pages/proxy.adoc
+++ b/modules/ROOT/pages/proxy.adoc
@@ -1,0 +1,112 @@
+= Proxied Internet Access
+
+If you are deploying to an environment requiring internet access via a proxy, you will want to configure services so that they can access resources as intended.
+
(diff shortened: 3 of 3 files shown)


nginx.conf
```diff
@@ -0,0 +1,16 @@
+server {
```
travier, line 1: Not sure you wanted to add that file in this one?

travier (commented):
LGTM overall. Two small nits:
- we use single space in the docs.
- not sure you wanted to add that nginx.conf file here

travier:
Thanks for working on this!

nginx.conf
fifofonix: I did mean to add this actually because it is necessary for MacOS docs contributors.  I've modified description above to reference the fact that this is added.

fifofonix:
> Thanks for working on this!

@travier you're welcome.  Issues resolved.  I'm not familiar with protocols here and whether I am expected to mark things above as resolved, or whether that is for you given you raised the questions?

nginx.conf
travier: Let's make another PR then? Have you tried the build.sh & preview.sh scripts?

travier:
> Issues resolved. I'm not familiar with protocols here and whether I am expected to mark things above as resolved, or whether that is for you given you raised the questions?

I think you can mark them as resolved yourself, at least that's what I do.

nginx.conf
dustymabe: yeah - I told him to add this file (he is on a mac and it's required). See https://gitlab.com/fedora/docs/docs-website/docs-fp-o/-/blob/prod/nginx.conf

@fifofonix - can you put the addition of this file in a separate commit with appropriate commit message?

dustymabe:
Thanks for working on this @fifofonix - looking good to me. Let's get that nginx.conf file broken out into a separate commit and then the extraneous two commits that you have right now squashed. At the end I'm thinking two commits:

- Add nginx.conf file
- Add internet proxy config doc

nginx.conf
fifofonix: Separate commits as requested.

dustymabe:
This LGTM - only thing I would add is maybe linking to the [original location](https://gitlab.com/fedora/docs/docs-website/docs-fp-o/-/blob/prod/nginx.conf) where the `nginx.conf` file was copied from in the commit message for that commit.

@travier do you want to do one more review pass?

fifofonix:
> This LGTM - only thing I would add is maybe linking to the [original location](https://gitlab.com/fedora/docs/docs-website/docs-fp-o/-/blob/prod/nginx.conf) where the `nginx.conf` file was copied from in the commit message for that commit.
> 
> @travier do you want to do one more review pass?

Done

travier:
If you can update your commit message to mention https://gitlab.com/fedora/docs/templates/fedora-docs-template/-/blob/main/nginx.conf instead as it's the template repo that would be great. LGTM otherwise.

PR #478, opened 2022-10-18
platforms: update minimum ESXi version; add Workstation/Fusion

bgilbert:
Bump hardware version to 17.  Also, it appears this page has always only referenced ESXi, which isn't correct; we run on Workstation and Fusion too.  Reword accordingly.


dustymabe (approved):
LGTM

bgilbert:
Followup in #507.

PR #479, opened 2022-10-21
add s390x IBMCloud docs; add entry to platforms page

dustymabe:


modules/ROOT/pages/platforms.adoc
```diff
@@ -31,6 +31,13 @@ The currently supported platforms and their identifiers are listed below.
 * QEMU (`qemu`): Hypervisor. See xref:provisioning-libvirt.adoc[Booting on libvirt]
 * OpenStack (cloud platform): `openstack`): Cloud platform. See xref:provisioning-openstack.adoc[Booting on OpenStack].
 
+=== s390x
+
+* IBM Cloud, VPC Generation 2 (`ibmcloud`): Cloud platform. See xref:provisioning-ibmcloud.adoc[Booting on IBM Cloud].
+* Bare metal (`metal`): With BIOS, or network boot, with standard or 4k Native disks. See xref:bare-metal.adoc[Installing on Bare Metal] or xref:live-booting-ipxe.adoc[Live-booting via iPXE].
```
dustymabe, line 37: I'll need some help on this line:

```
With BIOS, or network boot, with standard or 4k Native disks.
```

I don't know if BIOS is the right term. I don't know about network boot. Also don't think 4kNative disks work.. At least we have some tests disabled for them on s390x.

modules/ROOT/pages/provisioning-ibmcloud.adoc
```diff
@@ -71,7 +72,9 @@ ibmcloud cos upload --bucket=$BUCKET --key=$FILE --file=$FILE
 ----
 IMAGE=${FILE:0:-6}     # pull off .qcow2
 IMAGE=${IMAGE//[._]/-} # replace . and _ with -
-ibmcloud is image-create $IMAGE --file "cos://${REGION}/${BUCKET}/${FILE}" --os-name fedora-coreos-stable-amd64
+[ $ARCH == 'x86_64' ] && OSNAME='fedora-coreos-stable-amd64'
+[ $ARCH == 's390x' ] && OSNAME='red-8-s390x-byol'
```
jlebon, line 76: Feels like this deserves some explanation. E.g. something like "Fedora CoreOS is not yet a valid s390x OS name in IBM cloud, so we use RHEL8's"? (I presume that's what's happening here.)

modules/ROOT/pages/platforms.adoc
jlebon: Network boot is supported, we have tests for it. For BIOS, maybe just "From disk".
Let's just drop the 4k bit since we're not testing it anyway.

modules/ROOT/pages/provisioning-ibmcloud.adoc
dustymabe: added some context in a `NOTE:`.

jlebon (approved):

Issue #467, opened 2022-09-27
document commands for bringing up VMs on s390x/ppc64le


dustymabe:
*Inspired by https://github.com/coreos/fedora-coreos-tracker/issues/1307*

We now have `s390x` and `ppc64le`. Let's update some of our examples to handle cases where `fw_cfg` doesn't work for those platforms.

- https://docs.fedoraproject.org/en-US/fedora-coreos/provisioning-libvirt/
- https://docs.fedoraproject.org/en-US/fedora-coreos/provisioning-qemu/

Need to be updated.

For libvirt the command floating around issue comments looks like:

```
virt-install ... --disk path=$PWD/config.ign,format=raw,readonly=on,serial=ignition
```

However we should investigate the `source.startupPolicy` argument to `--disk` to see if we can make it not fatal for the VM to boot if the Ignition file is later deleted.

```
source.startupPolicy
       It defines what to do with the disk if the source file is not accessible.
```

For QEMU:

```
-drive file=foobar.ign,if=none,format=raw,readonly=on,id=ignition \
  -device virtio-blk,serial=ignition,drive=ignition
```

dustymabe:
@LakshmiRavichandran1 might be able to help us get this updated.

LakshmiRavichandran1:
https://github.com/coreos/fedora-coreos-docs/pull/480/ could help here.

PR #480, opened 2022-10-21
coreos/fedora-coreos-docs: s390x: update commands for bringing up VMs in libvirt, qemu

LakshmiRavichandran1:
Fixes: https://github.com/coreos/fedora-coreos-docs/issues/467 s390x: document commands for bringing up VMs on s390x/ppc64le

### DIFF
--- a/modules/ROOT/pages/getting-started-libvirt.adoc
+++ b/modules/ROOT/pages/getting-started-libvirt.adoc
@@ -17,9 +17,6 @@ podman run --pull=always --rm -v $HOME/.local/share/libvirt/images/:/data -w /da
 +
 [source, bash]
 ----
-# Setup the correct SELinux label to allow access to the config
-chcon --verbose --type svirt_home_t /path/to/example.ign
-
 IGNITION_CONFIG="/path/to/example.ign"
 IMAGE="/path/to/image.qcow2"
 VM_NAME="fcos-test-01"
@@ -27,12 +24,19 @@ VCPUS="2"
 RAM_MB="2048"
 STREAM="stable"
 DISK_GB="10"
+# For x86 / aarch64,
+IGNITION_DEVICE_ARG="--qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=${IGNITION_CONFIG}""
+
+# For s390x / ppc64le,
+IGNITION_DEVICE_ARG="--disk path=${IGNITION_CONFIG},format=raw,readonly=on,serial=ignition,startup_policy=optional"
+
+# Setup the correct SELinux label to allow access to the config
+chcon --verbose --type svirt_home_t ${IGNITION_CONFIG}
 
 virt-install --connect="qemu:///system" --name="${VM_NAME}" --vcpus="${VCPUS}" --memory="${RAM_MB}" \
         --os-variant="fedora-coreos-$STREAM" --import --graphics=none \
         --disk="size=${DISK_GB},backing_store=${IMAGE}" \
-        --network bridge=virbr0 \
-        --qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=${IGNITION_CONFIG}"
+        --network bridge=virbr0 ${IGNITION_DEVICE_ARG}
 ----
 
 NOTE: `virt-install` requires both the OS image and Ignition file to be specified as absolute paths.
--- a/modules/ROOT/pages/provisioning-qemu.adoc
+++ b/modules/ROOT/pages/provisioning-qemu.adoc
@@ -14,7 +14,7 @@ TIP: If running with SELinux enabled, make sure your OS image and Ignition file 
 
 == Booting a new VM on QEMU
 
-This section shows how to boot a new VM on QEMU. On this platform, The Ignition file is passed to the VM via the `-fw_cfg` parameter, which sets the `opt/com.coreos/config` key in the QEMU firmware configuration device.
+This section shows how to boot a new VM on QEMU. Based on the platform, The Ignition file is passed to the VM, which sets the ignition key in the QEMU firmware configuration device.
 
 You can use `-snapshot` to make `qemu-kvm` allocate temporary storage for the VM, or `qemu-img create` to first create a layered qcow2.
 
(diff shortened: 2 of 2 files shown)


dustymabe:
I think the `--qemu-commandline="-fw_cfg` is still the required path on `x86_64` and `aarch64`. We need to update these docs to support either path if we can.

modules/ROOT/pages/getting-started-libvirt.adoc
```diff
@@ -32,13 +32,16 @@ virt-install --connect="qemu:///system" --name="${VM_NAME}" --vcpus="${VCPUS}" -
...
-        --qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=${IGNITION_CONFIG}"
+        --disk path=path/to/example.ign,format=raw,readonly=on,serial=ignition 
 ----
 
 NOTE: `virt-install` requires both the OS image and Ignition file to be specified as absolute paths.
 
 NOTE: Depending on your version of `virt-install`, you may not be able to use `--os-variant=fedora-coreos-*` and will get an error. In this case, you should pick an older Fedora variant (`--os-variant=fedora31` for example). You can find the variants that are supported by you current version of `virt-install` with `osinfo-query os | grep '^\s*fedora'`.
 
+NOTE: In the s390x architecture, for the `startupPolicy` in the disk definition of ignition source file, its optimal to use the default value `mandatory` or `optional`.
+The value `requisite` is allowed only for cdrom or floppy disk type.
```
dustymabe, line 43: I think we'd probably want to tell the user to use `optional` to make it non-fatal if the file went away and we'd want to put it in the example. We probably don't need a separate `NOTE:` for it.

dustymabe:
> I think the `--qemu-commandline="-fw_cfg` is still the required path on `x86_64` and `aarch64`. We need to update these docs to support either path if we can.

For example.. We could define a separate variable (`IGNITION_DEVICE`) with the needed arguments for `x86_64`/`aarch64` or `ppc64le`/`s390x`.

modules/ROOT/pages/getting-started-libvirt.adoc
```diff
@@ -27,18 +27,24 @@ VCPUS="2"
 RAM_MB="2048"
 STREAM="stable"
 DISK_GB="10"
+For x86 / aarch64,
+IGNITION_DEVICE="qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=${IGNITION_CONFIG}""
+
+For s390x / ppc64le,
+IGNITION_DEVICE="disk path=${IGNITION_CONFIG},format=raw,readonly=on,serial=ignition,startup_policy=optional"
```
dustymabe, line 34: Since this is simulating bash let's add a comment `#` character to the beginning of the comment. Also let's include the `--` in the variable here and change the name to `IGNITION_DEVICE_ARG`: 

```suggestion
# For x86 / aarch64,
IGNITION_DEVICE_ARG="--qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=${IGNITION_CONFIG}""
# For s390x / ppc64le,
IGNITION_DEVICE_ARG="--disk path=${IGNITION_CONFIG},format=raw,readonly=on,serial=ignition,startup_policy=optional"
```

Then below you can do something like:

```
--network bridge=virbr0 $IGNITION_DEVICE_ARG
```

and let's make the same modifications in the qemu steps too.

modules/ROOT/pages/provisioning-qemu.adoc
```diff
@@ -38,9 +38,15 @@ If you need to persist your changes, it is recommended to set up a dedicated per
...
+IGNITION_DEVICE="fw_cfg name=opt/com.coreos/config,file=path/to/example.ign"
+
+# for s390x/ppc64le:
+IGNITION_DEVICE="drive file=path/to/example.ign,if=none,format=raw,readonly=on,id=ignition -device virtio-blk,serial=ignition,drive=ignition"
+
 qemu-kvm -m 2048 -cpu host -nographic -snapshot \
 	-drive if=virtio,file=fedora-coreos-qemu.qcow2 \
-	-fw_cfg name=opt/com.coreos/config,file=path/to/example.ign \
+	-${IGNITION_DEVICE} \
 	-nic user,model=virtio,hostfwd=tcp::2222-:22
```
dustymabe, line 50: Let's take an opportunity here to break out a few more things into variables too.

i.e. let's use `IGNITION_CONFIG` and `IMAGE` variables like is done in the libvirt directions.

dustymabe (commented):
Thanks for working on this @LakshmiRavichandran1 - I made some suggestions for a few more improvements.

Also can we squash the commits down into a single commit and do something like the following for the commit message:

```
{libvirt,qemu}: add steps for `ppc64le` and `s390x`

This commit addes logic for adding in appropriate arguments based on
what architecture you are targetting.

Fixes https://github.com/coreos/fedora-coreos-docs/issues/467
```

jschintag:
I looked at the 2 docs, and i noticed that in `provisioning-qemu.adoc` there is also directly at the beginning in the text:
```
This section shows how to boot a new VM on QEMU. On this platform, The Ignition file is passed to the VM via the -fw_cfg parameter, which sets the opt/com.coreos/config key in the QEMU firmware configuration device.
```
I think we should reflect the different paths there as well.

dustymabe:
> I think we should reflect the different paths there as well.

yes, OR make the text there more generic so it doesn't mention specifics.

modules/ROOT/pages/getting-started-libvirt.adoc
LakshmiRavichandran1: addressed in eab3b9da4c5293a514974465205b2a8bfe8e28e7

modules/ROOT/pages/provisioning-qemu.adoc
LakshmiRavichandran1: addressed in eab3b9da4c5293a514974465205b2a8bfe8e28e7

jschintag:
Since i'm reviewing the doc, i noticed for libvirt:
```
# Setup the correct SELinux label to allow access to the config
chcon --verbose --type svirt_home_t /path/to/example.ign

IGNITION_CONFIG="/path/to/example.ign"
```
We should move the `chcon` to later and use the `IGNITION_CONFIG` variable

jschintag:
Also, i get an error when running `virt-install`
```
[jschinta@m8345043 fcos-test]$ virt-install --connect="qemu:///system" --name="${VM_NAME}" --vcpus="${VCPUS}" --memory="${RAM_MB}" \
        --os-variant="fedora-coreos-$STREAM" --import --graphics=none \
        --disk="size=${DISK_GB},backing_store=${IMAGE}" \
        --network bridge=virbr0 ${IGNITION_DEVICE_ARG}
ERROR    Unknown --disk options: ['startupPolicy']
[jschinta@m8345043 fcos-test]$ echo $IGNITION_DEVICE_ARG
--disk path=/home/jschinta/fcos-test/base.ign,format=raw,readonly=on,serial=ignition,startupPolicy=optional
[jschinta@m8345043 fcos-test]$
```

LakshmiRavichandran1:
@jschintag 's comments are addressed in bae1b337d04e2fc699926373f8f465025e5f832b

jschintag:
lgtm
Thank you @LakshmiRavichandran1 for taking care of this

bgilbert:
(FYI for next time: please push PR branches to a fork of this repo, not directly into the repo itself.)

dustymabe:
Nice work! Thanks @LakshmiRavichandran1

PR #483, opened 2022-10-29
Document new-style console configuration

bgilbert:
Mainly this affects the emergency shell docs.  Also add a section to the kargs page in case a user goes directly to that doc.

Delete the "temporary tweak" section, since it doesn't make sense now.  We previously needed it for removing unwanted console arguments so the booted console would be usable.  Now, with per-platform defaults, the user will likely want to add consoles rather than remove them, but in that case GRUB won't show a menu on the desired console in the first place.

Also stop using `console` as an example on the kargs page.  Manually changing the argument is still valid under some circumstances, but we can switch to a different one to avoid confusing the issue.

Also clean up some references to the serial console in a number of places where the "serial" part is an implementation detail.

### DIFF
--- a/modules/ROOT/pages/emergency-shell.adoc
+++ b/modules/ROOT/pages/emergency-shell.adoc
@@ -3,54 +3,64 @@
 Sometimes you may want to access the node console to perform troubleshooting steps or emergency maintenance.
 For instance, you may want to access the emergency shell on the console, in order to debug first boot provisioning issues.
 
-== Default console setup
+== Default console configuration
 
-All Fedora CoreOS (FCOS) images come with a default configuration for the console which is meant to accommodate most virtualized and bare-metal setups.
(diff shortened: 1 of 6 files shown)


modules/ROOT/pages/emergency-shell.adoc
```diff
@@ -3,54 +3,64 @@
...
+If you are launching FCOS from an image (in a cloud or a virtual machine), you can use Ignition to configure the console at provisioning time.
+
+.Example: Enabling primary serial and secondary graphical console
+[source,yaml]
+----
+variant: fcos
+version: 1.4.0
+kernel_arguments:
+  should_exist:
+    # Order is significant, so group both arguments into the same list entry
```
jlebon, line 43: Minor: missing period

modules/ROOT/pages/kernel-args.adoc
```diff
@@ -32,13 +32,29 @@ kernel_arguments:
...
+
+.Example: Configuring a primary serial console and secondary graphical console when installing to `/dev/sda`
+
+[source,bash]
+----
+$ sudo coreos-installer install --ignition-url https://example.com/config.ign \
+      --console tty0 --console ttyS0,115200n8 /dev/sda
+----
+
+This will add `console` arguments to the kernel command line, plus equivalent parameters to the GRUB bootloader configuration.
```
jlebon, line 47: Is this worth repeating rather than letting the reader follow the links?

bgilbert:
Updated!

jlebon (approved):

PR #486, opened 2022-11-16
Fix instructions to get Connection String

Sid-Sun:
Due to an ongoing bug with the azure cli: https://github.com/Azure/azure-cli/issues/24528 the `az storage account show-connection-string -n "${az_storage_account}"` command fails, since we already have az_resource_group defined here, we can workaround it :D


dustymabe (approved):
LGTM - thanks for the contribution!

PR #488, opened 2022-11-17
Point to a page that exists.

### LINKED ISSUE #487: First boot page doesn't exist
https://docs.fedoraproject.org/en-US/fedora-coreos/installation/ has a broken link to the provision page because it uses `link` instead of `xref`, so Antora preserves the `.adoc`. PR incoming.

funnelfiasco:
Fixes #487

Signed-off-by: Ben Cotton <bcotton@fedoraproject.org>

This might not actually be the page you wanted to direct people toward, but at least it is there. The page I'm editing doesn't seem to be active, but there are links to it in various places, so it's worth keeping it working.

### DIFF
--- a/modules/ROOT/pages/installation.adoc
+++ b/modules/ROOT/pages/installation.adoc
@@ -1,6 +1,6 @@
 = Installation Guide
 
-This installation guide aims at helping you link:first-boot.adoc[provision your first boot] as well as explain how to boot on the various platforms.
+This installation guide aims at helping you xref:getting-started.adoc[provision your first boot] as well as explain how to boot on the various platforms.
 
 [[booting-on-azure]]
 Booting on Azure


lucab:
Thanks for the fix! I think this is an orphan/empty page that could be directly dropped (`first-boot.adoc` actually got dropped in the same way).

funnelfiasco:
Dropping it would be fine, but in that case we should use page-aliases to catch inbound links:

https://docs.antora.org/antora/latest/page/page-aliases/

travier:
This is an old page indeed. Redirecting it to https://docs.fedoraproject.org/en-US/fedora-coreos/getting-started/ would be great.

jlebon:
Hmm, I'm not seeing any links to this `installation.adoc` from the existing pages. OOC, how did you land there? Redirecting sounds good, but it might be good to update pages linking to it too if possible.

funnelfiasco:
>  OOC, how did you land there? 

This was reported in the #fedora IRC channel, but they didn't say how they got there.

bgilbert:
Looking at the history, this page has never had any content, and the only link to it on the site was removed in July 2019.  I think it's reasonable to not bother redirecting external links in this case, since the page never had any content worth linking.  I've filed https://github.com/coreos/fedora-coreos-docs/pull/568 to just drop the page.

Issue #489, opened 2022-11-22
The command `virt-install` run failed


imichael2e2:
When running commands in `getting-started-libvirt.adoc`, several
errors are reported for `virt-install`:

```
Starting install...
Allocating 'fcos-test-02-2.qcow2'                                         |    0 B  00:00:00 ... 
Removing disk 'fcos-test-02-2.qcow2'                                      |    0 B  00:00:00     
ERROR    internal error: process exited while connecting to monitor: 2022-11-22T02:13:17.404468Z qemu-kvm: on,obsolete=deny,elevateprivileges=deny,spawn=deny,resourcecontrol=deny: Could not open 'on,obsolete=deny,elevateprivileges=deny,spawn=deny,resourcecontrol=deny': No such file or directory
Domain installation does not appear to have been successful.
```

With debug info, bash shows the command being run is
```
+ virt-install --connect=qemu:///system --name=fcos-test-02 --vcpus=1 --memory=1024 --os-variant=fedora-coreos-stable --import --graphics=none --disk=size=10,backing_store=/path/to/fedora-coreos-36.20221030.3.0-qemu.x86_64.qcow2 --network bridge=virbr0 --qemu-commandline=-fw_cfg
```
The `-fw_cfg blabla` is just left half, another part is missing.

It seems like the common quoting issue for Bash, so I tried to escape
the quotes like

```
IGNITION_DEVICE_ARG="--qemu-commandline=\"-fw_cfg name=opt/com.coreos/config,file=${IGNITION_CONFIG}\"" 
```

But another error got:

```
Starting install...
Allocating 'fcos-test-02-2.qcow2'                                         |    0 B  00:00:00 ... 
Removing disk 'fcos-test-02-2.qcow2'                                      |    0 B  00:00:00     
ERROR    internal error: process exited while connecting to monitor: qemu-kvm: -fw_cfg name=opt/com.coreos/config,file=/path/to/mycoreos.ign: invalid option
Domain installation does not appear to have been successful.
```

With debug info, Bash shows the command being run is:
```
[...]

PR #490, opened 2022-11-22
Fix the issue #489: The command "virt-install" run failed

imichael2e2:
Unescaped double quotes in double quotes like `"thisis="content""` will
be interpreted as a string: `thisis=content`, everything works except
that there are spaces in `content`, which leads to *Word Splitting* (man bash). 
In this scenario, it splits the expected one string: `-fw_cfg blablabla` into two
string: `-fw_cfg` and `blablabla`, which indirectly leads to the
"No such file..." error.

Escaped double quotes in double quotes like `"thisis=\"content\""`
will be interpreted as a single quoted string: `'thisis="content"'`,
which is passed to the program that Bash would spawn, the unnecessary
leading single quote will incorrectly parsed as the part of optname,
which is undesirable, in this scenario, `virt-install` will get a
option like `'--qemu-commandline=...'`, which also leads to "invalid
option" error.

As suggested in http://mywiki.wooledge.org/BashFAQ/050, we can employ
Bash's Arrays to work around this issue, as shown in git commit.

### DIFF
--- a/modules/ROOT/pages/getting-started-libvirt.adoc
+++ b/modules/ROOT/pages/getting-started-libvirt.adoc
@@ -25,10 +25,10 @@ RAM_MB="2048"
 STREAM="stable"
 DISK_GB="10"
 # For x86 / aarch64,
-IGNITION_DEVICE_ARG="--qemu-commandline="-fw_cfg name=opt/com.coreos/config,file=${IGNITION_CONFIG}""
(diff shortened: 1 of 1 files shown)


jlebon (approved):
LGTM, thanks for the detailed commit message!

imichael2e2:
Thanks! The Fedora community  deserves that!

PR #493, opened 2022-11-30
provisioning-gcp: Mention `user-data`

cgwalters:
This was not obvious to me.


modules/ROOT/pages/provisioning-gcp.adoc
```diff
@@ -71,3 +71,10 @@ gcloud compute instances create --metadata-from-file "user-data=${CONFIG}" --ima
 ----
 
 NOTE: By design, https://cloud.google.com/compute/docs/startupscript[startup scripts] are not supported on FCOS. Instead, it is recommended to encode any startup logic as systemd service units in the Ignition configuration.
+
+=== Ignition via the GCP console
```
jlebon, line 75: How about instead of making this a separate section, we rework the sentence a bit above to say:

> In order to launch a customized FCOS instance, a valid Ignition configuration must be passed as metadata under the `user-data` key at creation time. In the web console, this is available under the Management section. From the command-line, use `--metadata-from-file`:

cgwalters: Thanks, indeed that's better.  I made some more tweaks.

modules/ROOT/pages/provisioning-gcp.adoc
```diff
@@ -71,3 +74,6 @@ gcloud compute instances create --metadata-from-file "user-data=${CONFIG}" --ima
 ----
 
 NOTE: By design, https://cloud.google.com/compute/docs/startupscript[startup scripts] are not supported on FCOS. Instead, it is recommended to encode any startup logic as systemd service units in the Ignition configuration.
+Again, note you need to use the `user-data` key for Ignition; it will also not work to paste Ignition into this field in the web console.
+
+
```
jlebon, line 79: Minor/optional: extraneous lines

jlebon (approved):
LGTM!

jlebon (approved):

PR #503, opened 2023-02-06
provisioning-vultr: discourage using standard Vultr images

bgilbert:
Vultr includes all three FCOS streams in their distro list, but the images are from last September, and we don't want to encourage launching from obsolete images.  Update the intro text to say that these images exist but we don't recommend them.


jlebon (approved):
Have we tried reaching out to them?

travier (approved):

bgilbert:
@dustymabe is going to try.

PR #507, opened 2023-03-06
updating minimum esxi version in docs

scuba-st3v3:
- It appears the minimum esxi version is tied to `vmx-17` ([article](https://kb.vmware.com/s/article/1003746))
- History https://github.com/coreos/fedora-coreos-tracker/issues/1119


bgilbert (approved):
Good catch, thanks!

PR #509, opened 2023-03-09
Stabilize Butane spec 1.5.0

bgilbert:


bgilbert:
The new Butane release is tagged but hasn't shipped to all channels yet.  However, the Ignition 3.4.0 spec stabilization has already rolled out to FCOS stable, so the `-experimental` version of the  `fcos` `1.5.0` spec produces Ignition configs that current FCOS versions won't accept.  As such, merging this will be an improvement over the current state of the docs, so I'll go ahead.

PR #510, opened 2023-03-10
libvirt.adoc: add note about `DISK_GB`

HuijingHei:
NOTE: `DISK_GB` should be >= OS image size, if failed to boot with `dev-disk-by\x2dlabel-boot.device: Job dev-disk-by\x2dlabel-boot. device/start timed out.`, when you get emergency shell, run `sgdisk -p /dev/vda` to check the partition table.

### DIFF
--- a/modules/ROOT/pages/getting-started-libvirt.adoc
+++ b/modules/ROOT/pages/getting-started-libvirt.adoc
@@ -43,6 +43,8 @@ NOTE: `virt-install` requires both the OS image and Ignition file to be specifie
 
 NOTE: Depending on your version of `virt-install`, you may not be able to use `--os-variant=fedora-coreos-*` and will get an error. In this case, you should pick an older Fedora variant (`--os-variant=fedora31` for example). You can find the variants that are supported by you current version of `virt-install` with `osinfo-query os | grep '^\s*fedora'`.
 
+NOTE: `DISK_GB` should be at least as big as the default size of the image. For Fedora CoreOS, this is currently 10 GB.
+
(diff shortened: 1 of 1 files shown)


travier:
We're currently using 10 GB for disk size so you can not use anything smaller as that would just truncate the disk: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/image-base.yaml#L4

travier:
Maybe we should mention that 10 GB is the minimum here instead?

HuijingHei:
Actually `10G` works for `fedora-coreos`.
The problem is when I try rhcos, the default is 16G (define in https://github.com/openshift/os/blob/master/image-rhel-8.6.yaml#L2), which always makes the VM failed to boot.

modules/ROOT/pages/getting-started-libvirt.adoc
```diff
@@ -43,6 +43,8 @@ NOTE: `virt-install` requires both the OS image and Ignition file to be specifie
 
 NOTE: Depending on your version of `virt-install`, you may not be able to use `--os-variant=fedora-coreos-*` and will get an error. In this case, you should pick an older Fedora variant (`--os-variant=fedora31` for example). You can find the variants that are supported by you current version of `virt-install` with `osinfo-query os | grep '^\s*fedora'`.
 
+NOTE: `DISK_GB` should be >= OS image size, if failed to boot with `dev-disk-by\x2dlabel-boot.device: Job dev-disk-by\x2dlabel-boot.device/start timed out.`, when you get emergency shell, run `sgdisk -p /dev/vda` to check the partition table.
```
travier, line 46:
```suggestion
NOTE: `DISK_GB` should be at least as big as the default size of the image. For Fedora CoreOS, this is currently 10 GB.```

dustymabe (approved):

PR #515, opened 2023-03-22
provisioning: Document Podman Machine

cgwalters:
I was watching https://containerplumbing.org/sessions/2023/when_podman_desktop_eats which demoed customizing the FCOS OS image, and I think it's long overdue that we cross-link Podman Machine from the FCOS docs!

### DIFF
--- a/modules/ROOT/pages/platforms.adoc
+++ b/modules/ROOT/pages/platforms.adoc
@@ -19,6 +19,7 @@ The currently supported platforms and their identifiers are listed below.
 * Bare metal (`metal`): With BIOS, UEFI or network boot, with standard or 4k Native disks. See xref:bare-metal.adoc[Installing on Bare Metal] or xref:live-booting-ipxe.adoc[Live-booting via iPXE].
 * Nutanix (`nutanix`): Hypervisor.
 * OpenStack (cloud platform): `openstack`): Cloud platform. See xref:provisioning-openstack.adoc[Booting on OpenStack].
+* Podman Machine (local): See xref:provisioning-podman-machine.adoc[Booting on Podman]
 * QEMU (`qemu`): Hypervisor. See xref:provisioning-libvirt.adoc[Booting on libvirt]
 * VirtualBox ('virtualbox'): Hypervisor. See xref:provisioning-virtualbox.adoc[Booting on VirtualBox].
 * VMware ESXi, Fusion, and Workstation (`vmware`): Hypervisor. See xref:provisioning-vmware.adoc[Booting on VMware]. Fedora CoreOS images currently use https://kb.vmware.com/s/article/1003746[hardware version] 17, supporting VMware ESXi 7.0 or later, Fusion 12.0 or later, and Workstation 16.0 or later.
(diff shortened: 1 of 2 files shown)


cgwalters:
(You're right that it'd make sense too on the "using" page for sure)

Hmmm...perhaps what would make sense is to reorganize the "provisioning" page to have two sections:

- Cloud and Bare Metal (test/production)
- Local (non-cloud hypervisors) and I think Podman Machine would be featured prominently here

bgilbert:
The lack of a platform ID is a symptom of the fact that users aren't launching FCOS directly in that case.  (In fact, Podman Machine VMs will have multiple platform IDs under the covers, for macOS vs. Windows etc.)  They're using it as part of Podman, and FCOS is mostly an implementation detail.

From my perspective there isn't any difference between "how to use Podman Machine, which uses FCOS" and "how to use Typhoon, which can use FCOS".  There may be value to having them on the site, but they're not really about how to use FCOS, and we don't have a place to put them in our current structure.  In particular, I don't think the docs should privilege Podman Machine over other layered projects.

dustymabe:
> @cgwalters 
> (You're right that it'd make sense too on the "using" page for sure)

Yeah. I agree this would probably be best on the "using" page, but...


> @bgilbert 
> From my perspective there isn't any difference between "how to use Podman Machine, which uses FCOS" and "how to use Typhoon, which can use FCOS". There may be value to having them on the site, but they're not really about how to use FCOS, and we don't have a place to put them in our current structure.

I think it would be nice if our "using" page (or something like it) could have more expanded sections about either projects that layer on top of FCOS (OKD, Typhoon, Podman Machine) OR projects that could be used in conjunction with FCOS (something like instructions for use with k3s), which is where something like this could go.

cgwalters:
> From my perspective there isn't any difference between "how to use Podman Machine, which uses FCOS" and "how to use Typhoon, which can use FCOS".

I see your point, but there are also some concrete differences.  One notable one is that PM (just going to abbreviate) *always* uses FCOS, whereas Typhoon does not (though they do make it relatively clear though that you have to pick).

I'd also say that there's quite a difference between provisioning *one* machine (as everything else on that page does) that is accessible via ssh and provisioning a Kubernetes cluster.

> but they're not really about how to use FCOS

They are actually adding FCOS specific features; at Container Plumbing days Ashley demoed https://github.com/containers/podman/pull/17494

jlebon:
This seems to be catering to people who might use `podman machine` as a means to an end (of using FCOS)? People on macOS (and soon Windows) may prefer the easy UX provided by `podman machine` to get an FCOS machine rather than directly using QEMU/Hyper-V. If they actually want `podman machine` for `podman`, then they'd go to the podman docs.

For that purpose, I think I'd reframe it as more generic "Booting on macOS" and "Booting on Windows" pages listing the options available on those platforms (and linking to the relevant hypervisor-specific page, e.g. QEMU or VirtualBox or soon Hyper-V), but _also_ talking about `podman machine` as another way to bring up FCOS on those hypervisors.

cgwalters:
> For that purpose, I think I'd reframe it as more generic "Booting on macOS" and "Booting on Windows" pages listing the options available on those platforms (and linking to the relevant hypervisor-specific page, e.g. QEMU or VirtualBox or soon Hyper-V), but also talking about podman machine as another way to bring up FCOS on those hypervisors.

:100: agree.  Any objections to reworking the provisioning page around those lines from anyone?

dustymabe:
I think I agree. So basically we'd create two new pages called "Booting on macOS" and "Booting on Windows" under our provisioning tree?

SGTM

cgwalters:
If we add those sections, it does kind of argue to potentially also split off other sections as I [mentioned here](https://github.com/coreos/fedora-coreos-docs/pull/515#issuecomment-1480148304) right?

But then I was thinking about it more and there's a lot of overlap between where this thread went and what we have on https://getfedora.org/en/coreos/download?tab=cloud_launchable&stream=stable&arch=x86_64

So we could:

- Rework our docs to match the getfedora layout and nomenclature (i.e. we have separate "cloud launchable" and "bare metal" and "For cloud operators" sections in our docs too)
- Change each getfedora section to link to the corresponding docs section (this seems like a no-brainer?)
- Add a "macOS and Windows" section to both our docs and getfedora

?

bgilbert:
The "for cloud operators" section is meant to imply that users don't need to download those images, because cloud operators will ingest them directly and present them through the cloud UI.  That's not actually true in all cases, though.  On some clouds users need to download and re-upload the image, and on some they need to put the image URL directly into the cloud UI.

Leaving that aside, I don't think we should document those clouds as though only the cloud operator can run FCOS!  They'd all go in a "cloud" section if we had one.

PR #517, opened 2023-03-23
Update provisioning-vmware.adoc

deed02392:
Added a section on how to include your Ignition configuration if it's too large to be accepted as a command line argument.

### DIFF
--- a/modules/ROOT/pages/provisioning-vmware.adoc
+++ b/modules/ROOT/pages/provisioning-vmware.adoc
@@ -42,12 +42,19 @@ CONFIG_ENCODING='base64'
 CONFIG_ENCODED=$(cat example.ign | base64 -w0 -)
 ----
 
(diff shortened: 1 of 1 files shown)


Okeanos:
While this is technically a correct workaround there's an even better one available since `govc` [v0.27.0](https://github.com/vmware/govmomi/releases/tag/v0.27.0): Add feature to read file contents for ExtraConfig described in the docs in the [vm.change](https://github.com/vmware/govmomi/blob/main/govc/USAGE.md#vmchange) section.

This only applies to vCenter/vSphere deployments, though, and not VMware Fusion/Workstation (because they don't work with `govc`).

Okeanos:
Yeah, I totally understand what you mean.

I'll prepare an update for the FCOS documentation differentiating between your very cool workaround (to be used with Workstation/Fusion) and using native govc parameters for vSphere. (I'm the one who actually added it to govc after all because I ran into exactly this "Argument list too long" error 😅)

PR #520, opened 2023-03-28
network-configuration: Mention `.link` files

cgwalters:
While not everyone will want to do this, it has a lot of potential benefits.  One of them for example is that logging into a system and typing `ip` will inherently show more descriptive output.

The other main one though is that your system inherently becomes immune to naming drift across kernel and userspace changes.


dustymabe:
We have documentation for this already in https://docs.fedoraproject.org/en-US/fedora-coreos/customize-nic/#_using_a_systemd_link_file

PR #521, opened 2023-03-28
debugging-kernel-crashses: link to kexec manpage

jlebon:
To allow users to learn more if they'd like. And for consistency; we link to another manpage on that page already.


cgwalters (approved):

dustymabe:
Thanks for adding this!

PR #524, opened 2023-03-30
provisioning/vmware: showcase ability to use file references with govc

Okeanos:
As discussed in #517 there is another option when dealing with vSphere and govc to provide large Ignition configs by providing a file reference.

CC: @deed02392

### DIFF
--- a/modules/ROOT/pages/provisioning-vmware.adoc
+++ b/modules/ROOT/pages/provisioning-vmware.adoc
@@ -50,7 +50,25 @@ CONFIG_ENCODING='gzip+base64'
 CONFIG_ENCODED=$(cat example.ign | gzip -9 | base64 -w0 -)
 ----
 
-If your generated Ignition configuration is still too large, you will encounter an `Argument list too long` error or similar. The last resort is to use a configuration file. Instead of setting an environment variable containing your Ignition configuration, create an `ovftool` compatible configuration file in the directory you are invoking from like so:
+If your generated Ignition configuration is still too large, you will encounter an `Argument list too long` error or similar. The solution to that problem depends on whether you are working with vSphere or Workstation/Fusion.
+
+For vSphere the solution is simple because instead of inlining the configuration file within your shell, `govc` allows you to specify a path to a local file instead with the https://github.com/vmware/govmomi/blob/main/govc/USAGE.md#vmchange[`vm.change`]-command and will handle reading and writing it internally, circumventing any shell limitations.
+
+[source, bash]
+----
+CONFIG_ENCODING="gzip+base64"
+CONFIG_FILE="example.ign"
+CONFIG_FILE_ENCODED="${CONFIG_FILE}.gzip.b64"
+
+gzip -9c "${CONFIG_FILE}" | base64 -w0 - > "${CONFIG_FILE_ENCODED}"
+
+govc vm.change -vm "${VM_NAME}" -e "guestinfo.ignition.config.data.encoding=${CONFIG_ENCODING}"
+govc vm.change -vm "${VM_NAME}" -f "guestinfo.ignition.config.data=${CONFIG_FILE_ENCODED}" # using `-f` with a file path instead of `-e`
+----
+
+NOTE: Using `gzip` for this solution is optional and primarily used for consistent examples.
+
+In the case of Workstation/Fusion, or as a last resort in general, there is the option to use a configuration file. Instead of setting an environment variable containing your Ignition configuration, create an `ovftool` compatible configuration file in the directory you are invoking from like so:
 
 [source, bash]
 ----


cgwalters (approved):

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -50,7 +50,24 @@ CONFIG_ENCODING='gzip+base64'
...
+
+[source, bash]
+----
+CONFIG_ENCODING='gzip+base64'
+CONFIG_ENCODED=$(cat example.ign | gzip -9 | base64 -w0 -)
+CONFIG_FILE_ENCODED='example.ign.gzip.b64'
+echo "${CONFIG_ENCODED}" > "${CONFIG_FILE_ENCODED}"
+
+govc vm.change -vm "${VM_NAME}" -e "guestinfo.ignition.config.data.encoding=${CONFIG_ENCODING}"
+govc vm.change -vm "${VM_NAME}" -f "guestinfo.ignition.config.data=${CONFIG_FILE_ENCODED}" # using `-f` with a file path instead of `-e`
```
bgilbert, line 65: Oof, looks like all the examples here gratuitously `cat` things.  I don't think we need to make that worse though.

```suggestion
CONFIG_ENCODING="gzip+base64"
CONFIG_FILE="example.ign"
CONFIG_FILE_ENCODED="${CONFIG_FILE}.gz.b64"

gzip -9c "${CONFIG_FILE}" | base64 -w0 - > "${CONFIG_FILE_ENCODED}"
govc vm.change -vm "${VM_NAME}" -e "guestinfo.ignition.config.data.encoding=${CONFIG_ENCODING}"
govc vm.change -vm "${VM_NAME}" -f "guestinfo.ignition.config.data=${CONFIG_FILE_ENCODED}" # using `-f` with a file path instead of `-e`
```

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -50,7 +50,24 @@ CONFIG_ENCODING='gzip+base64'
...
+----
+
+NOTE: Using `gzip` for this solution is optional and primarily used for consistent examples.
+
+In case of Workstation/Fusion or as a last resort in general there is the option to use a configuration file. Instead of setting an environment variable containing your Ignition configuration, create an `ovftool` compatible configuration file in the directory you are invoking from like so:
```
bgilbert, line 70:
```suggestion
In the case of Workstation/Fusion, or as a last resort in general, there is the option to use a configuration file. Instead of setting an environment variable containing your Ignition configuration, create an `ovftool` compatible configuration file in the directory you are invoking from like so:
```

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -50,7 +50,25 @@ CONFIG_ENCODING='gzip+base64'
...
-If your generated Ignition configuration is still too large, you will encounter an `Argument list too long` error or similar. The last resort is to use a configuration file. Instead of setting an environment variable containing your Ignition configuration, create an `ovftool` compatible configuration file in the directory you are invoking from like so:
+If your generated Ignition configuration is still too large, you will encounter an `Argument list too long` error or similar. The solution to that problem depends on whether you are working with vSphere or Workstation/Fusion.
+
+For vSphere the solution is simple because instead of inlining the configuration file within your shell, `govc` allows you to specify a path to a local file instead with the https://github.com/vmware/govmomi/blob/main/govc/USAGE.md#vmchange[`vm.change`]-command and will handle reading and writing it internally, circumventing any shell limitations.
+
+[source, bash]
+----
+CONFIG_ENCODING="gzip+base64"
+CONFIG_FILE="example.ign"
+CONFIG_FILE_ENCODED="${CONFIG_FILE}.gzip.b64"
```
bgilbert, line 61: gzip produces `.gz` files, not `.gzip`, but not a big deal.

bgilbert (approved):
Thanks!

modules/ROOT/pages/provisioning-vmware.adoc
deed02392: That does seem like it could be a big deal, https://unix.stackexchange.com/questions/709708/why-gzip-utility-cares-about-extension

bgilbert: In general, sure.  In this case, the file isn't going to be passed back to `gunzip`, so the filename doesn't matter.

Okeanos: Uhhh, that may still be a good point because … how does FCOS/Ignition handle this when told it has to unpack `gzip+base64` via the encoding property but is served a `gz+base64` string instead?

bgilbert: Ignition only sees the contents, not the filename that `govc` read them from.

(Also, apparently Ignition is [pretty permissive](https://github.com/coreos/ignition/blob/0680c483fce356c3ddd2bcec913f41e847adc51e/internal/providers/vmware/vmware.go#L44) about how the encoding property is spelled.)

PR #525, opened 2023-03-30
adoc: clear explanation of cloud platform scope

sean-freeman:
For distinction, it is important to note that each hyperscaler Cloud Service Provider has some previous generation of Cloud Services and Networking environments. Misleading labels may cause confusion with end users, particularly as most Cloud Platforms have hardware generations for different Virtual Machines and Bare Metals using newer CPU and DRAM.

Below is the evidence for Cloud Service Provider versioning, where most products have suffix (Classic).

The PR includes an abbreviated version of the below content.

### AWS hyperscaler, Cloud Service Provider

- AWS EC2-Classic Networking environment (replaced by VPC networks environment). Deprecated in Aug-2022. See [AWS EC2-Classic documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-classic-platform.html).
- Amazon EC2 Previous Generation Instances. See [AWS EC2 Previous Generation Instances documentation](https://aws.amazon.com/ec2/previous-generation/).

### Google Cloud hyperscaler, Cloud Service Provider

- Google Cloud Legacy Networks (replaced by VPC Networks). Not available to provision. See [Google Cloud Legacy Networks documentation](https://cloud.google.com/vpc/docs/legacy).

### IBM Cloud hyperscaler, Cloud Service Provider

- IBM Cloud Classic Infrastructure environment, replaced by IBM Cloud VPC Infrastructure environment. See [IBM Cloud Classic Infrastructure compared with IBM Cloud VPC Infrastructure environments documentation](https://cloud.ibm.com/docs/cloud-infrastructure?topic=cloud-infrastructure-compare-infrastructure).
    - IBM Cloud Virtual Servers (Classic) based on Xen hypervisor; replaced by IBM Cloud Virtual Servers (with hardware generations) based on KVM hypervisor. See [IBM Cloud Virtual Servers (for Classic) documentation](https://cloud.ibm.com/docs/virtual-servers).
    - IBM Cloud VLAN (Classic) and VLAN Subnets (Classic); replaced by IBM Cloud VPC Networks and IBM Cloud VPC Subnets. See [IBM Cloud Classic VLANs documentation](https://cloud.ibm.com/docs/vlans?topic=vlans-about-vlans).

### Microsoft Azure hyperscaler, Cloud Service Provider

- Azure Service Manager (ASM) control plane (aka. environment), replaced by Azure Resource Manager (ARM) control plane.
    - Azure IaaS VM (Classic), managed by Azure Service Manager (ASM) control plane. Deprecation due Sept-2023. See [Azure Classic VM deprecation documentation](https://learn.microsoft.com/en-us/azure//virtual-machines/classic-vm-deprecation) and [Azure Resource Manager vs. Classic deployment models documentation](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/deployment-models).
        - Azure IaaS VM using Hyper-V Generation 1 format; replaced by Hyper-V Generation 2 format. See [Azure Classic VM API reference without Generation 2 support](https://learn.microsoft.com/en-us/previous-versions/azure/reference/jj157194(v=azure.100)) compared with [Azure VM API reference with Generation 2 support](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?tabs=HTTP).
    - Azure VNet (Classic), part of the Azure Cloud Services (Classic). Deprecation due Aug-2024. See [Azure Classic VNet documentation](https://learn.microsoft.com/en-us/previous-versions/azure/virtual-network/create-virtual-network-classic) and [Azure Cloud Services (Classic) documentation](https://learn.microsoft.com/en-us/azure/cloud-services/).
    - Azure Storage Accounts (Classic), part of the Azure Cloud Services (Classic). Deprecation due Aug-2024. See [Azure Classic Storage Account migration documentation](https://learn.microsoft.com/en-us/azure/storage/common/storage-account-migrate-classic) and [Azure Cloud Services (Classic) documentation](https://learn.microsoft.com/en-us/azure/cloud-services/).

### DIFF
--- a/modules/ROOT/pages/provisioning-ibmcloud.adoc
+++ b/modules/ROOT/pages/provisioning-ibmcloud.adoc
@@ -1,6 +1,8 @@
 = Provisioning Fedora CoreOS on IBM Cloud
 
-This guide shows how to provision new Fedora CoreOS (FCOS) instances in IBM Cloud VPC Generation 2 for either the `x86_64` or `s390x` architectures.
+This guide shows how to provision new Fedora CoreOS (FCOS) instances in IBM Cloud for either the `x86_64` or `s390x` architectures.
+
+NOTE: This guide describes IBM Cloud infrastructure services, specifically IBM Cloud Virtual Servers using VPC networking (also known as IBM Cloud VPC Infrastructure environment). It does not describe IBM Cloud Virtual Server (Classic), using the IBM Cloud Classic Infrastructure environment.
 
 == Prerequisites
 
@@ -33,7 +35,7 @@ ibmcloud resource group-create $RESOURCE_GROUP # Create the resource group if it
 ibmcloud target -g $RESOURCE_GROUP
 ----
 
-There are also several other pieces that need to be in place, like a VPC, SSH keys, networks, permissions, etc. Unfortunately, this guide is not a comprehensive IBM Cloud guide. If you are new to IBM Cloud please familiarize yourself using https://cloud.ibm.com/docs/vpc?topic=vpc-getting-started[the documentation for VPC Gen2] first.
+There are also several other pieces that need to be in place, like a VPC, SSH keys, networks, permissions, etc. Unfortunately, this guide is not a comprehensive IBM Cloud guide. If you are new to IBM Cloud please familiarize yourself using https://cloud.ibm.com/docs/vpc?topic=vpc-getting-started[the documentation for IBM Cloud VPC networks and related infrastructure services] first.
 
 === Creating an Image
 
--- a/modules/ROOT/pages/provisioning-ibmcloud.adoc
+++ b/modules/ROOT/pages/provisioning-ibmcloud.adoc
@@ -2,7 +2,7 @@
 
 This guide shows how to provision new Fedora CoreOS (FCOS) instances in IBM Cloud for either the `x86_64` or `s390x` architectures.
 
-NOTE: This guide does not describe Fedora CoreOS (FCOS) with https://cloud.ibm.com/docs/virtual-servers[IBM Cloud Virtual Servers (Classic)], using https://cloud.ibm.com/docs/vlans?topic=vlans-about-vlans[IBM Cloud VLAN (Classic) and VLAN Subnets (Classic)] within the https://cloud.ibm.com/docs/cloud-infrastructure?topic=cloud-infrastructure-compare-infrastructure[IBM Cloud Classic Infrastructure environment]. The guide exclusively refers to Fedora CoreOS (FCOS) with IBM Cloud infrastructure services, also known as IBM Cloud VPC Infrastructure environment.
+NOTE: Fedora CoreOS (FCOS) is not supported with https://cloud.ibm.com/docs/virtual-servers[IBM Cloud Virtual Servers (Classic)], using https://cloud.ibm.com/docs/vlans?topic=vlans-about-vlans[IBM Cloud VLAN (Classic) and VLAN Subnets (Classic)] within the https://cloud.ibm.com/docs/cloud-infrastructure?topic=cloud-infrastructure-compare-infrastructure[IBM Cloud Classic Infrastructure environment]. The guide exclusively refers to Fedora CoreOS (FCOS) with IBM Cloud infrastructure services, also known as IBM Cloud VPC Infrastructure environment.
 
 == Prerequisites
 
--- a/modules/ROOT/pages/provisioning-ibmcloud.adoc
+++ b/modules/ROOT/pages/provisioning-ibmcloud.adoc
@@ -2,7 +2,7 @@
 
 This guide shows how to provision new Fedora CoreOS (FCOS) instances in IBM Cloud for either the `x86_64` or `s390x` architectures.
 
-NOTE: This guide describes IBM Cloud infrastructure services, specifically IBM Cloud Virtual Servers using VPC networking (also known as IBM Cloud VPC Infrastructure environment). It does not describe IBM Cloud Virtual Server (Classic), using the IBM Cloud Classic Infrastructure environment.
+NOTE: This guide does not describe Fedora CoreOS (FCOS) with https://cloud.ibm.com/docs/virtual-servers[IBM Cloud Virtual Servers (Classic)], using https://cloud.ibm.com/docs/vlans?topic=vlans-about-vlans[IBM Cloud VLAN (Classic) and VLAN Subnets (Classic)] within the https://cloud.ibm.com/docs/cloud-infrastructure?topic=cloud-infrastructure-compare-infrastructure[IBM Cloud Classic Infrastructure environment]. The guide exclusively refers to Fedora CoreOS (FCOS) with IBM Cloud infrastructure services, also known as IBM Cloud VPC Infrastructure environment.
 
 == Prerequisites
 
--- a/modules/ROOT/pages/platforms.adoc
+++ b/modules/ROOT/pages/platforms.adoc
@@ -15,7 +15,7 @@ The currently supported platforms and their identifiers are listed below.
 * DigitalOcean (`digitalocean`): Cloud platform. See xref:provisioning-digitalocean.adoc[Booting on DigitalOcean].
 * Exoscale (`exoscale`): Cloud platform. See xref:provisioning-exoscale.adoc[Booting on Exoscale].
 * Google Cloud Platform (`gcp`): Cloud platform. See xref:provisioning-gcp.adoc[Booting on GCP].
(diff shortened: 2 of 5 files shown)


bgilbert:
It seems like this PR is mainly trying to avoid the appearance that only some IBM Cloud VM services are supported.  As far as I know, IBM Cloud still offers some VM services that are not supported by Fedora CoreOS.  Is this no longer correct?

As to other platforms, it's correct that Azure ASM is not supported, but Azure ASM is no longer available.  For AWS and GCP, as far as I know, Fedora CoreOS supports all of the older-generation VM technologies.  (S3-backed EC2 disks might not work, but our images are configured to use EBS, so that wouldn't be user-facing.)

sean-freeman:
The PR is trying to address parity in references to Cloud Service Providers and their associated technologies/services. There is no intention to disguise appearances, quite the opposite - it is to ensure end users understand the limitations of running Fedora CoreOS on any given cloud platform provider.

Because there are no limitations/scope defined, at present it could be reasonably expected by an end user to attempt Fedora CoreOS installations to (and fail because there is no Afterburn provider code handlers for the following):
- Azure IaaS VM (Classic) hosts. As cited ASM is still alive and well until both Azure IaaS VM (Classic) is deprecated (see end of 2023 date above) and Azure Storage Accounts (Classic) is deprecated (see end of 2024 date above)
- Google Cloud Legacy Networks. No deprecation date given

In reference specifically to Virtual Machine IaaS products from IBM Cloud, both are and remain supported for use by Fedora CoreOS:
- IBM Cloud Virtual Server, Ignition Platform ID `ibmcloud`.
- IBM Cloud Virtual Server (Classic), Ignition Platform ID `ibmcloud-classic`.

bgilbert:
Ignition doesn't support `ibmcloud-classic` and Fedora CoreOS doesn't ship images for it.  I had missed that Azure ASM is still around, and am okay with adding a note to the docs about it.  I'm pretty sure Google Cloud legacy networking will work with Fedora CoreOS; could you point out the specific functionality that will fail?

However, if the goal is to help users, an `Important note` with a bulleted list of text _that will be irrelevant to most users_ isn't the way to do it.  We try to keep our docs readable by keeping them clear, useful, and concise.  You're right that the list of platform identifiers is probably not the place to mention support limitations, so I'm okay with dropping `, VPC Generation 2` from `platforms.adoc`.  Other than that, a sentence or two in the first paragraph of an individual platform page is probably the way to go.  For example, the Azure page might say `Legacy Azure Service Manager virtual machines are not supported.`, with a link.  Similarly, if the pre-Gen2 IBM Cloud VMs are considered fully legacy, it might be reasonable to drop the "Generation 2" language, and instead have an additional sentence naming the execution environments that aren't supported.

sean-freeman:
Hard to point to specific functionality without attempting to run CoreOS in every scenario (test suite style), and I don't have the will to prove negative test results for past functionality - just being truthful :)

- Re. `ibmcloud-classic` there are no OS Images shipped for this, but the code switches exist such as [`afterburn agent ibmcloud-classic`](https://github.com/coreos/afterburn/blob/main/src/providers/ibmcloud_classic/mod.rs) and it exists in various places within documentation under GH Orgs /coreos, /afterburn, /openshift .
- Re. GCP Legacy Networks - If I recall correctly, hosts provisioned with GCP Legacy Networks and still running today would only have access to the legacy metadata service (deprecated). I don't profess to be an expert on these matters, and from a cursory search cannot find any further documentation on that topic. In my daily work life it has been important to specifically identify the cloud services as end users are not always aware of what they use.
- Re. IBM Cloud `, VPC Generation 2` suffix I just cleaned that up at the same time because someone mixed the terminologies in this documentation (seems OK in the OpenShift Installer GitHub). It should refer as "IBM Cloud" on it's own and referring to the previous environment with the suffix "Classic" -  that would be a consistent/correct reference in-line with the other cloud platform references.
    - The Generation refers to underlying hardware, for example Gen2 was primarily with Intel Cascade Lake CPUs and DDR4 and can be seen within the profile naming such as `mx2-` prefix and for Gen3 this would be `mx3-`. Same thing as AWS EC2 Instance Types iterating as `r4.` > `r5.` > `r6.` for hardware generations (citation: [Introducing Amazon EC2 R5 Instances, the next generation of memory-optimized instances](https://aws.amazon.com/about-aws/whats-new/2018/07/introducing-amazon-ec2-r5-instances/)). Throughout the rest of the docs there is no reference to hardware generations on other cloud platforms, so I removed the incorrect suffix to keep parity.

I'm OK with condensing further and putting each one liner in the respective platform's specific doc. I'll amend and commit.

sean-freeman:
Re. IBM Cloud - I wouldn't say the marketing team did a great job in that blog post, although if you read it - it does make sense. It's talking about Gen1 hardware > Gen2 hardware will provide faster provisioning, faster network interfaces, compute auto scaling etc. There is no "VPC Gen1" as such, it would be same as on AWS as saying "VPC R6" instead of "Classic R2". It should never have been interpreted as a flat list, it was and remains:

- IBM Cloud Classic Infrastructure environment (aka. Softlayer)
    - Virtual Server (Classic); no specified generations
    - Bare Metal (Classic); no specified generations
- IBM Cloud (aka. VPC Infrastructure environment)
    - Virtual Server, Gen1 hardware `Deprecated`
    - Virtual Server, Gen2 hardware
    - Bare Metal, Gen2 hardware
    - etc for future generations

The PR is primarily to create parity and ensure it is understood the scoping of CoreOS compatibility with predecessor service/feature/capability of each cloud platform. We've diverted down the IBM Cloud path a bit too much, and all of this was a lot easier to scope with traditional Hypervisors - just a software name and version number!

bgilbert:
> Re. `ibmcloud-classic` there are no OS Images shipped for this, but the code switches exist such as [`afterburn agent ibmcloud-classic`](https://github.com/coreos/afterburn/blob/main/src/providers/ibmcloud_classic/mod.rs) and it exists in various places within documentation under GH Orgs /coreos, /afterburn, /openshift .

Certain upstreams have code for `ibmcloud-classic`, yes.  But if Fedora CoreOS doesn't ship an image configured with that platform ID, the code won't be invoked.  Similarly, the existence of OpenShift documentation isn't necessarily relevant to what's available in Fedora CoreOS.  Fedora CoreOS does not support IBM Cloud Classic.

> Re. GCP Legacy Networks - If I recall correctly, hosts provisioned with GCP Legacy Networks and still running today would only have access to the legacy metadata service (deprecated).

Ignition only runs on first boot, and the Afterburn code dates to 2017.  We haven't knowingly removed any Afterburn support for the legacy GCP service.  And even if it were to fail, instances would continue to function.

bgilbert (changes requested):
You have more context than I do for the IBM Cloud offerings, so I'll defer to your expertise for the proper way to describe them.

The other clouds are a different matter.  To be direct with you, I view the qualifications added on the other cloud pages as FUD rather than an attempt to inform.  Concrete and accurate statements of the form `X is not supported`, as proposed in https://github.com/coreos/fedora-coreos-docs/pull/525#issuecomment-1491128579, are welcome and I'd be happy to review them.  However, language like "this guide does not describe" insinuates that there might be problems with the things not described, without actually making any claims about the boundaries of their support.  I will not approve a PR that includes such language, including on the IBM Cloud page.

sean-freeman:
@bgilbert Altered commit with the prefix amended to your suggestion, my previous attempt was to be consistent with the opening paragraph ("this guide shows...").

modules/ROOT/pages/provisioning-ibmcloud.adoc
```diff
@@ -1,6 +1,8 @@
 = Provisioning Fedora CoreOS on IBM Cloud
 
-This guide shows how to provision new Fedora CoreOS (FCOS) instances in IBM Cloud VPC Generation 2 for either the `x86_64` or `s390x` architectures.
+This guide shows how to provision new Fedora CoreOS (FCOS) instances in IBM Cloud for either the `x86_64` or `s390x` architectures.
```
bgilbert, line 3: For the record, [this post](https://www.ibm.com/cloud/blog/announcements/start-your-vpc-gen1-to-vpc-gen2-migration) says that VPC Generation 1 EOLed in 2021, so we can indeed drop the `Generation 2` suffix.

bgilbert (commented):
The proposed language still needs to be simplified and clarified to match the tone of our docs.  Rather than asking you for additional rounds of changes, it seemed easier to open a fresh PR with the desired text, so I've done that in #533 and credited you as a coauthor on the IBM Cloud portion.  Feel free to comment there.  I'll go ahead and close this.

PR #526, opened 2023-04-01
provisioning/vmware: simplify example, provide authentication tip

Okeanos:
As discussed in coreos/fedora-coreos-tracker#1158 the examples for VMware vSphere now:

- tell users about alternative authentication methods
- give simpler instructions for a hello-world run
  - reuse of library items is relegated to an optional section

### DIFF
--- a/modules/ROOT/pages/provisioning-vmware.adoc
+++ b/modules/ROOT/pages/provisioning-vmware.adoc
@@ -110,39 +110,20 @@ ssh core@<ip address>
 
 This section shows how to use vSphere facilities to configure and run VMs from the command-line. Similar steps can be performed via the graphical UI too.
 
-=== Importing the OVA
+TIP: While the examples below use `govc session.login` to authenticate, you can also use environment variables to provide credentials. Check the https://github.com/vmware/govmomi/tree/main/govc#usage[official documentation] for details.
 
-The downloaded OVA has to be first imported into vSphere library:
+=== Setting up a new VM
+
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -92,9 +92,40 @@ ssh core@<ip address>
 
 This section shows how to use vSphere facilities to configure and run VMs from the command-line. Similar steps can be performed via the graphical UI too.
 
-=== Importing the OVA
+TIP: While the examples below use `govc session.login` to authenticate, you can also use environment variables to provide credentials. Check the https://github.com/vmware/govmomi/tree/main/govc#usage[official documentation] for details.
```
Okeanos, line 95: URL could also be set to `https://github.com/vmware/govmomi/blob/main/govc/README.md#usage` … not sure whether it makes much of a difference?

dustymabe:
Looks like @lucab originally wrote this documentation. Anybody else familiar with VMWare enough to review this?

cc @mike-nguyen @bgilbert

fifofonix (approved):
Having just re-executed these steps I can confirm that the issues that precipitated this suggested change still exist. 

The suggested changes work for me and I concur with the new primary flow which removes the library import and therefore requires lesser privileges on vSphere.

jlebon (approved):
Makes sense to me.
Thanks for writing this!

jlebon:
And thanks @fifofonix for testing and providing feedback!

dustymabe:
Thanks @fifofonix @Okeanos !

Issue #534, opened 2023-04-23
[main] Doc issue in file modules/ROOT/pages/tutorial-updates.adoc


marcosgm:
The link to discover older releases is broken,  https://fedoraproject.org/en/coreos/?stream=stable now returns 404
It should point to the Release Notes instead https://fedoraproject.org/coreos/release-notes/?arch=x86_64&stream=stable

marcosgm:
Same thing for the GPG key, no longer in https://getfedora.org/static/fedora.gpg, it's now in https://fedoraproject.org/fedora.gpg

PR #535, opened 2023-04-23
Update Tutorial with new fedora URLs

marcosgm:
for older releases and GPG

Closes: https://github.com/coreos/fedora-coreos-docs/issues/534


jlebon (approved):
Thanks!

PR #537, opened 2023-04-27
Sync with upstream template

travier:
See: https://gitlab.com/fedora/docs/templates/fedora-docs-template


jlebon (approved):
LGTM assuming the latest steps were sanity-checked to work well with this repo. Should the commit message include the commit hash we synced with?

travier:
I'll have to double check that indeed.

PR #539, opened 2023-04-29
Update provisioning-qemu.adoc

MichaelOnSecurity:
Without the "\" the next line is ignored and thus is port forwarding is not enabled


cgwalters (approved):

cgwalters:
Thanks!

PR #540, opened 2023-05-01
Add a doc for container provisioning and updates

cgwalters:
The layering model is an entirely new way to do systems management.  Let's document the current state.

### DIFF
--- a/modules/ROOT/pages/deriving-container.adoc
+++ b/modules/ROOT/pages/deriving-container.adoc
@@ -1,0 +1,127 @@
+= Deriving from Fedora CoreOS as a bootable container image
+
+NOTICE: At the current time, the suggested model here does not integrate with the current "update barriers" model used for default Fedora CoreOS updates.  While relatively rare so far, this means that some container-based updates may see issues that were otherwise gated.  This issue is being worked on.  More information in https://github.com/coreos/fedora-coreos-tracker/issues/1263[this issue].
+
+== Bootable containers overview
+
+The operating system images used for Fedora CoreOS are available as a standard container image; there is one per stream. See xref:update-streams.adoc[Update Streams].
+
+- quay.io/fedora/fedora-coreos:stable
+- quay.io/fedora/fedora-coreos:next
+- quay.io/fedora/fedora-coreos:testing
+
+== Creating custom builds
+
+See the https://github.com/coreos/layering-examples[layering examples] git repository.  While the examples there tend to use Containerfile/Dockerfile style builds, you can use any tool which can build containers and push the result to a registry.
+
+The default suggested approach is to have container tags be the source of truth for initating upgrades, and to have the client systems poll.  Example systemd units for this are in [https://github.com/coreos/layering-examples/tree/main/autoupdate-host]; they boil down to a systemd `.service` which just runs `rpm-ostree upgrade --reboot` and a corresponding `.timer` unit to run it once a day.
+
+Of course, client side logic could be much more complex, including a privileged container image that runs completely arbitrary logic.
+
+== Using Ignition to switch to a bootable container image
+
+The https://coreos.github.io/rpm-ostree/container/[rpm-ostree container page] describes the commands for interacting with bootable ostree container images.
+
+This systemd unit will run only on the first boot, and switch a "stock" Fedora CoreOS node to your custom image.
+
+[source,yaml]
+----
+variant: fcos
+version: 1.4.0
+systemd:
+  units:
+    # Our custom unit for rebasing
+    - name: rebase-custom.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Fetch and deploy target image
+        # Only run on the firstboot
+        ConditionFirstBoot=true
+        After=network-online.target
+        [Service]
+        # This ordering is important
+        After=ignition-firstboot-complete.service
+        Type=oneshot
+        RemainAfterExit=yes
+        ExecStart=rpm-ostree rebase --reboot ostree-unverified-registry:quay.io/example/example-derived:latest
+        [Install]
+        WantedBy=multi-user.target
+----
+
+== Adding pull secrets for private container images
+
+If your registry requires authentication, then you can add a pull secret.  See https://coreos.github.io/rpm-ostree/container/#registry-authentication[container pull secrets].
+
+[source,yaml]
+----
+variant: fcos
+version: 1.4.0
+storage:
+  files:
+    - path: /etc/ostree/auth.json
+      mode: 0600
+      contents:
+        inline: >
+              {
+                   "auths": {
+                        "quay.io": {
+                             "auth": "..."
+                        }
+                   }
+              }
+----
+
+== Maintaining custom builds
+
+As the upstream container image (e.g. quay.io/fedora/fedora-coreos:stable) changes for security and functionality improvements, you are responsible for re-building your derived image.  Implementing this depends on your chosen container build system.
+
+However, key to this model is the assumption that if you're deploying Fedora CoreOS, you have already invested in container infrastructure and can manage the operating system updates in the same way as application containers.
+
+== Understanding Ignition versus container content
+
+=== Storage
+
+First, if you wish to configure storage (see xref:storage.adoc[Configuring Storage] ) that must still be done via the initial Ignition configuration. 
+If you wish to change a system's storage configuration, the current recommendation is still that that be done via full system reprovisioning.
+
+=== Filesystem
+
+Ignition's philosophy is that it runs exactly once, and any subsequent changes should generally be done by node reprovisioning.  This can
+be very expensive in some environments.  Storing configuration directly
+with the OS content allows strongly coupling them and ensuring that
+the state of your machines is exactly what you build and replicate
+as a container image.
+
+Note however that content injected via Ignition will still persist as
+machine-local state into subsequent OS updates (including by default
+our unit to perform the switch, which will remain inactive; though your
+container could run code to delete or mask the unit).
+
+=== Machine local state
+
+A powerful ability of using OCI containers as a mechanism for host
+systems management is that one can create chains of derived images;
+for example, you might have a "base" image derived from Fedora CoreOS
+that include e.g. basic agents or logging configuration, but then further builds
+that customize that base for particular sites or cloud environments.
+
+It's even possible to create builds which are targeted solely for
+a specific machine.
+
+However, this may not always be ergonomic to do.  Ignition here
+can serve as a mechanism to apply machine-specific state.  Commonly,
+this machine specific state does not need to change often.  A classic
+example here is configuring `/etc/hostname`.
+
+=== Bootstrap configuration to fetch container images
+
+The pull secret described above is also a form of configuration on
+the system that must be present (in some environments) to pull the container
+image, and hence must be configured via Ignition today.
+
+==== Machine local bootstrap configuration
+
+There is a class of configuration which combines these two previous types,
+such as static IP addressing.  Here, Ignition again makes sense
+to perform this configuration.
--- a/modules/ROOT/pages/getting-started.adoc
+++ b/modules/ROOT/pages/getting-started.adoc
@@ -19,7 +19,7 @@ Fedora CoreOS does not have a separate install disk. Instead, every instance sta
 
 Each platform has specific logic to retrieve and apply the first boot configuration. For cloud deployments, Ignition gathers the configuration via user-data mechanisms. In the case of bare metal, Ignition can fetch its configuration from the disk or from a remote source.
 
-For more information on configuration, refer to the documentation for xref:producing-ign.adoc[Producing an Ignition File].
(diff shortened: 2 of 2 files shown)


modules/ROOT/pages/deriving-container.adoc
```diff
@@ -0,0 +1,94 @@
+= Deriving from Fedora CoreOS as a bootable container image
+
+== Bootable containers overview
+
+The operating system images used for Fedora CoreOS are available as a standard container image; there is one per stream:
+
+- quay.io/fedora/fedora-coreos:stable
+- quay.io/fedora/fedora-coreos:next
```
dustymabe, line 8: might as well just list them all since there's only one more (`testing`).

modules/ROOT/pages/deriving-container.adoc
```diff
@@ -0,0 +1,94 @@
...
+
+etc.
+
+== Creating custom builds
+
+See the https://github.com/coreos/layering-examples[layering examples] git repository.  While the examples there tend to use Container/Dockerfile style builds, you can use any tool which can build containers and push the result to a registry.
```
dustymabe, line 14: `s/Container/Containerfile`

modules/ROOT/pages/deriving-container.adoc
```diff
@@ -0,0 +1,94 @@
...
+
+== Using Ignition to retarget an instance
```
dustymabe, line 16:
```suggestion
== Using Ignition to rebase an instance
```

Maybe rebase instead?

modules/ROOT/pages/deriving-container.adoc
```diff
@@ -0,0 +1,94 @@
...
+    # Our custom unit for automatic upgrades
+    - name: autoupdate-host.service
+      enabled: false
+      contents: |
+        [Unit]
+        Description=Automatic host upgrades
+        [Service]
+        Type=simple
+        ExecStart=rpm-ostree upgrade --reboot
+        StandardOutput=null
```
dustymabe, line 55: `StandardOutput=null` seems like an odd thing to do. Why?

modules/ROOT/pages/deriving-container.adoc
```diff
@@ -0,0 +1,94 @@
...
+        [Timer]
+        OnBootSec=1h
+        OnUnitInactiveSec=1d
+        [Install]
+        WantedBy=timers.target
+----
+
+== Adding pull secrets for private container images
+
+See https://coreos.github.io/rpm-ostree/container/#registry-authentication[container pull secrets].
```
dustymabe, line 74:
```suggestion
If your registry repo requires authentication you can add a pull secret to the OSTree configuration. See https://coreos.github.io/rpm-ostree/container/#registry-authentication[container pull secrets].
```

modules/ROOT/pages/deriving-container.adoc
```diff
@@ -0,0 +1,94 @@
...
+      contents:
+        inline: >
+              {
+                   "auths": {
+                        "quay.io": {
+                             "auth": "..."
+                        }
+                   }
+              }
+      mode: 0600
```
dustymabe, line 92: I would put `mode` under `path`.

dustymabe (commented):
Some comments. I'm still not sure we should doc this (we were making progress on https://github.com/coreos/fedora-coreos-tracker/issues/1263 but then f38 and rhel9 fires took us away) yet.

If we do I think we should call out more clearly the need to implement your own build system and the loss of managed autoupdates (update graph). The autoupdates part is kind of implied because of the documented update unit/timer, but we should make it clearer.

modules/ROOT/pages/deriving-container.adoc
cgwalters: Because otherwise you get many log messages to the journal twice (once from the client, once from the daemon).

modules/ROOT/pages/deriving-container.adoc
cgwalters: While I love the term "rebase" myself...it feels obscure because it's a thing not present in docker/podman/k8s.

modules/ROOT/pages/deriving-container.adoc
cgwalters: nit: I don't think "registry repo" is a thing, there's just "registry".  The term "repo" is already overloaded.

cgwalters:
> I'm still not sure we should doc this (we were making progress on https://github.com/coreos/fedora-coreos-tracker/issues/1263 

I commented on the issue.  Indeed, documenting this is tantamount to support.  But I think the state of things will be pretty clear to users from the docs.

> If we do I think we should call out more clearly the need to implement your own build system and the loss of managed autoupdates (update graph). The autoupdates part is kind of implied because of the documented update unit/timer, but we should make it clearer.

I think automatic updates are still present, it just requires slightly more work.  But I believe that anyone who is doing anything nontrivial with a FCOS like system will already be invested in systems management infrastructure, and specifically using containers.

modules/ROOT/pages/deriving-container.adoc
dustymabe: Yeah. I think of "registry" like the TLD (i.e. `quay.io`) and the "repo" being the specific repository (i.e. `quay.io/fedora/fedora-coreos`). In this case `quay.io` doesn't necessarily require auth, but `quay.io/fedora/fedora-coreos` could.

I'm fine with just using "registry" here, just wanted to explain why the distinction in my head lead me towards "registry repo".

modules/ROOT/pages/deriving-container.adoc
dustymabe: Maybe a comment here would be useful 
```
# Prevent duplicate logging from daemon and this unit
```

modules/ROOT/pages/deriving-container.adoc
jlebon, line 1: Shouldn't this be linked from the nav tree? Not sure where it'd belong... Going along with the experimental labeling comment, maybe easiest is to have a new "Experimental features" parent with this as the first child?

jlebon (commented):
Documenting it as is today seems fine but given the high friction and shortcomings, let's label it as experimental?

modules/ROOT/pages/deriving-container.adoc
cgwalters: Or, to flip it around - in your view, what would be the criteria to "graduate"?

jlebon: It just doesn't seem to me like we've worked through all the details. Container layering impacts multiple parts of FCOS we're deeply opinionated about. We started on this in e.g. https://github.com/coreos/fedora-coreos-tracker/issues/1219 and https://github.com/coreos/fedora-coreos-tracker/issues/1263 but then... fires happened.

At a minimum, I think we should resolve https://github.com/coreos/fedora-coreos-tracker/issues/1263. But also, I think we would need a larger rework of what tests we run and how the docs are structured so it's properly integrated in our provisioning and configuration story. (But ideally, we also introduce a better UX for this stuff.)

How about something like this near the top:

> NOTE: Container layering is a new approach to provisioning and configuring Fedora CoreOS. The underlying features are considered stable, but its integration into Fedora CoreOS are subject to change and nodes may in the future require reprovisioning. Note that in this mode, automatic updates are *not* directly managed by the Fedora CoreOS team and rely on user-managed services.

?

cgwalters: > It just doesn't seem to me like we've worked through all the details.

Yeah but, speaking bluntly I feel like that's not going to happen unless I keep pushing it, hence this PR.

> At a minimum, I think we should resolve https://github.com/coreos/fedora-coreos-tracker/issues/1263.

OK, I have my opinion written there.


> but also, I think we would need a larger rework of what tests we run

That's true, filed https://github.com/coreos/fedora-coreos-tracker/issues/1484 

> But ideally, we also introduce a better UX for this stuff.)

That's partly in https://github.com/coreos/butane/issues/428

> and nodes may in the future require reprovisioning. 

Mmmm.  I guess that's a pivotal decision here.  I am not forseeing any changes which would require reprovisioning.  Are you?

>     Note that in this mode, autom1tic updates are not directly managed by the Fedora CoreOS team and rely on user-managed services.

This part is covered below there too, but sure we can emphasize it.  I tossed up this page in https://hackmd.io/QM1V-FujTmalikgi5JQHPw to avoid round trips.
(yeah it's not actually markdown sadly, but I'm just using it as a realtime editor)

bgilbert: > > It just doesn't seem to me like we've worked through all the details.
>
> Yeah but, speaking bluntly I feel like that's not going to happen unless I keep pushing it, hence this PR.

You're welcome to help work through the details, too.  :slightly_smiling_face:

cgwalters:
> and nodes may in the future require reprovisioning.

Breaking this one out to the toplevel because I disagree with it the most.  Everything done on the (rpm-)ostree side for this has been with an eye to compatibility and making things a seamless transition.  Many people are using it for various use cases, from their personal desktops and it shipped in OpenShift 4.12.

Saying that reprovisioning may be required is in fundamental conflict with that.  I personally am signed up and committed to debug any problems from this that may arise in the future.  I don't understand why others on the team don't feel the same way.

cgwalters:
> but its integration into Fedora CoreOS are subject to change 

Also of course, merging this is effectively saying that's not the case; the butane configuration here is proposed to work into the forseeable future.

The auto-update unit is pretty lame, but definitely functional.   I've got some PRs up to try to clean this up and more directly support it in https://github.com/coreos/rpm-ostree/pull/4392 but I wouldn't consider that a blocker, because ultimately the users pursuing this path are already in a position where they likely want to "own" more of the OS update logic, so these sample units are just a starting point.  In effect actually layering is kind of saying that things like the [zincati remote locking](https://github.com/coreos/zincati/blob/main/docs/usage/updates-strategy.md#lock-based-strategy) API is instead something that can be just owned by 3rd party agent logic; we provide lower level tools.  This also relates to https://github.com/coreos/zincati/pull/904

So the integration will certainly hopefully *improve*, from my PoV, but I would not say "change" as in "possibly breaking change".

modules/ROOT/pages/deriving-container.adoc
cgwalters: OK, I added a note to the top here which covers what is IMO the top last issue we considered "blocking" around the barriers.  It's just up front about that outstanding bug, but I'm hopeful we'll have more infrastructure there by the time we see the need for another barrier.

cgwalters:
OK, fleshed out the "When to use Ignition" section even more!

cgwalters:
I reworked https://github.com/coreos/fedora-coreos-tracker/issues/1363 to be a tracker that includes this PR

travier:
From my perspective, this is blocked on https://github.com/coreos/fedora-coreos-tracker/issues/1367

cgwalters:
> From my perspective, this is blocked on https://github.com/coreos/fedora-coreos-tracker/issues/1367

*blocked* is IMO a strong term.  Why is it blocked?

That's a pretty easy thing for someone who is not us to do on their own infrastructure.  In fact I'd say supporting "mirror the base image to a custom registry and control it using tooling you know" is really a lot of the point of this.

travier:
For me this is really blocking any real usage of ostree native containers for FCOS.

Users have to be able to test / run previous releases, use Git Ops to test their updates, use fixed versions in their CI, etc.

navaati:
Hi. @cgwalters in https://github.com/coreos/rpm-ostree/pull/4392 you said:
> This is aiming to help replace the hacky systemd unit created in https://github.com/coreos/fedora-coreos-docs/pull/540

I have no idea how it relates, now that the aforementioned PR is merged can you update this one to make use of it ? I may very well missing something :).

Also I see that https://github.com/coreos/butane/issues/428 is closed, does that mean that the idea of easily using that feature from ignition is abandoned ? What would the end-game UX look like ?

Thanks y’all FCOS team for carrying the dream of immutable infrastructure !

modules/ROOT/pages/deriving-container.adoc
```diff
@@ -0,0 +1,127 @@
...
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Fetch and deploy target image
+        # Only run on the firstboot
+        ConditionFirstBoot=true
+        After=network-online.target
+        [Service]
+        # This ordering is important
+        After=ignition-firstboot-complete.service
```
lawrencegripper, line 44: I think this is now `coreos-ignition-firstboot-complete.service`

lawrencegripper: image

From 

```
[core@forgejo-actionsrunner ~]$ cat /etc/os-release 
NAME="Fedora Linux"
VERSION="39.20240210.3.0 (CoreOS)"
ID=fedora
VERSION_ID=39
```

jlebon:
This is useful docs, though was written before https://docs.fedoraproject.org/en-US/bootc/ existed. A lot of the bits here are actually FCOS specific and how it interacts with fedora-bootc, so I think there's things to keep here. But we'll have a separate PR instead that puts it in context with the now extant fedora-bootc.

PR #552, opened 2023-05-20
lint: fix shellcheck findings in embedded shell scripts

Okeanos:
- prevent wordsplitting and accidental globbling by quoting variables
- use parameter expansion with braces for consistency
- remove a number of useless cat-expressions
- use printf with format string syntax

vmware:
- use correct gzip file ending (See https://github.com/coreos/fedora-coreos-docs/pull/524#discussion_r1159476263 for details) in example

### DIFF
--- a/modules/ROOT/pages/provisioning-azure.adoc
+++ b/modules/ROOT/pages/provisioning-azure.adoc
@@ -21,8 +21,8 @@ Once you have picked the relevant stream, download, verify, and decompress the l
 
 [source, bash]
 ----
-stream="stable"
-coreos-installer download --decompress -s "${stream}" -p azure -f vhd.xz
+STREAM="stable"
+coreos-installer download --decompress -s $STREAM -p azure -f vhd.xz
 ----
 
 Alternatively, you can manually download an Azure image from the https://fedoraproject.org/coreos/download/?stream=stable#cloud_images[download page]. Verify the download, following the instructions on that page, and decompress it.
--- a/modules/ROOT/pages/provisioning-exoscale.adoc
+++ b/modules/ROOT/pages/provisioning-exoscale.adoc
@@ -22,7 +22,7 @@ Exoscale provides https://community.exoscale.com/documentation/compute/custom-te
 [source, bash]
 ----
 STREAM="stable"
-coreos-installer download -d -s "${STREAM}" -p exoscale -f qcow2.xz
+coreos-installer download -d -s $STREAM -p exoscale -f qcow2.xz
 ----
 
 Alternatively, QCOW2 images can be downloaded from the https://fedoraproject.org/coreos/download/?stream=stable#cloud_images[download page] and manually decompressed.
@@ -35,15 +35,15 @@ Next you can https://community.exoscale.com/documentation/compute/custom-templat
 # Set the version and calcuate the checksum
 FCOS_VERSION='...'
 FILE="fedora-coreos-${FCOS_VERSION}-exoscale.x86_64.qcow2"
-CHECKSUM=$(md5sum $FILE | cut -d " " -f 1)
+CHECKSUM=$(md5sum "${FILE}" | cut -d " " -f 1)
 
 # Upload to object storage
 BUCKET='newbucket'
 exo storage mb "sos://${BUCKET}"
-exo storage upload --acl public-read $FILE "sos://${BUCKET}/image-import/"
+exo storage upload --acl public-read "${FILE}" "sos://${BUCKET}/image-import/"
 
 # Create the template using given URL and CHECKSUM
-URL=$(exo storage show "sos://${BUCKET}/image-import/$FILE" --output-template "{{.URL}}")
+URL=$(exo storage show "sos://${BUCKET}/image-import/${FILE}" --output-template "{{.URL}}")
 TEMPLATE="fedora-coreos-${FCOS_VERSION}"
 exo compute instance-template register --boot-mode=uefi $TEMPLATE $URL $CHECKSUM
 ----
@@ -72,9 +72,9 @@ TYPE='standard.medium'
 DISK='10' # in GiB
 SSHKEY='key-name'
 TEMPLATE=$TEMPLATE # template name set above
-exo compute instance create $NAME \
+exo compute instance create "${NAME}" \
     --disk-size $DISK \
-    --ssh-key $SSHKEY \
+    --ssh-key "${SSHKEY}" \
     --template $TEMPLATE \
     --template-visibility private \
     --cloud-init "path/to/ignition-file.ign"
@@ -82,7 +82,7 @@ exo compute instance create $NAME \
 
 NOTE: If just SSH access is desired and no further customization is required, you don't need to pass any Ignition file and can omit the `--cloud-init` argument.
 
-TIP: You can find out the instance's assigned IP by running `exo compute instance show $NAME`
+TIP: You can find out the instance's assigned IP by running `exo compute instance show "${NAME}"`
 
 You now should be able to SSH into the instance using the associated IP address.
 
--- a/modules/ROOT/pages/provisioning-qemu.adoc
+++ b/modules/ROOT/pages/provisioning-qemu.adoc
@@ -25,7 +25,7 @@ Fetch the latest image suitable for your target stream (or https://fedoraproject
 [source, bash]
 ----
 STREAM="stable"
-coreos-installer download -s "${STREAM}" -p qemu -f qcow2.xz --decompress -C ~/.local/share/libvirt/images/
+coreos-installer download -s $STREAM -p qemu -f qcow2.xz --decompress -C ~/.local/share/libvirt/images/
 ----
 
 === Setting up a new VM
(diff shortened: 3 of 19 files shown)


travier:
That looks great! Thanks a lot. I'll do an deeper review soon.

modules/ROOT/pages/provisioning-qemu.adoc
```diff
@@ -47,16 +47,16 @@ IGNITION_DEVICE_ARG="-fw_cfg name=opt/com.coreos/config,file=${IGNITION_CONFIG}"
 IGNITION_DEVICE_ARG="-drive file=${IGNITION_CONFIG},if=none,format=raw,readonly=on,id=ignition -device virtio-blk,serial=ignition,drive=ignition"
 
 qemu-kvm -m 2048 -cpu host -nographic -snapshot \
-  -drive if=virtio,file=${IMAGE} ${IGNITION_DEVICE_ARG} \
+  -drive if=virtio,file="${IMAGE}" "${IGNITION_DEVICE_ARG}" \
```
travier, line 50:
```suggestion
  -drive if=virtio,file="${IMAGE}" ${IGNITION_DEVICE_ARG} \
```

Looks like this one should be split on space

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -39,15 +39,15 @@ For maximum compatibility, it is recommended to use `base64` encoding and to pre
...
+CONFIG_ENCODED=$(base64 -w0 example.ign)
 ----
 
 An alternative to plain `base64` encoding is `gzip+base64` as described in the https://coreos.github.io/ignition/supported-platforms/[Ignition supported platforms]. This is especially useful when submitting the Ignition config via `govc` as an _inline_ argument. In that case the encoded config is limited to slightly under 128 KiB on Linux, 256 KiB on macOS, and 32 KiB on Windows (8 KiB if using `cmd.exe` or PowerShell). If your config is larger than that limit, you may be able to submit it inline after compressing it with `gzip`.
 
 [source, bash]
 ----
 CONFIG_ENCODING='gzip+base64'
-CONFIG_ENCODED=$(cat example.ign | gzip -9 | base64 -w0 -)
+CONFIG_ENCODED=$(gzip -9c example.ign | base64 -w0 -)
```
travier, line 50: Same as above

travier (changes requested):

modules/ROOT/pages/provisioning-exoscale.adoc
```diff
@@ -35,20 +35,20 @@ Next you can https://community.exoscale.com/documentation/compute/custom-templat
 # Set the version and calcuate the checksum
 FCOS_VERSION='...'
 FILE="fedora-coreos-${FCOS_VERSION}-exoscale.x86_64.qcow2"
-CHECKSUM=$(md5sum $FILE | cut -d " " -f 1)
+CHECKSUM="$(md5sum "${FILE}" | cut -d " " -f 1)"
```
dustymabe, line 38: I know `$()` allows you to nest things, but reading this still breaks my brain. I think we should just drop the outside quotes here:

```suggestion
CHECKSUM=$(md5sum "${FILE}" | cut -d " " -f 1)
```

dustymabe (commented):
Hmm. I'm a bit torn here. While I see value in making things more correct and observing good practices I also feel like a lot of these changes make things harder to read. All the extra quoting and brackets disrupt the flow as you are reading the docs and makes the lines go longer which means users will more often need to scroll.

Should we focus our efforts here on cases where there is a higher chance of one of these problems happening?

Okeanos:
> Hmm. I'm a bit torn here. While I see value in making things more correct and observing good practices I also feel like a lot of these changes make things harder to read. All the extra quoting and brackets disrupt the flow as you are reading the docs and makes the lines go longer which means users will more often need to scroll.
> 
> Should we focus our efforts here on cases where there is a higher chance of one of these problems happening?

I absolutely understand and I definitely see the value and validity in your review findings. Given the necessity to extract the shell blocks from the asciidoc surroundings to be able to use shellshock at all, loss of context was inevitable and a certainly lead to being overaggressive in seeing & applying suggestions.

I'll go over your findings, fix them – especially the quotes surrounding `$()` blocks (I really shouldn't have applied these in the first place … d'oh) for example – and then we'll see what else can be made nicer to read.

travier:
Let's focus on the cases where there is a potential for issue (space in file path for example) first and we'll see what to do for the other ones later.

modules/ROOT/pages/provisioning-exoscale.adoc
```diff
@@ -35,20 +35,20 @@ Next you can https://community.exoscale.com/documentation/compute/custom-templat
...
 
 # Upload to object storage
 BUCKET='newbucket'
 exo storage mb "sos://${BUCKET}"
-exo storage upload --acl public-read $FILE "sos://${BUCKET}/image-import/"
+exo storage upload --acl public-read "${FILE}" "sos://${BUCKET}/image-import/"
 
 # Create the template using given URL and CHECKSUM
-URL=$(exo storage show "sos://${BUCKET}/image-import/$FILE" --output-template "{{.URL}}")
+URL=$(exo storage show "sos://${BUCKET}/image-import/"${FILE}"" --output-template "{{.URL}}")
```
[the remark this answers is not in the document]
Okeanos: Should be fixed now.

modules/ROOT/pages/provisioning-qemu.adoc
Okeanos: Should be fixed now.

modules/ROOT/pages/provisioning-qemu.adoc
```diff
@@ -47,16 +47,16 @@ IGNITION_DEVICE_ARG="-fw_cfg name=opt/com.coreos/config,file=${IGNITION_CONFIG}"
...
 ----
 
 .Launching FCOS with QEMU (persistent storage)
 [source, bash]
 ----
-qemu-img create -f qcow2 -F qcow2 -b ${IMAGE} my-fcos-vm.qcow2
+qemu-img create -f qcow2 -F qcow2 -b "${IMAGE}" my-fcos-vm.qcow2
 qemu-kvm -m 2048 -cpu host -nographic \
-  -drive if=virtio,file=my-fcos-vm.qcow2 ${IGNITION_DEVICE_ARG} \
+  -drive if=virtio,file=my-fcos-vm.qcow2 "${IGNITION_DEVICE_ARG}" \
```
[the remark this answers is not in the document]
Okeanos: Should be fixed now.

modules/ROOT/pages/provisioning-exoscale.adoc
Okeanos: Should be fixed now.

Okeanos:
> Let's focus on the cases where there is a potential for issue (space in file path for example) first and we'll see what to do for the other ones later.

I think I removed most if not all quotes and additional curly braces where I could reasonably deduce/assume that the variable's value in question is very much unlikely to contain whitespace characters (or other shell-relevant characters like `#`).

At the same time I took the liberty (_duck_) of making the `coreos-installer` invocations reasonably consistent across the pages to offer a more consistent reading experience. I can revert those though if it makes the change set too noisy or is deemed frivolous.

The `virt-install` and `qemu-kvm` commands are a little tricky – we put (sometimes nested multiple times and then unwrapped at invocation time) variables in them that will absolutely contain path expressions (such as `${PWD}`) that are likely to contain spaces themselves. That's quoting hell in a nutshell. While I am tempted to "fix" this … I am not sure it's worth it given that these commands are likely not going to be executed by novices who'll stumble across this. I am overall just as willing to forgo quotes altogether here to not make it even weirder.

Please don't hesitate to call out anything else you want changed – I essentially took it upon myself to do this so no hard feeling if you want it differently or not at all.

modules/ROOT/pages/provisioning-vmware.adoc
travier: Let's keep your version.

modules/ROOT/pages/provisioning-vmware.adoc
```diff
@@ -39,15 +39,15 @@ For maximum compatibility, it is recommended to use `base64` encoding and to pre
 [source, bash]
 ----
 CONFIG_ENCODING='base64'
-CONFIG_ENCODED=$(cat example.ign | base64 -w0 -)
```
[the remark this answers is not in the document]
travier: OK, let's keep your version

modules/ROOT/pages/provisioning-azure.adoc
```diff
@@ -21,8 +21,8 @@ Once you have picked the relevant stream, download, verify, and decompress the l
 
 [source, bash]
 ----
-stream="stable"
-coreos-installer download --decompress -s "${stream}" -p azure -f vhd.xz
+STREAM="stable"
```
travier, line 24: Mixed feeling here. Shouldn't we instead remove all the brackets & quotes for all of those invocations as they will never need them?

Okeanos: Yeah, that's a good point – I'll get that clean up.

jlebon (approved):
This looks like a good overall improvement.
Thanks for working on it!

I'll give some time for others to have a look before merging.

travier (approved):

travier:
Thanks for this and sorry for the (way too long) delay.

PR #554, opened 2023-05-21
docs (toolbox) : Corrected Toolbx | Toolbox spelling error?

samboosa5k:
I'm not 100% sure if this choice of spelling was unintentional, but I saw it and thought it might need fixing :)


travier:
The `toolbx` misspelling is intentional as it's the new "official" name.

Not sure what to do if this is confusing. Maybe we can write `toolbx (tooblox)`? (not great either).

travier:
Nothing to do here as this is the new name.

PR #558, opened 2023-06-06
Added TPM binding to PCR 7 example

### LINKED ISSUE #431: Add TPM with PCR binding example
From:
- https://discussion.fedoraproject.org/t/coreos-plans-for-measured-boot/40537?u=frigo
- https://github.com/coreos/fedora-coreos-tracker/issues/1255

add an example to the docs that binds the LUKS disk encryption to the TPM's PCR 7 value.

lukewarmtemp:
Fixes #431 

- Added LUKS root filesystem encryption bound to PCR 7 example
- Added point that the value stored in PCR 7 will not change
- Added note that PCR 8 cannot be bound


cgwalters (approved):
Looks sane to me.  I might add that there's more documentation for the `config` fields in `man clevis-encrypt-tpm2`?

Issue #292, opened 2021-05-16
Rewrite FAQ


bgilbert:
The top half of the FAQ is pretty stale and could use an edit pass.

travier:
Clearing assignee as I'm not currently working on this.

PR #559, opened 2023-06-07
Drop mentions of CL/AH except migration guides; clean up FAQ

bgilbert:
Container Linux, Atomic Host, and Project Atomic are long gone, and the FAQ is pretty 2018.  Drop references to the former and do an edit pass over the latter.

Fixes https://github.com/coreos/fedora-coreos-docs/issues/292.


dustymabe (approved):
LGTM - thanks for working on this.

cgwalters (approved):

PR #572, opened 2023-08-01
bare-metal: document that `pxe customize` needs Ignition kargs

bgilbert:
Any customizations that require applying a live Ignition config only work if the Ignition kargs are supplied.  The `pxe customize` docs in `live-reference.adoc` already specify those kargs, but the ones in `bare-metal.adoc` don't.  Document the requirement.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1426.


dustymabe (approved):
LGTM

PR #573, opened 2023-08-03
tutorials: various updates and improvements from recent runs

dustymabe:


dustymabe:
@Okeanos i'm intrested in your thoughts on 42c3954.

cgwalters (approved):
Nice!

modules/ROOT/pages/tutorial-services.adoc
```diff
@@ -24,10 +26,11 @@ We will store this script into `/usr/local/bin/public-ipv4.sh` when we provision
 
 == Writing the systemd service
 
-We need to call the script from the previous section by using a systemd unit. Here is one that works for what we want, which is to execute on first boot and not again:
+We need to call the script from the previous section by using a systemd unit. Let's write a systemd unit into the `issuegen-public-ipv4.service` file that does what we want, which is to execute on first boot and not again:
 
 [source,service]
 ----
+cat <<'EOF' > issuegen-public-ipv4.service
```
Okeanos, line 33: This source block is still declared as type `service` but with the new shell redirects to write the output to a file it should probably be a `bash` block instead?

Okeanos (approved):
LGTM, just a small nit.

modules/ROOT/pages/tutorial-services.adoc
dustymabe: fixed!

dustymabe:
> Nice!

Thanks for the review @cgwalters

PR #575, opened 2023-08-21
[AWS] Add a full example to using a remote config from s3 storage 

### LINKED ISSUE #414: AWS: Include a full example using a remote config with S3 storage
This is a common issue with larger Ignition configs. Let's add a full example to the AWS provisioning page:

- include a full example of https://docs.fedoraproject.org/en-US/fedora-coreos/remote-ign/
- using S3 https://coreos.github.io/butane/config-fcos-v1_4/
- in the page https://docs.fedoraproject.org/en-US/fedora-coreos/provisioning-aws/

jbtrystram:
Add a guide to create an s3 bucket then upload an ignition config to it.
Then create the instance with a minimal config pointing to the s3 bucket 

Fixes #414

### DIFF
--- a/modules/ROOT/pages/provisioning-aws.adoc
+++ b/modules/ROOT/pages/provisioning-aws.adoc
@@ -59,3 +59,45 @@ You now should be able to SSH into the instance using the associated IP address.
 ----
 ssh core@<ip address>
 ----
+
+==== Remote ignition configuration
+
+As user-data is limited to 16 KB, you may need to use an external source for your ignition configuration.
+A common solution is to upload the config to a S3 bucket, as the following steps show:
+
+[source, bash]
+.Create a new s3 bucket
+----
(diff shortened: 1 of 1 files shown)


travier (approved):

travier:
Looks good. I've not tested it.

modules/ROOT/pages/provisioning-aws.adoc
```diff
@@ -59,3 +59,45 @@ You now should be able to SSH into the instance using the associated IP address.
 ----
 ssh core@<ip address>
 ----
+
+==== Remote ignition configuration
```
travier, line 63: Let's also remove one level here to let this section appear in the table of content

jbtrystram: Addressed in #576

PR #577, opened 2023-08-22
live-booting: mention new coreos.liveiso.fromram kernel argument

dustymabe:
This was added in https://github.com/coreos/fedora-coreos-config/pull/2544.

See also https://github.com/coreos/fedora-coreos-tracker/issues/1552

### DIFF
--- a/modules/ROOT/pages/live-booting.adoc
+++ b/modules/ROOT/pages/live-booting.adoc
@@ -25,12 +25,16 @@ podman run --security-opt label=disable --pull=always --rm -v .:/data -w /data \
     quay.io/coreos/coreos-installer:release download -f iso
 ----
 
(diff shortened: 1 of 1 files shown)


Floris497 (approved):
tested and works

travier (approved):

travier:
Wondering if we should make that a separated example to make it clearer that it's not needed / not the default.

dustymabe:
> Wondering if we should make that a separated example to make it clearer that it's not needed / not the default.

I started out this way. In fact I had it all written out and ready to submit. Then I realized that if we have two examples and one of them being optional the `-o customized.iso fedora-coreos-{stable-version}-live.x86_64.iso` doesn't really work so well. You'd need to add a first step that copies `fedora-coreos-{stable-version}-live.x86_64.iso` to `customized.iso` and then operate directly on that for both customize calls. 

Ultimately I ended up here. Definitely open to suggestions.

PR #588, opened 2023-09-26
provisioning-openstack: some minor tweaks

jlebon:
A few user-friendly tweaks for anyone not very familiar with OpenStack.
1. Add `--progress` to `openstack image create` to get a progress bar of the upload.
2. Note that the image is ready to use when it's in the `active` state.
3. Mention `--wait` parameter in instance launching tips.
4. Tweak the floating IP section so it's clearer that you don't have to do this if the instance's network is already externally accessible.


dustymabe (approved):
LGTM

PR #591, opened 2023-09-29
Customize NIC: Fix documentation

ravanelli:
There is a single quote in the butane configuration, which is causing the udev rule to fail.


dustymabe:
The changes look good but maybe we can get a more descriptive commig title and message?

yasminvalim (approved):

PR #593, opened 2023-10-04
provisioning-digitalocean: add jq dep and warn about creation time

jmarrero:
followup to https://github.com/coreos/fedora-coreos-tracker/issues/1586


c4rt0 (approved):
LGTM

PR #590, opened 2023-09-27
provisioning/rpi4: update to f39; fix copy u-boot.bin step

dustymabe:
Update to F38 since stable is on F38 and also update the u-boot.bin copy step for:

- use rpi_arm64/u-boot.bin from uboot-tools as recommended by pwhalen because it supports both rpi3 and rpi4.
- In bcm283x-firmware the default location of the uboot bin file was changed [1].

[1] https://src.fedoraproject.org/rpms/bcm283x-firmware/c/489ba9a1607d29ed171540a675a43e71aa0d1f37?branch=rawhide

PR #599, opened 2023-11-04
provisioning/rpi4: update rpi-u-boot.bin name

jmariondev:
As of FC38, `bcm283x-firmware`'s `config.txt` refers to a single `rpi-u-boot.bin` instead of different files for each RPi variant as in FC37, changed in this commit: https://src.fedoraproject.org/rpms/bcm283x-firmware/c/489ba9a1607d29ed171540a675a43e71aa0d1f37

Copying the u-boot bin file to the correct location makes this document possible to follow for FC38+.


dustymabe:
Thanks for the PR. There is a PR open already that should address this but I was holding off merging it until we figured out https://github.com/coreos/fedora-coreos-docs/pull/590#issuecomment-1783555180 - I think at this point I'll just update the PR to set the `nomodeset` kernel argument upon install and go ahead and get that PR merged.

jmariondev:
Sounds good, I just hit this last night and had to debug it. It makes sense I didn't notice the VGA issue as I was only watching the UART.

PR #605, opened 2023-12-12
Add permissions tip

buckaroogeek:
Add a user permission tip for libvirt provisioning. Unprivileged user qemu needs access to user's home directory and the storage directory for the image. chmod a+x can be used as one option. Thanks to Timothée Ravier for pointing this out.


travier:
This happens when as an unprivileged user, you create a VM using the system libvirtd pointing to a backing store file in you home directory.

You should either:
- Move the backing file to a system location
- Use the session libvirtd instance
- Open up permissions as needed for the qemu user to access you files.

PR #608, opened 2024-01-28
Do not update antora scheduled on forks

rffontenelle:
Avoid running scheduled workflow and failing due to token for PR not being set. One still can start manually this workflow on forks.


travier (approved):

travier:
Thanks

PR #610, opened 2024-01-31
Update provisioning-vmware.adoc

lukonjun:
Fix typo. guestinfo.ignition.config.data should use the env variable CONFIG_FILE_ENCODED like in the prior example


jlebon:
`CONFIG_ENCODED` is defined higher up as the base64 contents of the file. Did you hit any issues?

fifofonix:
It has been a while but I re-executed the steps on this page and didn't have an issue with `CONFIG_ENCODED` if it has been populated per the detailed steps as @jlebon points out.

lukonjun:
Hi sorry for the confusion. I went for the vSphere Option 

```bash
govc vm.change -vm "${VM_NAME}" -e "guestinfo.ignition.config.data.encoding=${CONFIG_ENCODING}"
govc vm.change -vm "${VM_NAME}" -f "guestinfo.ignition.config.data=${CONFIG_FILE_ENCODED}" # using `-f` with a file path instead of `-e`
```

And was therefore confused why it is `CONFIG_ENCODED` instead of `CONFIG_FILE_ENCODED`. I think this can be closed. Great Documentation in general btw, had a very easy time getting to know CoreOS 👍 .

PR #618, opened 2024-03-12
Add an iSCSI install guide

jbtrystram:
Derived from bare-metal


modules/ROOT/pages/bare-metal.adoc
```diff
@@ -174,6 +174,32 @@ A useful approach is to script generating a per-machine `.iso`. If you have a ha
 
 Alternatively, instead of generating per-machine ISOs, you can have a `--pre-install` script run a privileged container which inspects the target system and writes an appropriate https://coreos.github.io/coreos-installer/customizing-install/#config-file-format[installer config] to `/etc/coreos/installer.d`.
 
+=== Installing on iSCSI
+
+To install CoreOS on an iSCSI boot device, follow the same
+steps as described above to get the live environnement, simply adding the correct kernel arguments.
```
travier, line 180:
```suggestion
steps as described above to get the live environnement and add iSCSI related kernel arguments.
```

Suggestion. Things are rarely "simple" :)

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -174,6 +174,32 @@ A useful approach is to script generating a per-machine `.iso`. If you have a ha
...
+
+
+- Mount the iscsi target from the live environnement :
```
jlebon, line 183:
```suggestion
- Mount the iSCSI target from the live environment:
```

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -174,6 +174,32 @@ A useful approach is to script generating a per-machine `.iso`. If you have a ha
...
+[source, bash]
+----
+sudo iscsiadm -m discovery -t st -p 127.0.0.1
+sudo iscsiadm -m node -T iqn.2023-10.coreos.target.vm:coreos -l
+----
+
+- run `coreos-installer`:
```
jlebon, line 190:
```suggestion
- Append the necessary kargs when running `coreos-installer`:
```

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -174,6 +174,32 @@ A useful approach is to script generating a per-machine `.iso`. If you have a ha
...
+[source, bash]
+----
+sudo coreos-installer install \
```
jlebon, line 193: Maybe let's just include both the manual and iBFT examples in full? There's a syntax for labeling each one:

```adoc
.Installing to an iSCSI target with iBFT
[source, bash]
---
...
---

.Installing to an iSCSI target with manual configuration
[source, bash]
---
...
---
```

And then let's also note that all this can also be set at `coreos-installer iso customize` or `coreos-installer pxe customize` time? (We can use `--pre-install` to run the `iscsiadm` command, and then `--dest-device` and `--dest-karg-append` for the rest -- though would be good to sanity-check that this works as expected.)

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -177,28 +177,60 @@ Alternatively, instead of generating per-machine ISOs, you can have a `--pre-ins
 === Installing on iSCSI
 
 To install CoreOS on an iSCSI boot device, follow the same
-steps as described above to get the live environnement, simply adding the correct kernel arguments.
+steps as described above to get the live environnement, and add add the iSCSI related kernel arguments.
```
jlebon, line 180:
```suggestion
steps as described above to get the live environnement, and add the iSCSI-related kernel arguments.
```

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -177,28 +177,60 @@ Alternatively, instead of generating per-machine ISOs, you can have a `--pre-ins
...
 sudo coreos-installer install \
     /dev/disk/by-path/ip-127.0.0.1\:3260-iscsi-iqn.2023-10.coreos.target.vm\:coreos-lun-0 \
     --append-karg rd.iscsi.initiator=iqn.2024-02.com.yourorg.name:lun0 \
     --append-karg netroot=iscsi:iqn.2023-10.coreos.target.vm:coreos` \
     --console ttyS0 \
     --ignition-url https://example.com/example.ign
+---
+
+
+NOTE: All this can also be  set using `coreos-installer iso customize` or `coreos-installer pxe customize`. (See "Customizing installation" section above).
```
jlebon, line 214:
```suggestion
NOTE: All this can also be set using `coreos-installer iso customize` or `coreos-installer pxe customize`. (See "Customizing installation" section above).
```

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -177,28 +177,60 @@ Alternatively, instead of generating per-machine ISOs, you can have a `--pre-ins
...
+
+For example using iBFT:
+
+[source,bash,subs="attributes"]
+----
+# Create customized.iso which:
+# - Automatically mounts iscsi target using mount-iscsi.sh
```
jlebon, line 221:
```suggestion
# - Automatically mounts iSCSI target using mount-iscsi.sh
```

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -177,28 +177,60 @@ Alternatively, instead of generating per-machine ISOs, you can have a `--pre-ins
...
+# - Provisions the installed system with config.ign
+# - Configures the installed GRUB and kernel to use iBFT
```
jlebon, line 223: Technically these don't configure GRUB, only the kernel. :) Maybe just

```suggestion
# - Configures the installed system to use iBFT
```

?

But on that topic, I guess it's implied that the user will need to have e.g. an iPXE script for actually booting up. I wonder if we should at least have a `NOTE` about it if not also a sample iPXE script?

modules/ROOT/pages/bare-metal.adoc
```diff
@@ -177,28 +177,60 @@ Alternatively, instead of generating per-machine ISOs, you can have a `--pre-ins
...
     --append-karg rd.iscsi.initiator=iqn.2024-02.com.yourorg.name:lun0 \
     --append-karg netroot=iscsi:iqn.2023-10.coreos.target.vm:coreos` \
     --console ttyS0 \
     --ignition-url https://example.com/example.ign
+---
+
+
+NOTE: All this can also be  set using `coreos-installer iso customize` or `coreos-installer pxe customize`. (See "Customizing installation" section above).
+
+For example using iBFT:
```
jlebon, line 216: Nice! I like that we provide this example. Were you able to test this?

I guess logically this goes together with what we wrote in `NOTE` but I don't think it'll appear in the same box, does it? In that case, probably cleaner to just not make it a `NOTE` anymore.

modules/ROOT/pages/bare-metal.adoc
jbtrystram: My thinking was " if you are doing iBFT you know what you are doing already " but we could add that, it's not like it's a big addition

modules/ROOT/pages/bare-metal.adoc
jbtrystram: let's try it out during test days ? 
TBH I did not spend the time to fire up the whole iscsi test setup to verify, since the test setup we have in kola works, and this is pretty straight forward.
Let's mark it to give it an actual test run during the FCOS test wek ?

modules/ROOT/pages/bare-metal.adoc
jbtrystram: I added an iPXE boot script example.

jlebon (approved):
Nice!

PR #619, opened 2024-03-14
Remove 'simple', 'good practice' and related terms

travier:
Remove 'simple', 'good practice' and related terms

Things are generally not 'simple' thus the need for documentation. Let's
remove those terms from our docs as much as possible.

---

provisioning-vmware: Turn note into a NOTE

### DIFF
--- a/modules/ROOT/pages/producing-ign.adoc
+++ b/modules/ROOT/pages/producing-ign.adoc
@@ -17,7 +17,7 @@ During the transpilation process, Butane verifies the syntax of the YAML file, w
 
 Once you have an Ignition (`.ign`) file, you can use it to boot an FCOS system in a VM or install it on bare metal.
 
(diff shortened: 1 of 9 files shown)


modules/ROOT/pages/producing-ign.adoc
```diff
@@ -166,7 +166,7 @@ Invoke-RestMethod -Uri https://fedoraproject.org/fedora.gpg | gpg --import
 gpg --verify butane-x86_64-pc-windows-gnu.exe.asc
 ----
 
-== A simple example
+== Full example
```
jlebon, line 169: Maybe just "An example" ?

jlebon (approved):
LGTM overall.

modules/ROOT/pages/producing-ign.adoc
travier: I went with just "Example"

PR #633, opened 2024-04-22
storage: don't use `/dev/vda`

jlebon:
We have `/dev/disk/by-id/coreos-boot-disk` now which is just better. That symlink is already explained higher up on the same page, so no need to explain it again.


dustymabe (approved):
LGTM

PR #636, opened 2024-04-24
Wi-Fi: auto-start connection

hrismarin:
Adding the NetworkManager-wifi package as an extension does not automatically start the Wi-Fi connection. NetworkManager.service should be restarted to establish the connection.

### DIFF
--- a/modules/ROOT/pages/sysconfig-enabling-wifi.adoc
+++ b/modules/ROOT/pages/sysconfig-enabling-wifi.adoc
@@ -58,6 +58,7 @@ systemd:
         RemainAfterExit=yes
         ExecStart=/usr/bin/rpm-ostree install --apply-live --allow-inactive NetworkManager-wifi iwlwifi-dvm-firmware
         ExecStart=/bin/touch /var/lib/%N.stamp
+        ExecStartPost=/usr/bin/systemctl restart NetworkManager.service
         [Install]
         WantedBy=multi-user.target
 storage:


modules/ROOT/pages/sysconfig-enabling-wifi.adoc
```diff
@@ -58,6 +58,7 @@ systemd:
         RemainAfterExit=yes
         ExecStart=/usr/bin/rpm-ostree install --apply-live --allow-inactive NetworkManager-wifi iwlwifi-dvm-firmware
         ExecStart=/bin/touch /var/lib/%N.stamp
+        ExecStartPost=/usr/bin/systemctl restart NetworkManager.service
```
jlebon, line 61: Hmm, feels like for a oneshot service, there isn't much difference between `ExecStart=` and `ExecStartPost=`? Fine as is too though!

jlebon: That said, did you try `systemctl reload NetworkManager.service` instead? If that works, that should be preferred I think.

hrismarin: I tried `systemctl reload NetworkManager.service` but it didn't work.

travier (approved):

dustymabe:
considering https://github.com/coreos/fedora-coreos-tracker/issues/1691#issuecomment-2078790842 which I had forgot about, I guess we should actually change the docs here to not use `--apply-live` and to actually do a reboot after the layering operation is complete.

I think @jlebon is working on a doc update for this in the os extensions section and then we can update the docs on this page to do the same.

at that point we won't need to restart NetworkManager because the system will have rebooted

dustymabe:
https://github.com/coreos/fedora-coreos-docs/pull/637#event-12627358215 should mean that we are rebooting at least for now.. @hrismarin we can leave this PR open for when https://github.com/coreos/fedora-coreos-tracker/issues/1691 is fixed (@jlebon opened https://github.com/coreos/rpm-ostree/pull/4939 for that) if you like.

Issue #651, opened 2024-07-08
Script to "convert" an existing image for a platform where we don't provide ready made images


travier:
I'll make this into a doc page. Can be run from COSA for example.

```
#!/bin/bash

set -euo pipefail
# set -x

if [[ ${#} -ne 3 ]]; then
    echo "Usage: <source image> <dest image> <platform>"
    echo ""
    echo "Example:"
    echo "./$(basename "${0}") fedora-coreos-40.20240616.3.0-{qemu,heztner}.x86_64.qcow2 heztner"
    exit 1
fi

source="${1}"
dest="${2}"
platform="${3}"

if [[ ! -f "${source}" ]]; then
    echo "Source image ${source} does not exists"
    exit 1
fi

if [[ -f "${dest}" ]]; then
    echo "Destination image ${dest} already exists"
    exit 1
fi

cp --reflink=auto "${source}" "${dest}"
guestfish -a "${dest}" <<EOF
run
mount /dev/sda3 /
download /loader/entries/ostree-1.conf tmp.loader.entries.ostree-1.conf
<! sed -i "s/ignition.platform.id=qemu/ignition.platform.id=${platform}/" tmp.loader.entries.ostree-1.conf
upload tmp.loader.entries.ostree-1.conf /loader/entries/ostree-1.conf
EOF

rm -v ./tmp.loader.entries.ostree-1.conf

echo "Done"
```

travier:
Merging that into the docs pages directly for now.

PR #652, opened 2024-07-10
Provisioning: Add Oracle Cloud Infrastructure

travier:
Use: https://github.com/coreos/fedora-coreos-docs/issues/651


travier:
Draft as it's not complete yet. I still need to figure out the networking section.

modules/ROOT/pages/provisioning-oraclecloud.adoc
```diff
@@ -0,0 +1,200 @@
...
+echo "Done"
+----
++
+. Convert the QEMU image to an Oracle Cloud Infrastructure one:
++
+[source, bash]
+----
+source_image"fedora-coreos-qemu-image-name.qcow2"
+image_name="fcos-oraclecloud.qcow2"
+./covert-image.sh "${source_image}" "${image_name} oraclecloud
```
travier, line 77: I used `openstack` as a workaround until we get support in order in Ignition.
This uses the legacy IMDS endpoint (v1) so it must not be disabled when creating the image.

jlebon:
I'm sure you've seen this, though note that bare metal and VM shapes have different requirements: https://github.com/coreos/fedora-coreos-tracker/issues/414#issuecomment-1795808614. Short-term, getting the VM shapes to work shouldn't be too much work. For the bare metal shapes, we could for now recommend using the bare metal installation flows, until we have "dual" support in the same image.

travier:
Ah indeed, thanks for the pointer. Would adding those kargs directly to the image as part of the conversion step make sense to you? That way we could have users upload an image with the kargs pre-set and then select the right shape for it.

jlebon:
> For Bare Metal shapes, maybe we should add the kargs directly to the image as part of the conversion step that we have right now as we don't produce oracle cloud specific images. Then we would have users upload this image and select the right shape for it.

I think that's fine for now, yeah.

travier:
Based on top of https://github.com/coreos/fedora-coreos-docs/pull/654 for convenience as its likely to be merged earlier.

modules/ROOT/pages/provisioning-oraclecloud.adoc
```diff
@@ -0,0 +1,311 @@
...
+Note this is just using `coreos-installer` as a tool to download the QCOW2 disk image.
++
+NOTE: Both x86_64 and aarch64 architectures are supported on Oracle Cloud Infrastructure.
+
+. Copy paste the following Bash script into a file name `convert-image.sh`:
++
+.QEMU to Oracle Cloud Infrastructure image conversion script
+[source, bash]
+----
+#!/bin/bash
```
jlebon, line 47: Definitely would be nice to just stick this in a container image, but there's overhead there too in maintenance.

AsciiDoc supports including files in multiple documents. That would allow us to define this script just once to make it easier to maintain here. (And I guess make the plaform ID be another parameter.)

jlebon (commented):
Thanks for writing this up!

modules/ROOT/pages/provisioning-oraclecloud.adoc
travier: Agree that it would be nice to find a good place for this script.

michael-grundmann:
i tried your instructions, but it seems that the metadata endpoint used in afterburn is not available here:

Jul 19 12:42:30.959 WARN failed to locate config-drive, using the metadata service API instead
Jul 19 12:42:30.966 INFO Fetching http://169.254.169.254/latest/meta-data/public-keys: Attempt #1
Jul 19 12:42:30.972 INFO Fetch failed with 404: resource not found


Instead i found it available here: http://169.254.169.254/openstack/latest/meta_data.json or here http://169.254.169.254/opc/v1/instance/metadata/ssh_authorized_keys

ist there a way to overide the metadata endpoint that is used by afterterburn (i.e by kernel command line)?

travier:
Thanks for giving it a try.

This is due to the fact that we re-use the openstack implementation as is for now which is not the long term plan: https://github.com/coreos/fedora-coreos-tracker/issues/414#issuecomment-596855759

This is why I've commented out all Afterburn mentions in the docs for now.

As far as I know, afterburn does not let you configure that on the kernel command line.

modules/ROOT/pages/provisioning-oraclecloud.adoc
jlebon: At least, I think we should probably lift this for now to a separate page so that all emerging platforms can reference it (but say, "pass this platform ID to the script").

alaviss:
@travier I have recently added native support the OCI platform to ignition and afterburn under the ID `oraclecloud`. Do you mind updating this PR?

travier:
Wow, thanks, this looks great. Feel free to take the changes from this PR and make a new one as I'll be busy until next week at least.

modules/ROOT/pages/provisioning-oraclecloud.adoc
```diff
@@ -0,0 +1,314 @@
...
+This guide shows how to provision new Fedora CoreOS (FCOS) nodes on Oracle Cloud Infrastructure.
+Fedora CoreOS images are currently not published directly on Oracle Cloud Infrastructure.
+Thus you must first download a Fedora CoreOS QEMU (QCOW2) image, then convert it to an Oracle Cloud Infrastructure image and finally upload it to your Oracle Cloud Infrastructure account as a https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/importingcustomimagelinux.htm[custom image].
+
+IMPORTANT: Support for Fedora CoreOS on Oracle Cloud Infrastructure is considered emerging, in that it does not yet offer an optimized user experience.
+           See https://github.com/coreos/fedora-coreos-tracker/issues/414[issue #414] for more details.
+
+IMPORTANT: Support in Fedora CoreOS currently uses the legacy, OpenStack compatible, Instance Metadata Service in OCI to re-use existing OpenStack support in Ignition and Afterburn.
+           For more information about the security implications, see https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/gettingmetadata.htm[Instance Metadata Service v2].
+           This is temporary until support for OCI is added to Ignition and Afterburn.
```
travier, line 12: This will go away once we have images with the code in afterburn and ignition

modules/ROOT/pages/provisioning-oraclecloud.adoc
```diff
@@ -0,0 +1,314 @@
+= Provisioning Fedora CoreOS on Oracle Cloud Infrastructure (OCI)
+
+This guide shows how to provision new Fedora CoreOS (FCOS) nodes on Oracle Cloud Infrastructure.
+Fedora CoreOS images are currently not published directly on Oracle Cloud Infrastructure.
+Thus you must first download a Fedora CoreOS QEMU (QCOW2) image, then convert it to an Oracle Cloud Infrastructure image and finally upload it to your Oracle Cloud Infrastructure account as a https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/importingcustomimagelinux.htm[custom image].
```
travier, line 5: This can be simplified once we start publishing images for OCI.

modules/ROOT/pages/provisioning-oraclecloud.adoc
```diff
@@ -0,0 +1,314 @@
...
+
+== Prerequisites
+
+Before provisioning an FCOS machine, you must have an Ignition configuration file containing your customizations.
+If you do not have one, see xref:producing-ign.adoc[Producing an Ignition File].
+
+NOTE: Fedora CoreOS has a default `core` user that can be used to explore the OS.
+      If you want to use it, finalize its xref:authentication.adoc[configuration] by providing e.g. an SSH key.
+
+// If you do not want to use Ignition to get started, you can make use of the https://coreos.github.io/afterburn/platforms/[Afterburn support].
```
travier, line 22:
```suggestion
If you do not want to use Ignition to get started, you can make use of the https://coreos.github.io/afterburn/platforms/[Afterburn support].
```

modules/ROOT/pages/provisioning-oraclecloud.adoc
```diff
@@ -0,0 +1,314 @@
...
+
+Fedora CoreOS is designed to be updated automatically, with different schedules per stream.
+
+. Once you have picked the relevant stream, download the latest QEMU image from the https://fedoraproject.org/coreos/download/?stream=stable#baremetal[download page] or with podman (see https://coreos.github.io/coreos-installer/cmd/download/[documentation] for options):
++
+[source, bash]
+----
+arch="x86_64" # or aarch64
+podman run --security-opt label=disable --pull=always --rm -v .:/data -w /data \
+    quay.io/coreos/coreos-installer:release download -s stable -p qemu -f qcwo2 -a "${arch}"
```
travier, line 40:
```suggestion
    quay.io/coreos/coreos-installer:release download -s stable -p oraclecloud -f qcwo2 -a "${arch}"
```

modules/ROOT/pages/provisioning-oraclecloud.adoc
```diff
@@ -0,0 +1,314 @@
...
+----
++
+Note this is just using `coreos-installer` as a tool to download the QCOW2 disk image.
++
+NOTE: Both x86_64 and aarch64 architectures are supported on Oracle Cloud Infrastructure.
+
+. Copy paste the following Bash script into a file name `convert-image.sh`:
```
travier, line 47: Should go away once we start publishing images

PR #661, opened 2024-09-01
Minor RPI doc fix

vmpn:
Fix command


travier (approved):

vmpn:
I cannot speak for others but the original command failed for me in bash on debian

hrismarin:
The updated version of the command works on Fedora i.e. locates the EFI System partition correctly.

PR #669, opened 2024-10-09
gcp: add `Launch a confidential VM` part

HuijingHei:


modules/ROOT/pages/provisioning-gcp.adoc
```diff
@@ -85,3 +85,41 @@ gcloud compute instances create                \
...
+. Example verify Confidential VM
+[source, bash]
+----
+ssh core@<ip address>
+# Confirm the VM is using `AMD SEV-SNP` confidential type
+sudo systemd-detect-virt --cvm
+sev-snp
+---
+
+See https://cloud.google.com/confidential-computing/confidential-vm/docs/confidential-vm-overview[confidential type] to have an overview about confidential vm.
```
dustymabe, line 123: I would put this information at the top of the section in a `NOTE` and reword slightly.

```suggestion
NOTE: For an overview about confidential VMs on GCP see https://cloud.google.com/confidential-computing/confidential-vm/docs/confidential-vm-overview[confidential VM overview].
```

modules/ROOT/pages/provisioning-gcp.adoc
```diff
@@ -85,3 +85,41 @@ gcloud compute instances create                \
...
+
+NOTE: Currently, we don't support `--confidential-compute-type=TDX` on GCP.
```
dustymabe, line 125: I would move this note above `Example verify Confidential VM` section.

dustymabe: is there an issue that tracks TDX support? maybe link to it?

modules/ROOT/pages/provisioning-gcp.adoc
```diff
@@ -85,3 +85,41 @@ gcloud compute instances create                \
...
+    --image-project "fedora-coreos-cloud"      \
+    --image-family "fedora-coreos-${STREAM}"   \
+    --metadata-from-file "user-data=${CONFIG}" \
+    --confidential-compute-type "SEV_SNP"      \
+    --machine-type "${MACHINE_TYPE}"           \
+    --maintenance-policy terminate             \
+    --zone "${ZONE} "${NAME}"
+----
+
+. Example verify Confidential VM
```
dustymabe, line 114:
```suggestion
. Example Confidential VM Boot Verification
```

modules/ROOT/pages/provisioning-gcp.adoc
```diff
@@ -85,3 +85,41 @@ gcloud compute instances create                \
 
 NOTE: By design, https://cloud.google.com/compute/docs/startupscript[startup scripts] are not supported on FCOS. Instead, it is recommended to encode any startup logic as systemd service units in the Ignition configuration.
 Again, note you need to use the `user-data` key for Ignition; it will also not work to paste Ignition into this field in the web console.
+
+
+== Launch a confidential VM
+
+To launch a confidential FCOS instance, need to specify the confidential compute type, and related https://cloud.google.com/confidential-computing/confidential-vm/docs/supported-configurations[machine type] that supports the confidential compute type.
```
dustymabe, line 92:
```suggestion
To launch a confidential FCOS instance specify the confidential compute type, and a related https://cloud.google.com/confidential-computing/confidential-vm/docs/supported-configurations[machine type] that supports confidential compute.
```

modules/ROOT/pages/provisioning-gcp.adoc
```diff
@@ -85,3 +85,41 @@ gcloud compute instances create                \
 
 NOTE: By design, https://cloud.google.com/compute/docs/startupscript[startup scripts] are not supported on FCOS. Instead, it is recommended to encode any startup logic as systemd service units in the Ignition configuration.
 Again, note you need to use the `user-data` key for Ignition; it will also not work to paste Ignition into this field in the web console.
+
+
+== Launch a confidential VM
```
dustymabe, line 90:
```suggestion
== Launch a Confidential VM
```

HuijingHei:
Thanks @dustymabe for the kindly review and suggestions!

dustymabe (approved):
LGTM.

Maybe we can get someone to run through the steps before we merge, though.

travier:
For this PR and https://github.com/coreos/fedora-coreos-docs/pull/671, we should add a not that while the resulting VM is launched in Confidential Mode, it does not really is a "Confidential VM". The disk still needs to be encrypted, the security offered by Secure Boot right now is almost non-existent, etc.

At this stage, this is more a developer facing documentation as there would be very little benefits for users.

dustymabe:
> The disk still needs to be encrypted

does that mean you'd need to upload your own disk image (already encrypted)?

are there steps for the more "Confidential" option?

travier:
It varies per cloud and setups. Azure has support to encrypt a disk image for you (just the rootfs), bind it to a TPM and then pass it to a new Confidential VM. I don't think we support that in FCOS right now (at least I've not tested).

In the more general case, you would encrypt the disk on first boot. This requires in turn that you can verify the content that you are pulling from the rootfs so this essentially requires composefs with fs-verity. You also have to verify that you booted the right thing with the expected ignition config, so this calls for Secure Boot signed UKI and remote attestation.

So overall, the complete setup for Confidential Computing is not ready right now thus I would prefer that we don't present it in the docs as a "turn this switch on" section. It's good to have this documented (thanks @HuijingHei) but let's make it clear that it needs more work to actually get a confidential VM.

HuijingHei:
> It's good to have this documented but let's make it clear that it needs more work to actually get a confidential VM.

Maybe can hold this until we support FCOS as a real confidential VM, in case resulting confusion for users.

modules/ROOT/pages/provisioning-gcp.adoc
```diff
@@ -85,3 +85,40 @@ gcloud compute instances create                \
 
 NOTE: By design, https://cloud.google.com/compute/docs/startupscript[startup scripts] are not supported on FCOS. Instead, it is recommended to encode any startup logic as systemd service units in the Ignition configuration.
 Again, note you need to use the `user-data` key for Ignition; it will also not work to paste Ignition into this field in the web console.
+
+
+== Launch a Confidential VM
+
```
travier, line 91: Can you add the same warning that we added for Azure here and we'll merge this one as well? Thanks

travier (approved):
LGTM. Thanks!

PR #673, opened 2024-10-24
alternatives: update doc to use the usual `alternatives` commands

HuijingHei:
See https://github.com/coreos/fedora-coreos-tracker/issues/677#issuecomment-2429223322

### DIFF
--- a/modules/ROOT/pages/alternatives.adoc
+++ b/modules/ROOT/pages/alternatives.adoc
@@ -35,3 +35,45 @@ storage:
       overwrite: true
       hard: false
 ----
+
+== Using alternatives commands
+
+Starting with Fedora CoreOS based on Fedora 41, you can use `alternatives` commands to configure the default command.
+
+. Example Butane config using a systemd unit to configure the default iptables backend
+[source,yaml,subs="attributes"]
+----
+variant: fcos
+version: {butane-latest-stable-spec}
+systemd:
+  units:
+    - name: custom-iptables-default.service
+      enabled: true
+      contents: |
+        [Unit]
+        Description=Set the default backend for iptables 
+        [Service]
+        ExecStart=/usr/sbin/alternatives --set iptables /usr/sbin/iptables-legacy
+        RemainAfterExit=yes
+        [Install]
+        WantedBy=multi-user.target
+----
+
+NOTE: We don't recommend configuring the default iptables backend to `iptables-legacy`. This is just an example.
+
+You can also manually run the `alternatives` commands to configure the default command runtime.
+
+. Example to manually configure the default iptables backend
+[source,bash]
+----
+# Check the link info
+alternatives --display iptables
+iptables --version
+
+# Configure iptables to point to iptables-nft
+sudo alternatives --set iptables /usr/sbin/iptables-nft
+
+# Verify iptables version is iptables-nft
+alternatives --display iptables
+iptables --version
+----


modules/ROOT/pages/alternatives.adoc
```diff
@@ -1,37 +1,21 @@
 = Setting alternatives
 
-Due to an https://github.com/fedora-sysv/chkconfig/issues/9[ongoing issue] in how alternatives configurations are stored on the system, Fedora CoreOS systems can not use the usual `alternatives` commands to configure them.
+Now Fedora CoreOS systems can use the usual `alternatives` commands to configure the default command, as https://github.com/fedora-sysv/chkconfig/issues/9[issue] is fixed.
```
travier, line 3: Let's keep the previous content at the bottom as this still applies until everyone as moved to F41 based FCOS. Something like: "Starting with Fedora CoreOS based on Fedora 41, ...". The bottom section can be named "manual setup" or something like that.

travier, line 3: Let's move the fixed issue reference to the bottom or completely remove it as we don't reference all fixed issues in the doc.

modules/ROOT/pages/alternatives.adoc
```diff
@@ -1,37 +1,21 @@
...
 
-Instead, until this issue is resolved, you can set the symlinks directly in `/etc/alternatives`. For example, to use the legacy-based variants of the `iptables` commands:
+For example, to use the legacy-based variants of the `iptables` commands:
 
-[source,yaml,subs="attributes"]
+[source,bash,subs="attributes"]
```
travier, line 7: Let's have a Butane example first and then a manual example using shell commands.

modules/ROOT/pages/alternatives.adoc
```diff
@@ -1,37 +1,21 @@
...
-      overwrite: true
-      hard: false
-    - path: /etc/alternatives/ip6tables-save
-      target: /usr/sbin/ip6tables-legacy-save
-      overwrite: true
-      hard: false
+# The legacy priority is 10, update nft to lower priority, so legacy wins
+pfx=/usr/sbin/iptables
+pfx6=/usr/sbin/ip6tables
+sudo alternatives --install ${pfx} iptables ${pfx}-nft 5 \
```
travier, line 12: Hum, that looks weird. I'll give it a try.

HuijingHei: Thanks for the kind review!

travier:
This looks good. In coordination with https://github.com/coreos/fedora-coreos-tracker/issues/1818, I think we should refocus this page on the reverse of what we have for instructions right now, which is to say that we should give the instructions to "update" systems to the nft backend instead, as we'll be removing the legacy one.

travier:
Let's see how that works in a systemd unit as well to run this via Ignition.

modules/ROOT/pages/alternatives.adoc
```diff
@@ -35,3 +35,24 @@ storage:
...
       hard: false
 ----
+
+== Manual setup
+
+Starting with Fedora CoreOS based on Fedora 41, you can use `alternatives` commands to configure the default command manually.
+
+For example, to use the legacy-based variants of the `iptables` commands:
+
+.Example using alternatives to configure default command manually
```
travier, line 45:
```suggestion
.Example using alternatives to configure default the iptables backend
```

modules/ROOT/pages/alternatives.adoc
```diff
@@ -35,3 +35,24 @@ storage:
       overwrite: true
       hard: false
 ----
+
+== Manual setup
```
travier, line 39:
```suggestion
== Manual configuration
```

modules/ROOT/pages/alternatives.adoc
```diff
@@ -35,3 +35,43 @@ storage:
       overwrite: true
       hard: false
 ----
+
+== Using alternatives commands
+
+Starting with Fedora CoreOS based on Fedora 41, you can use `alternatives` commands to configure the default command.
+
+. Example systemd unit to configure default the iptables backend
```
travier, line 43:
```suggestion
. Example Butane config using a systemd unit to configure the default iptables backend
```

modules/ROOT/pages/alternatives.adoc
```diff
@@ -35,3 +35,43 @@ storage:
...
+        [Service]
+        ExecStart=/usr/sbin/alternatives --set iptables /usr/sbin/iptables-nft
+        RemainAfterExit=yes
+        [Install]
+        WantedBy=multi-user.target
+----
+
+You can also manually running `alternatives` commands to configure the default command runtime.
+
+.Example manually to configure default the iptables backend
```
travier, line 64:
```suggestion
.Example to manually configure the default iptables backend
```

dustymabe:
your examples here set `iptables` to use `iptables-nft` but that happens to already be the default, correct? on my freshly installed `41.20241027.2.0` node I see:

```
core@localhost:~$ iptables --version
iptables v1.8.10 (nf_tables)
core@localhost:~$ alternatives --display iptables
iptables - status is manual.
 link currently points to /usr/sbin/iptables-nft
/usr/sbin/iptables-legacy - priority 10
 follower ip6tables: /usr/sbin/ip6tables-legacy
 follower ip6tables-restore: /usr/sbin/ip6tables-legacy-restore
 follower ip6tables-save: /usr/sbin/ip6tables-legacy-save
 follower iptables-restore: /usr/sbin/iptables-legacy-restore
 follower iptables-save: /usr/sbin/iptables-legacy-save
/usr/sbin/iptables-nft - priority 10
 follower ip6tables: /usr/sbin/ip6tables-nft
 follower ip6tables-restore: /usr/sbin/ip6tables-nft-restore
 follower ip6tables-save: /usr/sbin/ip6tables-nft-save
 follower iptables-restore: /usr/sbin/iptables-nft-restore
 follower iptables-save: /usr/sbin/iptables-nft-save
Current `best' version is /usr/sbin/iptables-legacy.
```

also.. what does this output mean? I see `link currently points to /usr/sbin/iptables-nft`, but also `Current 'best' version is /usr/sbin/iptables-legacy.`, which is a bit confusing.

HuijingHei:
> your examples here set `iptables` to use `iptables-nft` but that happens to already be the default, correct?

> I see `link currently points to /usr/sbin/iptables-nft`, but also `Current 'best' version is /usr/sbin/iptables-legacy.`, which is a bit confusing.

I think it is because we add [postprocess](https://github.com/coreos/fedora-coreos-config/commit/513d60ed032b433cf645401f273869a638403073#diff-2c993c33dfed3829e59bfd9c90a811ef882875d7296adc5d28449b5d9db0f4d6R119) to make it manually default to `iptables-nft`. If we remove the postprocess part, it will be default to `legacy` in auto mode (in my testing).

```
[core@cosa-devsh ~]$ iptables --version
iptables v1.8.10 (legacy)

[core@cosa-devsh ~]$ alternatives --display iptables
iptables - status is auto.
 link currently points to /usr/sbin/iptables-legacy
...
```

Both `iptables-legacy` and `iptables-nft` have the same priority 10 (see https://github.com/coreos/fedora-coreos-tracker/issues/1818), guess it chooses the current best alternatives by alphabetical order.

dustymabe:
> I think it is because we add [postprocess](https://github.com/coreos/fedora-coreos-config/commit/513d60ed032b433cf645401f273869a638403073#diff-2c993c33dfed3829e59bfd9c90a811ef882875d7296adc5d28449b5d9db0f4d6R119) to make it manually default to `iptables-nft`. If we remove the postprocess part, it will be default to `legacy` in auto mode (in my testing).

right. I'm just pointing out that our example should probably be changing it from what we already ship as default? Maybe there is another better example?

travier:
The idea was to give as example the migration that we would want to have users do in https://github.com/coreos/fedora-coreos-tracker/issues/1818 (moving legacy systems to the nft backend).

But I agree that it does not work for the Butane config as we already do that. I don't think we should show the reverse however as we don't want users to go to the legacy backend as we ideally should remove it.

Not sure if we actually have another "alternatives" command to use for the example.

dustymabe:
agree. Thanks for the clarification

HuijingHei:
> Not sure if we actually have another "alternatives" command to use for the example.

There is one `cifs-idmap-plugin` in `alternatives --list`, mabye we can use it? See [doc](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html/windows_integration_guide/smb-sssd#smb-sssd-switch) for reference.
```
[core@cosa-devsh ~]$ sudo alternatives --display cifs-idmap-plugin
cifs-idmap-plugin - status is auto.
 link currently points to /usr/lib64/cifs-utils/cifs_idmap_sss.so
/usr/lib64/cifs-utils/cifs_idmap_sss.so - priority 20
/usr/lib64/cifs-utils/idmapwb.so - priority 10
Current `best' version is /usr/lib64/cifs-utils/cifs_idmap_sss.so.
```

dustymabe:
> Agree this is just an example to run alternatives using butane or manually command, then is this OK to merge?

I would update the PR to set iptables legacy in the butane example. Other than that - LGTM.

dustymabe:
and, of course @travier may have a different opinion.

HuijingHei:
How about adding a note to remind that if want to set iptables to legacy, just replace the target to `/usr/sbin/iptables-legacy`, instead of update the example?

dustymabe:
I guess. It's just confusing to have an example that sets a value to what is already the default. i.e. if I was testing this documentation I'd have no way to verify it worked 

- boot without butane from this example
    - value == iptables-nft
- boot with butane from this example
    - value == iptables-nft

I'd think it was a problem with the docs personally.

HuijingHei:
Update doc: the butane example to configure the default iptables to `legacy`, and change to `nft` using `alternative` commands, not sure my understanding is correct, thanks @dustymabe @travier for the kind review.

modules/ROOT/pages/alternatives.adoc
```diff
@@ -35,3 +35,45 @@ storage:
...
+        [Unit]
+        Description=Set the default backend for iptables 
+        [Service]
+        ExecStart=/usr/sbin/alternatives --set iptables /usr/sbin/iptables-legacy
+        RemainAfterExit=yes
+        [Install]
+        WantedBy=multi-user.target
+----
+
+NOTE: We don't recommend configuring the default iptables backend to `iptables-legacy`, it's just an example.
```
dustymabe, line 62:
```suggestion
NOTE: We don't recommend configuring the default iptables backend to `iptables-legacy`. This is just an example.
```

modules/ROOT/pages/alternatives.adoc
```diff
@@ -35,3 +35,45 @@ storage:
...
+
+You can also manually running `alternatives` commands to configure the default command runtime.
```
dustymabe, line 64:
```suggestion
You can also manually run the `alternatives` commands to configure the default command runtime.
```

modules/ROOT/pages/alternatives.adoc
```diff
@@ -35,3 +35,45 @@ storage:
...
+
+. Example to manually configure the default iptables backend
+[source,bash]
+----
+# Check the link info
+alternatives --display iptables
+iptables --version
+
+# Config iptables to point to iptables-nft
```
dustymabe, line 73:
```suggestion
# Configure iptables to point to iptables-nft
```

dustymabe:
looks pretty good to me - a few final comments.

dustymabe (approved):

PR #688, opened 2024-12-27
Add docker-cli to the list of packages to be removed

psoberoi:
docker-cli conflicts with docker-ce-cli, which is a dependency of docker-ce-stable.


travier:
Thanks. We have https://github.com/coreos/fedora-coreos-docs/pull/689 which is more complete but we can merge this one first and rebase the second.

travier (approved):

PR #690, opened 2024-12-30
antora.yml: Update latest Butane spec attribute

hrismarin:
Update latest Butane spec attribute to version 1.6.0


travier (approved):
Thanks!

PR #702, opened 2025-02-05
provisioning azure: add remote ign ex on private azure blob

prestist:
Do not merge until after ignition has its next release, and it lands in fcos next.

Its now landed I can see it when doing cosa fetch.


yasminvalim (approved):
/lgtm

PR #703, opened 2025-02-07
update-versions.py: Auto update latest Butane spec

hrismarin:
Automatically update the latest butane spec attribute by synchronizing it with Butane specs docs.

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -12,3 +12,4 @@ asciidoc:
     ignition-version: 2.22.0
     butane-version: 0.24.0
     butane-latest-stable-spec: 1.6.0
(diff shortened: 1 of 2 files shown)


hrismarin:
Add an attribute for the latest version of the Ignition specs and synchronize it with the corresponding docs.

travier:
Sounds like a good start. Fetching the raw page in Markdown and filtering it a bit should be a bit safer even if still hackish.

To make it better we would have to somehow expose the latest versions in the repo as a file or something.

prestist:
Thank you for working on this; from my perspective it LGTM. 

@travier I agree; we could have like a stable/latest file in both ignition/butane and just fetch it?

PR #710, opened 2025-02-27
faq: add entry about Fedora Bootc and FCOS

jlebon:
As Fedora Bootc gains more visibility, there are recurring questions about its relationship to FCOS. Let's add an entry in the FAQ for this.

### DIFF
--- a/modules/ROOT/pages/faq.adoc
+++ b/modules/ROOT/pages/faq.adoc
@@ -23,6 +23,28 @@ variety of containerized use cases, RHEL CoreOS provides a
 focused OS for OpenShift, released and life-cycled in tandem
 with the platform.
 
+== How does Fedora CoreOS relate to Fedora Bootc?
+
+https://docs.fedoraproject.org/en-US/bootc/[Fedora Bootc] is
+a Fedora project aimed at building Fedora and CentOS-based
+https://containers.github.io/bootable/[bootable containers]. The goal is to
+eventually have Fedora CoreOS build on top of Fedora Bootc both literally and in
+the broader sense of being part of the same ecosystem. The roadmap for this can
+be found https://github.com/coreos/fedora-coreos-tracker/issues/1726[on GitHub].
+
+Fedora CoreOS adds
+https://docs.fedoraproject.org/en-US/fedora-coreos/update-streams/[a stream model],
+platform-specific integration and disk images, provisioning via Ignition, and
+more tooling around automatic updates. The goal is also to have Fedora CoreOS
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/faq.adoc
```diff
@@ -23,6 +23,28 @@ variety of containerized use cases, RHEL CoreOS provides a
...
+https://containers.github.io/bootable/[bootable containers], The goal is to
+eventually have Fedora CoreOS build on top of Fedora Bootc both literally and in
+the broader sense of being part of the same ecosystem. The roadmap for this can
+be found https://github.com/coreos/fedora-coreos-tracker/issues/1726[on GitHub].
+
+Fedora CoreOS adds
+https://docs.fedoraproject.org/en-US/fedora-coreos/update-streams/[a stream model],
+platform-specific integration and disk images, provisioning via Ignition, and
+more tooling around automatic updates. The goal is also to have Fedora CoreOS
+integrate more strongly with the Kubernetes ecosystem.
```
dustymabe, line 39: I would just drop this last sentence. Not sure it's needed here.

```suggestion
more tooling around automatic updates.
```

modules/ROOT/pages/faq.adoc
```diff
@@ -23,6 +23,28 @@ variety of containerized use cases, RHEL CoreOS provides a
 focused OS for OpenShift, released and life-cycled in tandem
 with the platform.
 
+== How does Fedora CoreOS relate to Fedora Bootc?
+
+https://docs.fedoraproject.org/en-US/bootc/[Fedora Bootc] is
+a Fedora project aimed at building Fedora and CentOS-based
+https://containers.github.io/bootable/[bootable containers], The goal is to
```
dustymabe, line 30:
```suggestion
https://containers.github.io/bootable/[bootable containers]. The goal is to
```

dustymabe (approved):
Generally LGTM. A few suggestions.

modules/ROOT/pages/faq.adoc
cgwalters: I'm curious how FCOS would have a big differentiator for kube vs fedora-bootc.

travier: I'm working on cri-o/kubernetes sysexts and we could potentially build and provide those as official sysexts for FCOS for example, tied to the release cadence.

dustymabe: If we can pair that with some docs I think that would be really nice.

PR #711, opened 2025-02-27
Add applehv provisioning documentation

baude:
Now that vfkit makes it easy to inject ignition files while booting FCOS, we can add some simple documentation for provisioning.  This one is a hybrid between the hyperv and qemu provisioning documents.


baude:
@cfergeau PTAL

modules/ROOT/pages/provisioning-applehv.adoc
```diff
@@ -0,0 +1,56 @@
...
+.Launching FCOS with Vfkit
+[source, bash]
+----
+IGNITION_CONFIG="/path/to/example.ign"
+IMAGE="/path/to/image.qcow2"
+
+./vfkit --cpus 2 --memory 2048 \
+  --bootloader efi,variable-store=efi-variable-store,create \
+  --device virtio-blk,path=${IMAGE} \
+  --device virtio-net,nat \
```
cfergeau, line 38: maybe you can hardcode a mac address here so that the VM IP is always the same? ok with me if you consider this outside of the scope of this document.

modules/ROOT/pages/provisioning-applehv.adoc
```diff
@@ -0,0 +1,56 @@
+= Provisioning Fedora CoreOS on AppleHV
+
+This guide shows how to provision new Fedora CoreOS (FCOS) instances on MacOS using https://github.com/crc-org/vfkit[vfkit].
```
cfergeau, line 3: nit: Apple spells it « macOS »

modules/ROOT/pages/provisioning-applehv.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+You will also need to build or acquire https://github.com/crc-org/vfkit[vfkit].  Prebuilt binaries are also available from its https://github.com/crc-org/vfkit/releases/tag/v0.6.0[releases] page.  Vfkit, like QEMU, has many options which are outside the scope of this provisioning example.  Please consider reading their https://github.com/crc-org/vfkit/tree/main/doc[documentation].
+
+== Booting a new VM on MacOS
+
+This section shows how to boot a new VM with vfkit. Vfkit is known to work with both Intel and Apple Silicon based Macs.
+
+=== Fetching the AppleHV image
+
+Fetch the latest image suitable for your target stream (or https://fedoraproject.org/coreos/download/[download and verify] it from the web).  Remember to download the approach image based on the architecture of your Mac.  Once downloaded, you will also need to decompress the image.
```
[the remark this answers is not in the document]
cfergeau: I would also explicitly say that the appropriate image is the « Apple Hypervisor » one:

> Remember to download the appropriate "Apple Hypervisor" image based on the architecture of your Mac

?

cfergeau (commented):
a few minor comments, but looks good to me!

modules/ROOT/pages/provisioning-applehv.adoc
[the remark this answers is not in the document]
baude: this is all copied from other provisioning examples, so i didnt touch anything

modules/ROOT/pages/provisioning-applehv.adoc
```diff
@@ -0,0 +1,56 @@
...
+Note: The AppleHV hypervisor does not allow you to see early boot and kernel messages.  While you will see a grub boot menu, you will not see anything until later in the boot.
+
+=== Exploring the OS
+
+image::vfkit.png[Vfkit GUI]
+
+When FCOS is completed booting, you will see the IP address of the VM displayed in the GUI window.  Vfkit will lease an address in the 192.168.64/0/24 network.  At this point, you can either choose to login or SSH to the VM.  Unlike some other virtualization providers, you can SSH to the virtual machine from the host.
+[source, bash]
+----
+ssh core@192.168.64.5
```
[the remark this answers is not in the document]
baude: no it doesnt assuming you use a key in a place you know the system will look. in my case, i used an existing key in ~/.ssh

modules/ROOT/pages/provisioning-applehv.adoc
baude: i was trying to make it dead simple.  I dont object.  See what others say

modules/ROOT/pages/provisioning-applehv.adoc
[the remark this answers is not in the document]
dustymabe: hmm. I definitely agree that `appropriate` is the right word to use here. If other examples use the work `approach` we should fix them.

I'm happy with the suggestion by @cfergeau in https://github.com/coreos/fedora-coreos-docs/pull/711/files#r1975056354 if you are @baude

modules/ROOT/pages/provisioning-applehv.adoc
dustymabe: Does a specific MAC make the configured IP deterministic?

I lean more towards not harcoding a MAC address here..

modules/ROOT/pages/provisioning-applehv.adoc
```diff
@@ -0,0 +1,56 @@
...
+[source, bash]
+----
+IGNITION_CONFIG="/path/to/example.ign"
+IMAGE="/path/to/image.qcow2"
+
+./vfkit --cpus 2 --memory 2048 \
+  --bootloader efi,variable-store=efi-variable-store,create \
+  --device virtio-blk,path=${IMAGE} \
+  --device virtio-net,nat \
+  --ignition ${IGNITION_CONFIG} \
```
dustymabe, line 39: interesting that vfkit supports an `--ignition` arg. neat.

modules/ROOT/pages/provisioning-applehv.adoc
```diff
@@ -0,0 +1,56 @@
...
+  --gui
+----
+
+Note: The AppleHV hypervisor does not allow you to see early boot and kernel messages.  While you will see a grub boot menu, you will not see anything until later in the boot.
+
+=== Exploring the OS
+
+image::vfkit.png[Vfkit GUI]
+
+When FCOS is completed booting, you will see the IP address of the VM displayed in the GUI window.  Vfkit will lease an address in the 192.168.64/0/24 network.  At this point, you can either choose to login or SSH to the VM.  Unlike some other virtualization providers, you can SSH to the virtual machine from the host.
```
dustymabe, line 52:
```suggestion
When FCOS is completed booting, you will see the IP address of the VM displayed in the GUI window.  Vfkit will lease an address in the `192.168.64.0/24` network.  At this point, you can either choose to login or SSH to the VM.  Unlike some other virtualization providers, you can SSH to the virtual machine from the host.
```

modules/ROOT/pages/provisioning-applehv.adoc
```diff
@@ -0,0 +1,56 @@
...
+  --device virtio-blk,path=${IMAGE} \
+  --device virtio-net,nat \
+  --ignition ${IGNITION_CONFIG} \
+  --device virtio-input,keyboard \
+  --device virtio-input,pointing \
+  --device virtio-gpu,width=800,height=600 \
+  --gui
+----
+
+Note: The AppleHV hypervisor does not allow you to see early boot and kernel messages.  While you will see a grub boot menu, you will not see anything until later in the boot.
```
dustymabe, line 46: Suggestion -> `GRUB`

It's odd to me that a user can see GRUB but no kernel messages. That would indicate to me that it's not technically impossible, just that we haven't found the right settings?

modules/ROOT/pages/provisioning-applehv.adoc
baude: not deterministic imho

modules/ROOT/pages/provisioning-applehv.adoc
baude: My understanding is that it is not exposed by Apple.  Is this correct @cfergeau ?

dustymabe: but you would think if GRUB can output to a console (either VGA or serial) the kernel could too.. but I certainly don't know this platform at all and will trust the experts

modules/ROOT/pages/provisioning-applehv.adoc
cfergeau: It's more deterministic in the sense that you will always get the same mac address every time you run the VM. But I agree it makes the example more complicated, and it's not really important.

modules/ROOT/pages/provisioning-applehv.adoc
cfergeau: I've filed https://github.com/crc-org/vfkit/issues/270 to document this properly on vfkit's end, as the behaviour varies a lot depending on how you start the VM.

travier (approved):

modules/ROOT/pages/provisioning-applehv.adoc
dustymabe: Thanks @cfergeau. I opened a PR to try to make the change in the images we ship: https://github.com/coreos/fedora-coreos-config/pull/3552

PR #715, opened 2025-03-24
Toolbx: Remove references to a version

hrismarin:
Remove references to a toolbox version (i.e. 35)
so it doesn't need to be updated for each release.

Improve AsciiDoc syntax.

This change was discussed during the
preparatory video session for the F42 test day.

### DIFF
--- a/modules/ROOT/pages/debugging-with-toolbox.adoc
+++ b/modules/ROOT/pages/debugging-with-toolbox.adoc
@@ -1,52 +1,42 @@
 = Debugging with Toolbx
 
-The FCOS image is kept minimal by design to reduce the image size and the
-attack surface. This means that it does not include every troubleshooting tools
-that a normal OS may include. Instead, the recommended approach is to leverage
-containers with the https://containertoolbx.org/[toolbox] utility
-included in the image.
+The FCOS image is kept minimal by design to reduce the image size and the attack surface.
+This means that it does not include every troubleshooting tools that a normal OS may include.
+Instead, the recommended approach is to leverage containers with the https://containertoolbx.org/[toolbox] utility included in the image.
 
 == What is Toolbx?
 
-Toolbx is a utility that allows you to create privileged containers meant to
-debug and troubleshoot your instance. It is a wrapper around podman which
-starts long running containers with default mounts and namespaces to facilitate
-debugging the host system.
+Toolbx is a utility that allows you to create privileged containers meant to debug and troubleshoot your instance.
+It is a wrapper around podman which starts long running containers with default mounts and namespaces to facilitate debugging the host system.
 
-These containers can then be used to install tools that you may need for
-troubleshooting.
+These containers can then be used to install tools that you may need for troubleshooting.
 
 == Using Toolbx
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/debugging-with-toolbox.adoc
```diff
@@ -1,52 +1,42 @@
 = Debugging with Toolbx
 
-The FCOS image is kept minimal by design to reduce the image size and the
-attack surface. This means that it does not include every troubleshooting tools
-that a normal OS may include. Instead, the recommended approach is to leverage
-containers with the https://containertoolbx.org/[toolbox] utility
-included in the image.
+The FCOS image is kept minimal by design to reduce the image size and the attack surface.
+This means that it does not include every troubleshooting tools that a normal OS may include.
```
dustymabe, line 4:
```suggestion
This means that it does not include every troubleshooting tool that a normal OS may include.
```

modules/ROOT/pages/debugging-with-toolbox.adoc
```diff
@@ -1,52 +1,42 @@
...
-These containers can then be used to install tools that you may need for
-troubleshooting.
+These containers can then be used to install tools that you may need for troubleshooting.
 
 == Using Toolbx
 
-You can create a new toolbox by running the command below. On the first run,
-this will ask you if you want to use the image
-`registry.fedoraproject.org/fedora-toolbox:35`.
+You can create a new toolbox by running the command below.
```
dustymabe, line 16: I would keep some of what was there:

```suggestion
You can create a new toolbox by running the command below. On the first run it will ask you if you want to download an image. Answer yes with `y`. 
```

dustymabe (approved):
Mostly LGTM - some suggestions:

dustymabe:
Thanks @hrismarin

dustymabe (approved):
LGTM

dustymabe:
Will merge in a few hours to allow for others to review.

travier (approved):
The default is nice as it gives you a toolbox named for your current release. But setting an explicit name works a well.

We could also add an attribute for the release version to https://github.com/coreos/fedora-coreos-docs/blob/main/antora.yml and update it as needed for all the references in the doc (if there are others).

But that's not blocking this. 👍🏻

hrismarin:
> The default is nice as it gives you a toolbox named for your current release. But setting an explicit name works a well.
> 
> We could also add an attribute for the release version to https://github.com/coreos/fedora-coreos-docs/blob/main/antora.yml and update it as needed for all the references in the doc (if there are others).

I was thinking of adding an attribute for the release version to https://github.com/coreos/fedora-coreos-docs/blob/main/antora.yml and would do it in another PR if you don't mind. I can then see if there are any other references to the toolbox release tag in the docs and try to find a way to automate it to match the current version.

dustymabe:
It would be nice to have an attribute and have it automatically updated, but absent of that I think this update in this PR is reasonable.

@hrismarin let us know which way you'd like to go.

hrismarin:
I think we should merge this PR to avoid confusion in toolbox versions. Later we can make changes with the attribute and automation step by step.

PR #720, opened 2025-04-11
Fix setting up ssh and containers tutorial

tlbueno:
- Change etcd tag from latest to v3.5.21 as latest tag do not exist in https://quay.io/repository/coreos/etcd?tab=tags&tag=latest
- Fix the etcd curl commands as etcd version 3 has a different endpoint to create and query keys.

### DIFF
--- a/modules/ROOT/pages/tutorial-containers.adoc
+++ b/modules/ROOT/pages/tutorial-containers.adoc
@@ -72,7 +72,7 @@ storage:
 
           [Container]
           ContainerName=etcd
-          Image=quay.io/coreos/etcd:latest
+          Image=quay.io/coreos/etcd:v3.5.21
           Network=host
           Volume=etcd-data:/etcd-data
           Exec=/usr/local/bin/etcd                                \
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/tutorial-containers.adoc
```diff
@@ -193,22 +193,26 @@ CONTAINER ID  IMAGE                       COMMAND               CREATED        S
 And we can set a key/value pair in etcd. For now let’s set the key `fedora` to the value `fun`:
 
 ----
-[core@tutorial ~]$ curl -L -X PUT http://127.0.0.1:2379/v2/keys/fedora -d value="fun"
-{"action":"set","node":{"key":"/fedora","value":"fun","modifiedIndex":4,"createdIndex":4}}
-[core@tutorial ~]$ curl -L http://127.0.0.1:2379/v2/keys/ 2>/dev/null | jq .
+[core@tutorial ~]$ curl -L http://127.0.0.1:2379/v3/kv/put -X POST -d '{"key": "ZmVkb3Jh", "value": "ZnVu"}'
```
dustymabe, line 196: Can we still have the key = `fedora` and the value = `fun` ?

dustymabe (approved):
LGTM

modules/ROOT/pages/tutorial-containers.adoc
tlbueno: I don´t think so. As far as I understook in the documentation it needs to be a base64 string.

dustymabe: ahh. I see: 

```
[dustymabe@media ~]$ echo -n 'fedora' | base64 
ZmVkb3Jh
```

Well that's no fun :)

PR #741, opened 2025-07-24
Add provisioning guides for Oracle Cloud Infrastructure

alaviss:
Based on @travier work at https://github.com/coreos/fedora-coreos-docs/pull/652

Changes made includes:

- Removed details about building custom images as they are now built by the project.
- Updated writing style and scripts to align with Azure and Hetzner guides.
- Fixed all bash syntax errors.
- Added image capability configuration as required for confidential computing and AArch64 shapes.
- Updated list of shapes to current offerings.
- Subnets are no longer locked to one availability domain (AD).
- Use AD names instead of ID when launching instances to avoid HTTP 400 errors.
- Added a launch example for Ampere-based instances.

I've validated this using the latest `testing` image.

Closes https://github.com/coreos/fedora-coreos-docs/pull/652

### DIFF
--- a/modules/ROOT/pages/provisioning-oraclecloud.adoc
+++ b/modules/ROOT/pages/provisioning-oraclecloud.adoc
@@ -2,14 +2,10 @@
 
 This guide shows how to provision new Fedora CoreOS (FCOS) nodes on Oracle Cloud Infrastructure.
 Fedora CoreOS images are currently not published directly on Oracle Cloud Infrastructure.
-Thus you must first download a Fedora CoreOS QEMU (QCOW2) image, then convert it to an Oracle Cloud Infrastructure image and finally upload it to your Oracle Cloud Infrastructure account as a https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/importingcustomimagelinux.htm[custom image].
+Thus you must first download the Fedora CoreOS image for Oracle Cloud Infrastructure, then upload it to your account as a https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/importingcustomimagelinux.htm[custom image].
 
 IMPORTANT: Support for Fedora CoreOS on Oracle Cloud Infrastructure is considered emerging, in that it does not yet offer an optimized user experience.
            See https://github.com/coreos/fedora-coreos-tracker/issues/414[issue #414] for more details.
-
-IMPORTANT: Support in Fedora CoreOS currently uses the legacy, OpenStack compatible, Instance Metadata Service in OCI to re-use existing OpenStack support in Ignition and Afterburn.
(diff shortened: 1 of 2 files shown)


travier (approved):
Thanks! LGTM. I have not tested those instructions again but this should be fine. Let's wait until everything is good in stable and merge those?

modules/ROOT/pages/provisioning-oraclecloud.adoc
```diff
@@ -0,0 +1,326 @@
...
+    "VM.DenseIO.E4.Flex"
+    "VM.DenseIO.E5.Flex"
+    "VM.GPU2.1"
+    "VM.GPU3.1"
+    "VM.GPU3.2"
+    "VM.GPU3.4"
+    "VM.GPU.A10.1"
+    "VM.GPU.A10.2"
+    "VM.Optimized3.Flex"
+)
```
dustymabe, line 176: These lists seem prone to get out of date. Is there any way to dynamically determine what values to use here?

I'm sure you've investigated, but there isn't some way to mark it as compatible with all shapes without having to provide all the strings for each one?

modules/ROOT/pages/provisioning-oraclecloud.adoc
```diff
@@ -0,0 +1,326 @@
...
+    --ssh-authorized-keys-file "${SSHKEYS}"
+----
++
+.Example launching customized Ampere instance
+[source, bash]
+----
+NAME=fedora-coreos
+SHAPE=VM.Standard.A1.Flex
+DISK=50                            # size of boot volume in GBs
+OCPUS=1                            # number of allocated OCPUs
```
dustymabe, line 290:
```suggestion
OCPUS=2                            # number of allocated OCPUs
```

Maybe we should use at least 2 here. I've seen some weird behavior before when only 1 CPU is available in VMs because some processes assume multithreading and a single CPU can get deadlocked.

dustymabe (commented):
Thank you for working on this.

It would be great if we could get someone to run through and execute the steps, but absent of that I'm happy to merge.

aednichols:
@dustymabe I did not use the `oci` commands, but following the equivalent steps in the GUI resulted in a working instance.

dustymabe:
changes here LGTM - maybe squash them down into a single commit with a `Co-Authored By` for @travier ?

alaviss:
I think you should be able to do that when you merge.

travier (approved):
Thanks!

travier:
> I think you should be able to do that when you merge.

I did that via the GitHub UI but we generally like to have things in a ready shape so that we don't make mistakes and actually review what is being committed.

Anyway, both works. Thanks!

PR #749, opened 2025-09-01
provisioning (applehv): Add vfkit Homebrew/MacPorts

osalbahr:
- See https://repology.org/project/vfkit/versions


travier (approved):
Thanks

PR #750, opened 2025-09-01
producing-ign: Add password instructions link

osalbahr:


travier (commented):
Let's not that it will not be accepted over SSH by deafult?

osalbahr:
I'm not sure what you mean. Could you clarify?

jlebon:
Password authentication is disabled by default for SSH: https://github.com/coreos/fedora-coreos-config/blob/testing-devel/overlay.d/15fcos/etc/ssh/sshd_config.d/40-disable-passwords.conf

Maybe just something like:

> You may also set a login password for the console (password authentication over SSH is disabled).

jlebon (commented):
Thanks!

PR #755, opened 2025-09-22
provisioning/ibmcloud: update instructions for floating ips

dustymabe:
Not sure what changed but when starting an instance I can't use the floating-ip-update command any longer. Apparently the default now is to attach a "Network Attachment" to an instance rather than a "Network Interface" directly. Anyway I found some other commands that work on the "Virtual NIC" that is part of the attachment so I'll just document that command now.


yasminvalim:
Thanks for working on this!

PR #763, opened 2025-09-28
hostname: Document 'overwrite: true' with Afterburn's hostname

travier:
Some platforms have hostname support in Afterburn. In this case, the file must be overwriten.

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1964


dustymabe (approved):
LGTM

PR #772, opened 2025-11-09
Fix kernel filename in live-booting.adoc

decko:
The kernel file is missing a dot and could cause confusion for first-time or distracted users.


dustymabe (approved):
LGTM.

Thanks. This change that introduced the need for this was https://github.com/coreos/fedora-coreos-tracker/issues/1725

PR #779, opened 2025-12-15
bootloader-updates: automatic updates for the bootloader are enabled by default

rich-purnell:
Starting from F43, the `bootloader-update.service` is enabled by default.

Related commit: https://gitlab.com/fedora/bootc/base-images/-/commit/6af3341f8037c872985e5b0e8a92e9470b18ddb7

### DIFF
--- a/modules/ROOT/pages/bootloader-updates.adoc
+++ b/modules/ROOT/pages/bootloader-updates.adoc
@@ -2,8 +2,8 @@
 
 == bootupd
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/bootloader-updates.adoc
dustymabe: I think we update the MBR now too. @HuijingHei do you know?

dustymabe:
Thanks @rich-purnell - I think some of the flow here can be cleaned up but we can do that in a followup PR.

PR #784, opened 2026-01-07
provisioning: include instructions for RPI3

sbaeurle:
Include instructions found in https://discussion.fedoraproject.org/t/fcos-unusually-slow-on-raspberry-pi-3b/40635/16 into the documentation.

Commands verified locally with 3 RPi 3B+ running FCOS

### DIFF
--- a/modules/ROOT/pages/provisioning-raspberry-pi4.adoc
+++ b/modules/ROOT/pages/provisioning-raspberry-pi4.adoc
@@ -1,4 +1,4 @@
-= Provisioning Fedora CoreOS on the Raspberry Pi 4
+= Provisioning Fedora CoreOS on the Raspberry Pi 4/3
 
 Fedora CoreOS produces 64-bit ARM (`aarch64`) artifacts. These images can be used as the Operating System for the Raspberry Pi 4 device. Before trying to get FCOS up and running on your Raspberry Pi 4 you'll want to xref:#_updating_eeprom_on_raspberry_pi_4[Update the EEPROM] to the latest version and choose how you want to boot the Raspberry Pi 4. There are two options:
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/provisioning-raspberry-pi4.adoc
```diff
@@ -83,6 +83,37 @@ Now take the USB/SD card and attach it to the RPi4 and boot.
...
 TIP: It can take some time to boot, especially if the disk is slow. Be patient. You may not see anything on the screen for 20-30 seconds.
 
+== Adjust Partition Table on Raspberry Pi 3 models
+
+In contrast to RPi4 GPT-compatible EEPROM based bootloader, the RPi3 Chip utilizes a hardcoded boot routine that looks for the first FAT32/FAT16 partition in the MBR of the SD card.
+To support Core OS on these devices, on must configure a hybrid MBR enabling the chip to find the boot partition.
+
+[source, bash]
+----
+sudo gdisk
```
dustymabe: I wonder if it would be better to script this somehow, like by piping configuration through sfdisk or something?

Then it's less interactive (less chance for things to go wrong).

modules/ROOT/pages/provisioning-raspberry-pi4.adoc
```diff
@@ -83,6 +83,37 @@ Now take the USB/SD card and attach it to the RPi4 and boot.
...
+Set the bootable flag? (Y/N): Y
+
+Unused partition space(s) found. Use one to protect more partitions? (Y/N): N
+
+w - write table to disk and exit
+
+Confirm any questions with: Y
+----
+
+During the first boot, CoreOS overwrites the partition table, thus requiring a second creation of the hybrid MBR
```
dustymabe: Hmm. I wouldn't expect this but I guess maybe the growing of the root partition somehow undoes the changes you made.

Can you try booting an instance with an extra kernel parameter to disable the growfs from happening:

```
[dustymabe@media ~]$ cat /tmp/foo.sh 
#!/bin/bash                                                            
# Create dropin for disabling # ignition-ostree-growfs.service
DROPIN_CRED=$(base64 -w 0 <<EOF
[Service]
ExecStart=
ExecStart=echo "skipping ignition-ostree-growfs.service"
EOF
)
echo "systemd.set_credential_binary=systemd.unit-dropin.ignition-ostree-growfs.service:${DROPIN_CRED}"
[dustymabe@media ~]$ bash /tmp/foo.sh 
systemd.set_credential_binary=systemd.unit-dropin.ignition-ostree-growfs.service:W1NlcnZpY2VdCkV4ZWNTdGFydD0KRXhlY1N0YXJ0PWVjaG8gInNraXBwaW5nIGlnbml0aW9uLW9zdHJlZS1ncm93ZnMuc2VydmljZSIK
```

So the kernel parameter would then be:

```
systemd.set_credential_binary=systemd.unit-dropin.ignition-ostree-growfs.service:W1NlcnZpY2VdCkV4ZWNTdGFydD0KRXhlY1N0YXJ0PWVjaG8gInNraXBwaW5nIGlnbml0aW9uLW9zdHJlZS1ncm93ZnMuc2VydmljZSIK
```

When you do this you should get a CLHM warning about the root filesystem being smaller than recommended:

```
Fedora CoreOS 43.20260106.dev.0

############################################################################
WARNING: The root filesystem is too small. It is strongly recommended to
allocate at least 8 GiB of space to allow for upgrades. From June 2021, this
condition will trigger a failure in some cases. For more information, see:
https://docs.fedoraproject.org/en-US/fedora-coreos/storage/

You may delete this warning using:
sudo rm /etc/motd.d/60-coreos-rootfs-size.motd
############################################################################

Tracker: https://github.com/coreos/fedora-coreos-tracker
Discuss: https://discussion.fedoraproject.org/tag/coreos

[core@cosa-devsh ~]$ 

```

PR #811, opened 2026-06-05
provisioning: Update requirements for Ignition configs

### LINKED ISSUE #748: Showcase provisioning with SSH keys setup by Afterburn first, Ignition second
In our provisioning pages, we say that users must create an Ignition config to setup a node and this is not true for the platforms where we have ssh key support via Afterburn (it is usually written on a line just below but that's too late and contradictory).

Instead, we should showcase the follow using the SSH keys provisioned in the cloud platform first, as the user will not have to provide an Ignition config at all in this case to get started.

In a second step on the provisioning page, we show the Ignition path.

---

Related to: https://github.com/coreos/fedora-coreos-tracker/issues/2006

travier:
Clearly note that the user can start and access an instance without providing an Ignition config on the platforms where we have Afterburn support.

Fixes: https://github.com/coreos/fedora-coreos-docs/issues/748

### DIFF
--- a/modules/ROOT/pages/getting-started.adoc
+++ b/modules/ROOT/pages/getting-started.adoc
@@ -21,6 +21,10 @@ Each platform has specific logic to retrieve and apply the first boot configurat
 
 For more information on configuration, refer to the documentation for xref:producing-ign.adoc[Producing an Ignition File].
 
+NOTE: While we recommend using Ignition to configure Fedora CoreOS, it is not required to get started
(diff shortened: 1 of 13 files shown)


travier:
This re-factor the prerequisite block on all platforms with Afterburn support. We could also do something similar for those without if folks think that would make sense.

nikodunk:
Looks fantastic, thank you!

dustymabe (approved):

PR #815, opened 2026-06-11

coreosbot-releng:


hrismarin (approved):
The images are available for download from the Fedora website.
The warning that causes failed checks will be fixed in the next Butane release.

PR #817, opened 2026-06-26
customizing-nic: update docs about ifname kernel argument

dustymabe:
https://github.com/coreos/fedora-coreos-tracker/issues/553 was fixed some time ago so we can update the docs here.


marmijo (approved):
LGTM

PR #824, opened 2026-08-19
docs: correct systemd automount example

### LINKED ISSUE #823: Improve systemd automount documentation
Currently in [CoreOS docs](https://docs.fedoraproject.org/en-US/fedora-coreos/storage/#_adding_network_storage), the recommended way of adding an automount is:
```yaml
variant: fcos
version: 1.3.0
systemd:
  units:
    - name: var-mnt-data.mount
      contents: |
        [Unit]
        Description=Mount data directory

        [Mount]
        What=example.org:/data
        Where=/var/mnt/data
        Type=nfs4

        [Install]
        WantedBy=multi-user.target

    - name: var-mnt-data.automount
      enabled: true
      contents: |
        [Unit]
        Description=Automount data directory

        [Automount]
        TimeoutIdleSec=20min
        Where=/var/mnt/data

        [Install]
        WantedBy=multi-user.target
```

Here the [Install] section in .mount unit is redundant and potentially unsafe if user tries to enable the .mount unit manually after installation. Ideally there should be no [Install] section inside "on-demand activated" units such as when using Timers, Automounts etc. [RHEL Docs](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/10/html/managing_file_systems/mounting-file-systems-on-demand#using-systemd-automount-to-mount-a-file-system-on-demand-with-a-mount-unit) also follow this convention.
Not having [Install] section in .mount unit would prevent user from running `systemctl enable var-mnt-data.mount` manually as this would fail with no Install target.

Corrected version would be:
```yaml
variant: fcos
version: 1.3.0
systemd:
  units:
    - name: var-mnt-data.mount
      contents: |
        [Unit]
        Description=Mount data directory

        [Mount]
        What=example.org:/data
        Where=/var/mnt/data
        Type=nfs4

    - name: var-mnt-data.automount
      enabled: true
      contents: |
        [Unit]
        Description=Automount data directory

        [Automount]
        TimeoutIdleSec=20min
        Where=/var/mnt/data

        [Install]
        WantedBy=multi-user.target
```

Amir-A664:
## Summary

- remove the Install section from the mount unit in the automount example
- clarify that the automount unit activates the mount unit on demand

The mount unit is intended to be activated by the automount unit. Providing an Install section suggests that users can enable the mount unit directly, bypassing the documented on-demand behavior.

Fixes #823

## Testing

- `ci/check.py -v`
- Antora documentation build with the official Fedora UI bundle

### DIFF
--- a/modules/ROOT/pages/storage.adoc
+++ b/modules/ROOT/pages/storage.adoc
@@ -482,9 +482,6 @@ systemd:
         Where=/var/mnt/data
         Type=nfs4
 
-        [Install]
-        WantedBy=multi-user.target
-
     - name: var-mnt-data.automount
       enabled: true
       contents: |
@@ -498,6 +495,8 @@ systemd:
         [Install]
         WantedBy=multi-user.target
 ----
+
+Only the `.automount` unit is enabled. It activates the `.mount` unit on demand when the mount point is accessed.
 
 == Advanced examples
 


dustymabe (approved):
LGTM

Amir-A664:
Yayyy

